PR #95, opened 2018-09-04
Feature/sig 527 change category visible feedback

jasperswart:
this adds a loader to the categoy and the status buttons


src/signals/incident-management/containers/IncidentCategoryContainer/components/Add/index.js
```diff
@@ -20,7 +21,8 @@ class Add extends React.Component { // eslint-disable-line react/prefer-stateles
   }
 
   render() {
-    const { subcategoryList } = this.props;
+    const { subcategoryList, loading } = this.props;
+    this.categoryForm.controls.forceRender.setValue(Math.random());
```
RobertLodder: What is the idea behind this forceRender, having a random value?

src/signals/incident-management/containers/IncidentStatusContainer/components/Add/index.js
```diff
@@ -53,6 +55,11 @@ class Add extends React.Component { // eslint-disable-line react/prefer-stateles
 
                   <button className="action primary" type="submit" disabled={invalid}>
```
RobertLodder: Do we also want to disable the button when (loading === true)?

src/signals/incident-management/containers/IncidentCategoryContainer/components/Add/index.js
```diff
@@ -29,10 +31,23 @@ class Add extends React.Component { // eslint-disable-line react/prefer-stateles
...
+                    render={SelectInput}
+                    name="sub"
+                    display="Categorie"
+                    control={this.categoryForm.get('sub')}
+                    values={subcategoryList}
+                    multiple={false}
+                    emptyOptionText="Selecteer..."
+                  />
 
                   <button className="action primary" type="submit" disabled={invalid}>
```
RobertLodder: Do we also want to disable the button when (loading === true)?

src/signals/incident-management/containers/IncidentCategoryContainer/components/Add/index.js
jasperswart: the Child form containing the button does not rerender when the Parent is rendered with the new loading value. so I added a form control  forceRender that changes when the Parent is rendered. i agree it is Rather ugly.

src/signals/incident-management/containers/IncidentCategoryContainer/components/Add/index.js
jasperswart: yes i will add that

src/signals/incident-management/containers/IncidentStatusContainer/components/Add/index.js
jasperswart: yes i will add that

src/signals/incident-management/containers/IncidentCategoryContainer/components/Add/index.js
RobertLodder: It's neater to use the loading prop for that, so something like:
`this.categoryForm.controls.loading.setValue(loading);`
instead of some random value for forceRender. Maybe add a comment to make clear why it's done.

jasperswart: ja dat is eleganter. ik stop het erin

PR #111, opened 2018-09-18
refactored render of wizard step

jasperswart:
using Render methode of wizard is down this should be implemented according to the docs.


KelvinPaska:
test this please

PR #119, opened 2018-09-20
Bugfix/SIG-488 map now works correctly

jasperswart:


src/signals/incident/components/IncidentPreview/components/Map/__snapshots__/index.test.js.snap
```diff
@@ -30,7 +30,7 @@ exports[`Preview component <Map /> should render only address correctly 1`] = `
       >
         <div>
           <div>
-            Hell 666C
+            Hell 666C, undefined undefined
```
RobertLodder: The test data is incomplete, if undefined is shown here?

jasperswart: scherp! ja verbeterd

PR #138, opened 2018-10-11
fix Foutmelding in frontend, niet gemachtigd knop te drukken SIG-700

jasperswart:


src/signals/incident-management/containers/IncidentStatusContainer/reducer.js
```diff
@@ -43,8 +43,11 @@ function incidentStatusContainerReducer(state = initialState, action) {
     case REQUEST_STATUS_LIST_ERROR:
     case REQUEST_STATUS_CREATE_ERROR:
       return state
+      .set('loading', false)
+      .set('loadingExternal', false)
```
RobertLodder: Should have the same indentation level as the '.set's below?

PR #152, opened 2018-10-22
updated red info bar

jasperswart:


src/signals/incident/containers/IncidentContainer/index.js
```diff
@@ -36,7 +36,7 @@ export class IncidentContainer extends React.Component {
       <div className="incident-container">
         <div className="incident-container__alert">
           <b>*** BELANGRIJK ***</b><br />
-            Melding over horeca of evenementen? Vul dan bij de melding altijd uw contactgegevens in, dan kunnen wij u beter en sneller helpen. Spoed? Bel dan met 14 020.<br /><br />
+            Melding over horeca of evenementen? Vul dan bij de melding altijd uw contactgegevens in, dan kunnen wij u beter en sneller helpen.<br /><br />
```
RobertLodder: The snapshot is not in accordance with the changed text, so the unit test will fail

PR #159, opened 2018-11-01
Master to dev

jasperswart:


bvhme:
@RobertLodder a case where this can happen is if a hotfix is applied from a hotfix branch directly into production/release, which is not the case here but can happen.

PR #166, opened 2018-11-13
Bugfix/sig 617 add phone number validation

jasperswart:
is finished with all tests now!


src/signals/incident/definitions/wizard-step-3-telefoon.js
```diff
@@ -12,11 +13,15 @@ export default {
...
-          type: 'text'
+          type: 'text',
+          autoRemove: /[^\d+-]/g
         },
         render: FormComponents.TextInput,
         options: {
-          validators: [Validators.maxLength(17)]
+          validators: [
+            Validators.maxLength(17),
+            validatePhoneNumber
```
RobertLodder: Now that illegal phone number characters are automatically removed, there is not much use for this validator

PR #254, opened 2019-06-26
Quickfix incidents without priority: hiding element

RRMoelker:


RRMoelker:
Besproken, forcing merge

PR #258, opened 2019-07-02
Sig 1303 internationalization

RRMoelker:
Top navigation links formatted through messages


src/components/MainMenu/index.js
```diff
@@ -33,36 +35,36 @@ export class MainMenu extends React.Component { // eslint-disable-line react/pre
...
                     <span className="linklabel">
-                      Beheer standaard teksten
+                      <FormattedMessage {...messages.manage_standaard_teksten} />
                     </span>
                   </NavLink>
-                </li> : ''
+                </li>
               }
-              {/* this.props.isAuthenticated ?
+              {/* this.props.isAuthenticated &&
```
janjaap: Why not remove this block? It's not used.

src/components/MainMenu/messages.js
```diff
@@ -0,0 +1,19 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  new_melding: {
+    id: 'menu.new_melding',
+    description: 'Top navigation new melding label',
+    defaultMessage: 'Nieuwe melding', // TODO decide if we want to use functionality (overlap in functionality)
```
janjaap: It's good practice to have default values in case someone forgets to add a translation string to the json file. If you think this is an issue that should be addressed then please create a ticket instead of a mention in the code here.

src/containers/LanguageProvider/index.js
```diff
@@ -14,18 +14,18 @@ import { IntlProvider } from 'react-intl';
...
-    return (
-      <IntlProvider locale={this.props.locale} key={this.props.locale} messages={this.props.messages[this.props.locale]} defaultLocale="nl">
-        {React.Children.only(this.props.children)}
-      </IntlProvider>
-    );
-  }
-}
+
+export const LanguageProvider = ({ messages, locale, children }) => {
+  console.log(messages);
```
janjaap: This line should be removed

src/translations/NOTE.md
```diff
@@ -0,0 +1,8 @@
+Keys for .json files are generated by `npm run extract-intl`.
+
+Any keys added to the json file will be **removed** when running the above command!
+
+The "correct" way is to proxy message configuration through `messages.js` files.
+See: https://github.com/react-boilerplate/react-boilerplate/blob/master/docs/js/i18n.md
+
+(don't shoot the messenger)
```
janjaap: Is this line necessary?

src/translations/nl.json
```diff
@@ -1,11 +1,6 @@
...
-  "boilerplate.components.Footer.license.message": "Dit project valt onder Mozilla Public License 2.0.",
-  "boilerplate.components.Header.features": "",
-  "boilerplate.components.Header.home": "",
-  "boilerplate.containers.HomePage.start_project.header": "Welkom bij de Signalen Informatievoorziening Amsterdam",
-  "boilerplate.containers.HomePage.start_project.message": "Ontwikkeld door Datapunt Amsterdam",
-  "boilerplate.containers.LocaleToggle.nl": "",
-  "boilerplate.containers.LocaleToggle.en": "",
-  "boilerplate.containers.NotFoundPage.header": "Pagina niet gevonden."
+  "boilerplate.containers.NotFoundPage.header": "Pagina niet gevonden.",
+  "menu.afhandelen": "Afhandelen",
```
janjaap: What if the string 'Afhandelen' is used in a component other than `Menu`? The prefix `menu` doesn't make sense there and, in the future, can lead to duplicate translations because of that. I would choose a context independent prefix, if at all.

RRMoelker: The idea of these translation is to target a specific use right? So the string "Afhandelen" might occur somewhere else as well. My sense is that we want it to be declared multiple times. Because the contexts are different which means the meaning may be different.

So having a `menu.afhandelen` makes sense. As well as maybe a button somewhere named: `"admin.melding.detail.afhandelen": "Afhandelen"`

src/components/MainMenu/index.js
RRMoelker: Yeah, why not

src/components/MainMenu/messages.js
RRMoelker: I don't think we want the default values. Because we also have the same value in the messages for the default language. So there is some duplication there.

Anyway, the react boilerplate extract messages command doesn't work if we leave this one out. So let's keep it in.

BTW, the todo was there so we would forget discussing it in this PR. I'll remove it.

src/containers/LanguageProvider/index.js
RRMoelker: ty

PR #267, opened 2019-07-16
Sig 1303 ovl pluralization

RRMoelker:


Dockerfile
```diff
@@ -31,13 +28,17 @@ COPY environment.conf.${BUILD_ENV}.json /app/environment.conf.json
...
 # Install NPM dependencies. Also:
 RUN npm --production=false \
         --unsafe-perm \
         --verbose \
-       install
-RUN npm cache clean --force
+       install && npm cache clean --force
+
+# Copy sources
+COPY src /app/src
```
janjaap: I would move this line below line 40; code in `src` is more likely to change than code in the `internals` and `server` folders

src/components/MainMenu/__snapshots__/index.test.js.snap
```diff
@@ -21,7 +21,12 @@ exports[`<MainMenu /> rendering should render Nieuwe Melding NavLink component w
             <span
               className="linklabel"
             >
-              Nieuwe melding
+              <FormattedMessage
```
janjaap: Why isn't the rendered result of the `FormattedMessage` component in the snapshot?

src/containers/App/router.js
```diff
@@ -0,0 +1,29 @@
...
+      component={
+        (props) => (<KtoContainer yesNo={props.match.params.yesNo} uuid={props.match.params.uuid} />)
+      }
+    />
+    <Route path="" component={NotFoundPage} />
+  </Switch>
+);
+
+Router.propTypes = {
+  match: PropTypes.object
```
janjaap: This should be
```jsx
  match: PropTypes.shape({
    params: PropTypes.shape({
      yesNo: PropTypes.string,      // required?
      uuid: PropTypes.string,       // required?
    }).isRequired,
  }).isRequired,
```

src/signals/incident/components/IncidentForm/index.js
```diff
@@ -29,7 +30,9 @@ class IncidentForm extends React.Component {
 
   setForm(form, incidentContainer) {
     this.form = form;
+    const incident = get(incidentContainer, 'incident', {});
```
janjaap: Why do you need an extra dependency here? Wouldn't 
```jsx
const { incident = {} } = incidentContainer || {};
```
also do the trick?

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
+  } = meta;
 
-      <div className="header__errors__item">
-        {touched
-        && hasError('email')
-        && 'Het moet een geldig e-mailadres zijn'}
-      </div>
+  const incident = get(parent, 'meta.incident', {});
+  const shallowValues = flattenObject(incident, '', ':');
+  const isLabelDescriptor = label !== null && typeof label === 'object';
```
janjaap: With the types listed below, the condition will also be truthy. Is that intentional?

```jsx
typeof {a: 1} === 'object';
typeof [1, 2, 4] === 'object';
typeof new Date() === 'object';
typeof /regex/ === 'object';
typeof new Boolean(true) === 'object'; 
typeof new Number(1) === 'object'; 
typeof new String('abc') === 'object';
```

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
+          <span className="header--not-required">(optioneel)</span>
+          : ''}
       </div>
+      <div className="header__subtitle">{subtitle}</div>
+
+      <div className="header__errors">
+        <div className="header__errors__item">
+          {touched
+          && hasError('required')
+          && 'Dit is een verplicht veld'}
```
janjaap: Translations?

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
-        && `U kunt maximaal ${getError('maxLength').requiredLength} tekens invoeren.`}
+  return (
+    <div className={`header ${touched && (hasError('required') || hasError('email') || hasError('maxLength') || hasError('custom')) ? 'header--invalid' : ''}`}>
+      <div className="header__label">{
+        isLabelDescriptor ?
+          <FormattedMessage {...label} values={shallowValues} />
+          : label
+        }
+        {(label && (!options || !options.validators)) || (options && options.validators && !options.validators.includes(Validators.required)) ?
+          <span className="header--not-required">(optioneel)</span>
```
janjaap: Translations?

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
+        </div>
 
-      <div className="header__errors__item">
-        {touched
-        && hasError('custom')
-        && getError('custom')}
+        <div className="header__errors__item">
+          {touched
+          && hasError('email')
+          && 'Het moet een geldig e-mailadres zijn'}
```
janjaap: Translations?

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
+        </div>
+
+        <div className="header__errors__item">
+          {touched
+          && hasError('maxLength')
+          && `U kunt maximaal ${getError('maxLength').requiredLength} tekens invoeren.`}
```
janjaap: Translations?

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
     </div>
-  </div>
-);
+  );
+};
 
 Header.propTypes = {
-  meta: PropTypes.object,
+  meta: PropTypes.shape({
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
janjaap: Using `PropTypes.object` should fail; discouraged and I believe even disallowed by the linter configuration. Use `PropTypes.shape({})` instead and define the props that you expect to receive.

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
+    subtitle: PropTypes.string,
+  }),
   options: PropTypes.object,
   touched: PropTypes.bool,
   hasError: PropTypes.func,
   getError: PropTypes.func,
-  children: PropTypes.object
+  children: PropTypes.object,
```
janjaap: Shouldn't this be `PropTypes.node`?

src/signals/incident/components/form/Header/index.js
```diff
@@ -1,58 +1,80 @@
...
+  parent: PropTypes.object
```
janjaap: Same here.

src/signals/incident/components/form/RadioInput/index.js
```diff
@@ -1,52 +1,84 @@
...
+            </div>
+          </Header>
+        </div>
+      }
+    </div>
+  );
+};
+
+_RadioInput.propTypes = {
+  intl: PropTypes.object,
```
janjaap: Should be
```jsx
  intl: intlShape.isRequired,
```
and line 5 should be
```jsx
import { injectIntl, intlShape } from 'react-intl';
```

src/signals/incident/components/form/RadioInput/index.js
```diff
@@ -1,52 +1,84 @@
...
+ * React-intl injector creates a class component from our function component.
+ * This class component is wrapped inside a function compontent because react-reactive-forms
+ * doesn't seem to work with class components in our way of using the form generator
+ * (it expects a functional component).
+ */
+const IntlProvider = injectIntl(_RadioInput);
+const FunctionWrappedComponent = (props) => <IntlProvider {...props} parent={props.parent} />;
+
+FunctionWrappedComponent.propTypes = {
+  parent: PropTypes.object
```
janjaap: Can this be defined as `parent: PropTypes.shape({})`? Or is there to knowledge about the props that `parent` has?

src/signals/incident/containers/IncidentContainer/debug/initialState.js
```diff
@@ -51,9 +51,9 @@ export default {
   // ...overlastPersonen,
 
   // wegen-verkeer-straatmeulilair:
-  ...brug,
+  // ...brug,
```
janjaap: Can't we remove the commented out lines? Doesn't look like we need them...

src/signals/incident/containers/IncidentContainer/debug/straatverlichting.js
```diff
@@ -16,8 +16,8 @@ export default {
     label: 'Lichtpunt is zichtbaar beschadigd en/of incompleet'
   },
   extra_straatverlichting_hoeveel: {
-    id: '1_lichtpunt',
-    label: '1 lichtpunt'
+    id: 'meerdere_lichtpunten',
+    label: 'Een aantal lichtpunten die bi…an/hangen'
```
janjaap: Is the use of the ellipsis intended?

src/signals/incident/services/format-incident-message/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Either formats message through i18n intl context or simply passes back string value
+ * @param intl
+ * @param value
+ * @param incident
+ * @returns {*}
+ */
+const formatIncidentMessage = (intl, value, incident) => {
+  if (typeof value === 'object') {
```
janjaap: Same comment about checking for type applies here.

src/signals/incident/services/object-flatten/index.js
```diff
@@ -0,0 +1,10 @@
+// based on source: https://30secondsofcode.org/object#flattenobject
+const flattenObject = (obj, prefix = '', separator = '.') =>
+  Object.keys(obj).reduce((acc, k) => {
+    const pre = prefix.length ? prefix + separator : '';
```
janjaap: Use template literals:
```jsx
 const pre = prefix.length ? `${prefix}${separator}` : '';
```

src/signals/incident/services/object-flatten/index.js
```diff
@@ -0,0 +1,10 @@
...
+    if (typeof obj[k] === 'object') Object.assign(acc, flattenObject(obj[k], pre + k, separator));
```
janjaap: Same comment about type checking applies here.
Also, use template literals

src/signals/incident/services/object-flatten/index.js
```diff
@@ -0,0 +1,10 @@
...
+    else acc[pre + k] = obj[k];
```
janjaap: Use template literals

src/translations/en.json
```diff
@@ -1,14 +1,3 @@
...
-  "boilerplate.components.Header.features": "Features",
-  "boilerplate.components.Header.home": "Home",
-  "boilerplate.containers.HomePage.start_project.header": "Start your next react project in seconds",
-  "boilerplate.containers.HomePage.start_project.message": "A highly scalable, offline-first foundation with the best DX and a focus on performance and best practices",
-  "boilerplate.containers.HomePage.tryme.atPrefix": "@",
-  "boilerplate.containers.HomePage.tryme.header": "Try me!",
-  "boilerplate.containers.HomePage.tryme.message": "Show Github repositories by",
-  "boilerplate.containers.LocaleToggle.nl": "nl",
-  "boilerplate.containers.LocaleToggle.en": "en",
   "boilerplate.containers.NotFoundPage.header": "Page not found."
```
janjaap: Not all translations seem to be present in this file.

Dockerfile
RRMoelker: :)

src/components/MainMenu/__snapshots__/index.test.js.snap
RRMoelker: Test is using shallow rendering, so child component is being stubbed.

src/containers/App/router.js
RRMoelker: Agreed. Specific params depend on actual route, so not making them required

src/components/MainMenu/__snapshots__/index.test.js.snap
janjaap: But what is the point of having such an output in a snapshot? Wouldn't it be better to have the rendered output in the snapshot?

src/signals/incident/components/form/Header/index.js
RRMoelker: The idea is that the input is either a LabelDescriptor or a simple string. The check here is simply to distinguish between those two. So i think the current code satisfies that requirement.

However, clearly it's unclear what kind of object it is. So clarified the prop type a bit.

src/signals/incident/services/format-incident-message/index.js
RRMoelker: I made the check more specific. Though it seems we should consider a proper typing system if we're going to pin down objects in JS to this degree.

src/signals/incident/services/object-flatten/index.js
RRMoelker: Disagree, don't think it will improve readability here

src/signals/incident/services/object-flatten/index.js
RRMoelker: Disagree, highly specialized function. No need to go overboard with type checking and the likes.

src/translations/en.json
RRMoelker: Correct, given that we only support `nl` at the moment. I don't think we should get these translations up.

src/signals/incident/services/object-flatten/index.js
janjaap: In my opinion it will. Using `a + b` implies an arithmetic operation instead of a string concatenation.

src/components/MainMenu/__snapshots__/index.test.js.snap
RRMoelker: I prefer testing only the file (unit). So that does not include the children.

Also testing all the children is a valid (and even more valuable) test. But mixing these unit tests with more broad tests leads to issues where test break unexpectedly. In addition to being a lot slower.

src/signals/incident/components/IncidentForm/index.js
RRMoelker: Sure

src/signals/incident/services/object-flatten/index.js
RRMoelker: I think I got a nicer solution with a template literal. Should be to your liking. Resolving

src/components/MainMenu/__snapshots__/index.test.js.snap
janjaap: But you're not testing anything here; you're generating a snapshot. If you were to verify that a `FormattedMessage` component is present in the tree, then we would not need a snapshot in the first place.

RRMoelker: I would have tested the presence of a FormattedMessage. The snapshot was already there, in general I feel these snapshots are used to write less test code rather than do a true snapshot comparison. And there is something to say about ease of use. However I don't want to focus on updating all these old tests in this PR.

janjaap: Fair enough

PR #276, opened 2019-07-23
(Chore) Packages updates

janjaap:
This pull request upgrades most of the dependencies for this project. In summary, it:
- syncs the dependencies with [`react-boilerplate` v4.0.0](https://github.com/react-boilerplate/react-boilerplate/blob/v4.0.0/package.json#L66-L166)
- adds prettier and eslint config
- moves Babel config from `package.json` to its own config file
- replaces `react-router-redux` with `connected
- removes the `internals/templates` folder and its corresponding script commands
- alters the Webpack config for all environments so that it plays nice with the new (major) dependency versions
- replaces `immutable` with `produce`. All occurrences of `toJS` and `fromJS` have been removed from the code and reducers and selectors are modified accordingly
- all calls to set and get data from global state have changed so that they don't rely on `get()` and `set()` calls


janjaap:
Closing this PR; superseded by smaller PRs

PR #278, opened 2019-07-25
SIG-1483 Remove filters from overview page

janjaap:
This PR:
- adds the package `lodash.clonedeep`
- makes the site footer sticky so that all pages cover at least the full height of the viewport
- for some selectors sets the initial state; this makes testing easier when rendering components that depend on data that should come from the global state
- removes the `Filter` component from the `IncidentOverviewPage` container and turns it into a separate container and component inside the `signals/incident-management` folder
- rewrites tests so that snapshots are replaced by rendered element assertions


src/components/Footer/__snapshots__/index.test.js.snap
```diff
@@ -1,69 +1,57 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
jasperswart: zullen we dan overal de snapshots weghalen?

janjaap: Slowly but surely replace them, yes.

PR #296, opened 2019-08-28
SIG-1562 Filter styling modifications

janjaap:
This PR:
- adds two polyfills; `FormData` and `NodeList.forEach`
- has polyfill loading handled by Webpack instead of bluntly importing it in `app.js`
- upgrades the `@datapunt/asc-` packages to their latest versions
- uses SVG icons from `@datapunt/asc-assets` as intended (see https://amsterdam.github.io/amsterdam-styled-components for reference)
- makes sure that button labels are shown in the correct font
- renames and reorders menu items in the `SiteHeader` component
- corrects the date input placeholder text for the `FilterForm` component


src/components/Modal/index.js
```diff
@@ -67,15 +61,16 @@ const Modal = ({ children, title, isOpen, onClose, ...rest }) => (
           <Heading as="h2">{title}</Heading>
         </Column>
 
-        <CloseButton
+        <Button
           data-testid="closeBtn"
           square
           type="button"
-          as="button"
```
jasperswart: add $as

src/components/PageHeader/index.js
```diff
@@ -45,20 +52,24 @@ const PageHeader = ({ className, children, title }) => {
     };
   });
 
+  const StyledHeading = styled(Heading)`
+    font-size: 20px;
+  `;
```
jasperswart: you should also add line-height I think

src/components/SiteHeader/index.js
```diff
@@ -5,36 +5,28 @@ import styled from 'styled-components';
 import Media from 'react-media';
 
 import CONFIGURATION from 'shared/services/configuration/configuration';
-import LoginIcon from '@datapunt/asc-assets/lib/Icons/Login.svg';
-import LogoutIcon from '@datapunt/asc-assets/lib/Icons/Logout.svg';
+import { Login as LoginIcon, Logout as LogoutIcon } from '@datapunt/asc-assets';
 import {
   Header as HeaderComponent,
+  MenuButton,
```
jasperswart: I also put the MenuButton in the asc 0.12.7 header in my branch

src/components/Modal/index.js
janjaap: Isn't necessary; component already renders a `<button>` element. `$as` Would only be required if I would want to render the `Button` component as a `<span>`, for instance.

src/components/PageHeader/index.js
janjaap: `line-height` Has already been provided by the `asc-ui` package. Don't need to set it again.

janjaap:
@jasperswart Please review again; comments addressed.

PR #297, opened 2019-08-29
SIG-1561 Filter by ID

janjaap:
This PR:
- Adds the `SearchBar` component to the `SiteHeader`:
  Screenshot 2019-08-29 at 14 59 19
  The input element only accepts numeric values to prevent the API from returning a `400 - Bad request` header.
- Updates the `@datapunt/asc-` packages to their latest version and adjusts component implementation accordingly
- Changes the `SiteHeader` component so that it can handle the search form submission
- Changes the `PageHeader` component so that it can render a subtitle
- Adds `search` model


src/containers/SiteHeader/index.js
```diff
@@ -9,9 +9,9 @@ import {
 } from 'containers/App/selectors';
 import SiteHeader from 'components/SiteHeader';
 import { withRouter } from 'react-router-dom';
+import { isAuthenticated } from 'shared/services/auth/auth';
```
jasperswart: daar is een selector voor in app/selectors

PR #328, opened 2019-09-18
SIG-1640 Checkbox group selection fix

janjaap:
This PR contains changes that fix an issue where checkboxes in a group were checked where they shouldn't be. The origin of the issue was values of slug properties of categories coming from the API. This fix in this PR no longer matches on slugs, but on id's (or keys).

### DIFF
--- a/src/shared/services/map-categories/index.js
+++ b/src/shared/services/map-categories/index.js
@@ -25,7 +25,6 @@ function mapCategories(data) {
 
           mainToSub[category.slug].push({
             id: subcategory._links && subcategory._links.self && subcategory._links.self.href,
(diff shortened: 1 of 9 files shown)


src/shared/services/map-categories/index.js
```diff
@@ -24,7 +24,7 @@ function mapCategories(data) {
           });
 
           mainToSub[category.slug].push({
-            key: subcategory.slug, // replacing 'key' prop since its value isn't used and it makes more sense for the filter form to have to deal with just one prop; 'key'
+            id: subcategory._links && subcategory._links.self && subcategory._links.self.href,
```
jasperswart: ter info: hernoemen van key nara id kan hier geen kwaad, maar de de FieldWrapper gaat uit van keys en geen id's

PR #330, opened 2019-09-19
SIG-1650 Edit filter button fix

janjaap:
This PR fixes an issue where, in IE11, an error would be thrown when `document.dispatchEvent` would be called.

### DIFF
--- a/src/signals/incident-management/containers/MyFilters/components/FilterItem/__tests__/FilterItem.test.js
+++ b/src/signals/incident-management/containers/MyFilters/components/FilterItem/__tests__/FilterItem.test.js
@@ -83,6 +83,32 @@ describe('signals/incident-management/containers/MyFilters/components/FilterItem
     expect(event.preventDefault).toHaveBeenCalled();
   });
 
+  it('should dispatch event in IE11', () => {
+    const Event = global.Event;
+    global.Event = null;
+    const createEventSpy = jest.spyOn(document, 'createEvent');
+    const props = {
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/containers/MyFilters/components/FilterItem/__tests__/FilterItem.test.js
```diff
@@ -83,6 +83,32 @@ describe('signals/incident-management/containers/MyFilters/components/FilterItem
...
+
+    const handleEditFilterButton = getByTestId('handleEditFilterButton');
+    const event = createEvent.click(handleEditFilterButton, { button: 1 });
+    event.preventDefault = jest.fn();
+
+    fireEvent(handleEditFilterButton, event);
+
+    expect(createEventSpy).toHaveBeenCalled();
+
+    global.Event = Event;
```
gilleswittenberg: Maybe nicer to move `global.Event = null;` and this line to a `before` and `after`

janjaap: Could do, but then I'd need a `describe` to wrap the test in it, otherwise it would be for all tests in this suite. If there would be more tests that require this approach, I would say yes. For now, it seems to be a bit overkill.

gilleswittenberg: Agree

PR #349, opened 2019-09-25
(Chore) Optimise Dockerfile

janjaap:
This PR:
- groups commands in the `Dockerfile` and moves commands to other lines to improve container reuse
- applies different deploy environments for both `develop` and `master` branches
- adds a `build` stage for branches other than `develop` and `master`
- contains a change in the `Jenkinsfile` that will improve build times, because of more efficient image reuse

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,6 +22,13 @@ node {
         env.GIT_COMMIT = scmVars.GIT_COMMIT
     }
 
+    stage("Get cached build") {
+        docker.withRegistry('https://repo.data.amsterdam.nl','docker-registry') {
+            def cachedImage = docker.image("ois/signalsfrontend:acceptance")
+            cachedImage.pull()
+        }
+    }
+
     stage("Lint") {
       String  PROJECT = "sia-eslint"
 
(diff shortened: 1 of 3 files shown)


janjaap:
@snelis Would it improve the build if I were to add the build args in [`Jenkinsfile`](https://github.com/Amsterdam/signals-frontend/blob/develop/Jenkinsfile#L56-L59) to the [`lint start`](https://github.com/Amsterdam/signals-frontend/blob/develop/Jenkinsfile#L29) and [`unittests start`](https://github.com/Amsterdam/signals-frontend/blob/develop/Jenkinsfile#L42) commands?

Jenkinsfile
```diff
@@ -22,6 +22,13 @@ node {
         env.GIT_COMMIT = scmVars.GIT_COMMIT
     }
 
+    stage("Get cached build") {
+        docker.withRegistry('https://repo.data.amsterdam.nl','docker-registry') {
+            def cachedImage = docker.image("ois/signalsfrontend:acceptance")
+            cachedImage.pull()
+        }
+    }
+
```
sercanyaldiz: @janjaap slim van je!

janjaap: Credit go to @snelis

PR #351, opened 2019-09-26
(Chore) Remove authenticated selector

janjaap:
This PR contains changes that prevents component from rerendering when the `makeSelectIsAuthenticated` selector updates its value and passes a different prop value to the appropriate components. 
The previous asynchronous behaviour caused several issues.; logging in wouldn't always redirect the user to the incidents overview page and the header component would render its `tall` version and, after getting a different value for the authentication, would render its `short` version.

### DIFF
--- a/src/containers/App/reducer.js
+++ b/src/containers/App/reducer.js
@@ -46,7 +46,10 @@ export const initialState = fromJS({
 function appReducer(state = initialState, action) {
   switch (action.type) {
     case AUTHENTICATE_USER:
-      global.sessionStorage.setItem(ACCESS_TOKEN, action.payload.accessToken);
+      if (action.payload && action.payload.accessToken) {
+        global.sessionStorage.setItem(ACCESS_TOKEN, action.payload.accessToken);
+      }
+
       return state;
 
     case AUTHORIZE_USER:
--- a/src/containers/App/reducer.js
+++ b/src/containers/App/reducer.js
@@ -13,6 +13,8 @@
 import { fromJS } from 'immutable';
 
 import {
+  LOGOUT,
+  AUTHENTICATE_USER,
   AUTHORIZE_USER,
   SHOW_GLOBAL_ERROR,
   RESET_GLOBAL_ERROR,
@@ -20,7 +22,7 @@ import {
   UPLOAD_REQUEST,
   UPLOAD_PROGRESS,
   UPLOAD_SUCCESS,
-  UPLOAD_FAILURE
+  UPLOAD_FAILURE,
 } from './constants';
 
 // The initial state of the App
@@ -29,16 +31,25 @@ export const initialState = fromJS({
   error: false,
   upload: {},
   userPermissions: [],
+  userName: undefined,
+  userScopes: undefined,
+  accessToken: undefined,
   categories: {
     main: [],
     sub: [],
     mainToSub: {},
-  }
+  },
 });
 
 function appReducer(state = initialState, action) {
   switch (action.type) {
+    case AUTHENTICATE_USER:
+      global.sessionStorage.setItem('accessToken', action.payload.accessToken);
+      return state;
+
     case AUTHORIZE_USER:
+      global.sessionStorage.setItem('accessToken', action.payload.accessToken);
+
       return state
         .set('userName', action.payload.userName)
         .set('userScopes', fromJS(action.payload.userScopes))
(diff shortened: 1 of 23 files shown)


src/containers/App/reducer.js
```diff
@@ -29,16 +33,28 @@ export const initialState = fromJS({
...
     mainToSub: {},
-  }
+  },
 });
 
 function appReducer(state = initialState, action) {
   switch (action.type) {
+    case AUTHENTICATE_USER:
+      if (action.payload && action.payload.accessToken) {
+        global.sessionStorage.setItem(ACCESS_TOKEN, action.payload.accessToken);
```
jasperswart: the global.sessionStorage is being set in the saga already: the login and logout functions in auth.js

src/containers/App/reducer.js
```diff
@@ -58,27 +74,38 @@ function appReducer(state = initialState, action) {
...
+          progress: action.payload,
+        }),
+      );
 
     case UPLOAD_SUCCESS:
     case UPLOAD_FAILURE:
+      return state.set('upload', fromJS({}));
+
+    case LOGOUT:
+      global.sessionStorage.removeItem(ACCESS_TOKEN);
```
jasperswart: de global.sessionStorage is being set in the saga already: the login and logout functions in auth.js

src/shared/services/api/api.test.js
```diff
@@ -17,17 +16,46 @@ describe('api service', () => {
   const queryString = 'name1=value1&name2=value2&value3=foo&value3=bar';
   const url = 'https://url/to/test';
   const token = 'bearer-token';
+  let origSessionStorage;
```
jasperswart: why do you not use `getAccessToken.mockImplementation(() => 'mock')` here?

src/signals/incident/containers/IncidentContainer/index.test.js
```diff
@@ -5,24 +5,38 @@ import { withAppContext } from 'test/utils';
...
     props = {
       incidentContainer: { incident: {} },
       getClassification: jest.fn(),
       updateIncident: jest.fn(),
       createIncident: jest.fn(),
-      isAuthenticated: false
+    };
+    origSessionStorage = global.sessionStorage;
+
+    global.sessionStorage = {
```
jasperswart: why do you not use `getAccessToken.mockImplementation(() => 'mock')` here to mock api service?

src/containers/App/reducer.js
janjaap: Good call. Removed it.

src/containers/App/reducer.js
janjaap: Another good call; also removed it.

src/shared/services/api/api.test.js
janjaap: I've added a package that mocks the sessionstorage instead.

src/signals/incident/containers/IncidentContainer/index.test.js
janjaap: See previous comment

src/shared/services/api/api.test.js
jasperswart: die package hoef je niet te gebruiken, als je:

import { getAccessToken } from 'shared/services/auth/auth';
jest.mock('shared/services/auth/auth');

en dan in elke test: 
getAccessToken.mockImplementation(() => true) // of false

dan laad hij geen andere services in.

dat zit ook op andere plekken die van getAccessToken gebruikmaken.

janjaap: Problem is that, when I start mocking that module, I need to make more changes in other, unrelated, components. Even if I restore mocks after the test suite has run. Using the localstorage mock package overcomes that problem.

src/shared/services/api/api.test.js
```diff
@@ -1,6 +1,5 @@
-import { call, select } from 'redux-saga/effects';
+import { call } from 'redux-saga/effects';
 import request from 'utils/request';
-import { makeSelectAccessToken } from 'containers/App/selectors';
 
 import { generateParams, authCall, authCallWithPayload, authPostCall, authPatchCall } from './api';
```
jasperswart: the jest.mock('containers/App/selectors' can be removed

src/shared/services/api/api.test.js
jasperswart: ok laat maar dan

src/shared/services/api/api.test.js
janjaap: You're right; I've removed it.

PR #371, opened 2019-10-07
SIG-1702 Missing service belofte

janjaap:
This PR fixes an issue where, when the prediction service was not available or threw another error, the 'thank you' page wouldn't show a generic promise to the incident creator.
In order to do so, this PR:
- updates `redux-saga-test-plan` package to its latest stable version
- adds a method to the API services: `postCall`
- adds fixtures for `incident` and `priority`
- calls `getClassificationError` action from the `getClassification` saga with an empty response and categories (actual fix)
- has the `getClassification` saga retry a couple of times before it throws a catchable error

### DIFF
--- a/src/utils/__tests__/fixtures/incident.json
+++ b/src/utils/__tests__/fixtures/incident.json
@@ -1,0 +1,89 @@
+{
+  "reporter": {
+    "email": "",
+    "phone": ""
+  },
+  "notes": [],
+  "extra_properties": null,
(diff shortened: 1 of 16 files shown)


src/utils/__tests__/fixtures/incident.json
```diff
@@ -0,0 +1,89 @@
+{
```
jasperswart: goed idee

PR #377, opened 2019-10-08
(Fix) Filter state

janjaap:
This PR contains a fix for two issues. The first was that the filter tag labels weren't rendered in the `PageHeader` component. The second issue caused not showing the correct filter data in the filter form.

With regards to the second issue, the logic of displaying filtered issues and editing a filter is the following:
- Clicking on 'Toon filter' in the 'Mijn filters' overview will set the state values `activeFilter` as well as `editFilter` so that the filter overview is updated and the filter form will be populated with that filter's data
- Clicking on 'Wijzig filter' will only set the `editFilter` state value (`activeFilter` will remain untouched, regardless of its value) and shows the filter form with the filter's data
- Clicking on the `Filteren` button in the overview page header will always populate the filter form with the `editFilter` state value
- Clicking on `Nieuw filter` in the filter form will clear the `editFilter` state value and will leave the `activeFilter` state value untouched
- Clicking on `Annuleren` in the filter form will reset the `editFilter` state value and replace it with the state value of `activeFilter` (regardless of its value at that time)
- Clicking on 'Filteren' in the filter form will set both the `activeFilter` and `editFilter` state values 

Or, schematically:

| Action | activeFilter | editFilter |
| --- | :---: | :---: |
| show filter | filter | filter |
| edit filter | leave untouched | filter  |
| new filter | leave untouched | clear |
| cancel edit | leave untouched | activeFilter |
| apply edit | filter | filter |

Next to the aforementioned changes, this PR also contains a change in how pagination and column sorting in the incident overview page is handled. Before, all filter parameters were sent along with the dispatched action `requestIncidents`. This requirement lead to too many responsibilities inside the corresponding saga and forced components across the codebase to send the full payload to the action, which made them less flexible and more complex than necessary.


jasperswart:
het filteren werkt nu goed, maar als ik eerste een filter maak en dan een search naar id, dan heb ik een filter EN en search

jasperswart:
Screenshot 2019-10-11 at 09 25 05

janjaap:
@jasperswart Ouch, good catch

src/signals/incident-management/__tests__/selectors.test.js
```diff
@@ -49,7 +52,7 @@ const filters = [
   },
 ];
 
-describe('signals/incident-management/selectors', () => {
+describe.only('signals/incident-management/selectors', () => {
```
aditudorache: remove only

janjaap: Good catch

janjaap:
@aditudorache I fully agree. I'll be more vigilant next time.
Thanks for the review!

jasperswart:
het is bijna goed: als je nu op ID zoekt vanaf de detail pagina, dan doet hij niets. hij zou terug moeten gaan naar de lijst pagina en op id hebben gezocht

PR #380, opened 2019-10-08
SIG-1713 Textarea scroll lines fix

janjaap:
This PR fixes an issue in IE11 where, when the contents of a textarea was scrolled, it would leave a trail of horizontal lines in front of the content. The cause of this behaviour was the element's box-shadow.

In order to fix the problem, the `<textarea>` elements in the application have been replaced by styled components. This also goes for `<input type"text">` elements where applicable.

### DIFF
--- a/src/components/TextArea/__tests__/TextArea.test.js
+++ b/src/components/TextArea/__tests__/TextArea.test.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { withAppContext } from 'test/utils';
+
+import TextArea from '../';
+
+describe('components/TextArea', () => {
+  it('renders correctly', () => {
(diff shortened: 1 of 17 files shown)


janjaap:
@jasperswart That's exactly what I did; remove the drop shadow. Pattern lab indeed has the old styling that contains the dreadful drop shadow.

jasperswart:
I would expect 1 styling change: in global.scss with nasty !important and not complete refactoring of all these components.

janjaap:
@jasperswart Well, if we want to get rid of `amsterdam-stijl` and incorporate `amsterdam-styled-components` instead, we'd better not put styling in our global style sheet, but use style components instead, which I did. So I'm not inclined to agree with your suggestion 🤷🏻‍♂️

PR #401, opened 2019-10-21
[SIG-1823] Update other packages

janjaap:
This PR updates packages that could not logically be grouped.
Note that, because of the removal of `react-loadable`, changes to generator templates were necessary and because of a breaking change in one of the `@datapunt/asc-` packages, a change to the header CSS was required.


src/components/SiteHeader/index.js
```diff
@@ -195,7 +203,7 @@ const MenuItems = ({
 };
 
 export const SiteHeader = props => {
```
aditudorache: destructure parameter props ?

janjaap: Could do, but I'm passing all props to underlying child components as well and then it's easier to to desctructure, otherwise I find myself having to list all props that I need to pass for each component 🤷‍♂

PR #415, opened 2019-10-29
Sig 1836 voeg toe unit test voor default texts components

jasperswart:
In this PR:
- DefaultTextsAdmin container is now styled component
- SelectForm is now a functional and styled component
- DefaultTextsForm is now a functional and styled component
- added enzyme-less unit test for: DefaultTextsAdmin, SelectForm and DefaultTextsForm
- added data-testid property to form components to make testing possible 
- removed obsolete sass files
- updated coverage


src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -1,18 +1,26 @@
...
-import { Button, themeColor } from '@datapunt/asc-ui';
 
-import './style.scss';
+const StyledWrapper = styled.div`
+  margin-top: 33px;
+  width: 100%;
+`;
+
+const StyledColumn = styled(Column)`
+  display: block;
```
janjaap: You shouldn't be doing this; a `Column` component has flex-box layout which is dictated by its parent which it either a `Row` or a `Column` component. Making it behave as a block-level component breaks the ability to align the column.
Looking at the elements that are rendered by the `DefaultTextsForm` component, it becomes difficult to use `Column` component and have them line out to the grid, because the flex-box alignment is broken by the use of `<div>` and other elements that do not have flex-box layout.

You can do this:
```jsx
<Column span={8}>
  <Column span={2}></Column>
  <Column span={6}></Column>
</Column>
```

But not this:
```jsx
<Column span={8}>
  <Column span={2}></Column>
  <div>
    <Column span={6}></Column>
  </div>
</Column>
```

and expect the behaviour to be the same.

You could try this:
```jsx
<Column span={8}>
  <Column span={2}></Column>
  <div>
    <Row hasMargin={false}>
      <Column span={6}></Column>
    </Row>
  </div>
</Column>
```

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -21,232 +29,244 @@ const StyledButton = styled(Button)`
...
-          )}
-        />
-      </div>
-    );
-  }
-}
+  const subscribe = () => {
+    items.forEach((item, index) => {
+      form.get(item).valueChanges.subscribe(data => {
+        onSaveDefaultTextsItem({ index, data });
```
janjaap: Why is there both a `handleSubmit` handler and the handler on this line? Don't they do the same?

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.test.js
```diff
@@ -1,10 +1,155 @@
...
+    );
+
+    expect(queryByTestId('defaultTextFormForm')).not.toBeNull();
+
+    fields.forEach(i => {
+      expect(queryByTestId(`title${i}`)).not.toBeNull();
+      expect(queryByTestId(`text${i}`)).not.toBeNull();
+    });
+
+    expect(queryByTestId('defaultTextFormItemButton0Up')).toBeDisabled();
```
janjaap: I'm pretty sure you can group a lot of these assertions by using a query selector, like so:
```jsx
container.querySelectorAll('button:nth-of-type(n+2)').forEach(button => {
  expect(button).toBeDisabled();
});
```
See https://alistapart.com/article/quantity-queries-for-css/ for reference

src/signals/incident-management/containers/DefaultTextsAdmin/components/SelectForm/index.js
```diff
@@ -1,126 +1,114 @@
...
+    }
+  }, []);
+
+  useEffect(() => {
+    subs = subCategories;
+    form.updateValueAndValidity();
+  }, [subCategories]);
+
+  return (
+    <Fragment>
```
janjaap: Fragment isn't needed in this case

src/signals/incident-management/containers/DefaultTextsAdmin/index.js
```diff
@@ -44,12 +45,14 @@ const DefaultTextsAdmin = ({
   },
 }) => (
   <Fragment>
```
janjaap: Fragment isn't needed in this case

src/signals/incident-management/containers/DefaultTextsAdmin/index.js
```diff
@@ -44,12 +45,14 @@ const DefaultTextsAdmin = ({
...
-    <StyledPageHeader title="Beheer standaard teksten" />
     <Row>
+      <Column span={12}>
+        <StyledH1 $as="h1">Beheer standaard teksten</StyledH1>
```
janjaap: Isn't the default element an `<h1>` already?

src/signals/incident-management/containers/DefaultTextsAdmin/index.js
```diff
@@ -69,9 +72,23 @@ const DefaultTextsAdmin = ({
   </Fragment>
 );
 
+DefaultTextsAdmin.defaultProps = {
+  defaultTextsAdmin: {
+    defaultTexts: [],
+    defaultTextsOptionList: [],
+    categoryUrl: 'bla',
```
janjaap: Should this prop have the value 'bla' as default?

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
jasperswart: gedaan

src/signals/incident-management/components/CopyFileInput/index.js
```diff
@@ -21,7 +21,12 @@ export const CopyFileInput = props => {
         ))}
 
         <div className="copy-file-input__control invoer antwoord">
-          <input name="" id={`form${name}`} type="checkbox" {...handler('checkbox')} />
+          <input
+            name={name}
+            data-testid={name}
+            id={`form${name}`}
+            type="checkbox" {...handler('checkbox')}
```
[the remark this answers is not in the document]
jasperswart: gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
jasperswart: that is a good point we can remove that

jasperswart: I removed all related parts

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.test.js
jasperswart: I have created an alternative

src/signals/incident-management/containers/DefaultTextsAdmin/index.js
jasperswart: gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/index.js
jasperswart: gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/index.js
jasperswart: yes gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -15,11 +15,20 @@ import { ChevronDown, ChevronUp } from '@datapunt/asc-assets';
 
 const StyledWrapper = styled.div`
   margin-top: 33px;
-  width: 100%;
+  flex-basis: 100%;
```
janjaap: Only makes sense having this rule if `display: flex;` has been set too

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -188,8 +172,8 @@ const DefaultTextsForm =({
             />
 
             {items.map((item, index) => (
-              <Row key={item}>
-                <StyledColumn span={7}>
+              <div key={item}>
+                <StyledLeftColumn span={7}>
```
janjaap: Remove span attribute

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -203,8 +187,8 @@ const DefaultTextsForm =({
                     name={`text${index}`}
                     control={form.get(`${item}.text`)}
                   />
-                </StyledColumn>
-                <StyledColumn span={1}>
+                </StyledLeftColumn>
+                <StyledRightColumn span={1}>
```
janjaap: Remove span attribute

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -228,22 +212,20 @@ const DefaultTextsForm =({
...
 
-            <Row>
-              <StyledColumn span={12}>
-                <Button
-                  data-testid="defaultTextFormSubmitButton"
-                  variant="secondary"
-                  type="submit"
-                  disabled={invalid}
-                >
+            <div>
```
janjaap: Is wrapping the button in a div necessary? The preceding element is already a block-level element.

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
jasperswart: `the parent is a <Column> so flex-basis  is correct here`

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
jasperswart: gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
jasperswart: gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
jasperswart: gedaan

src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
janjaap: Yep, you're right.

PR #429, opened 2019-11-04
[SIG-1732] Mobile header fix

janjaap:
This PR fixes an issue where the site header would not be displayed correctly on smaller screens. This was due to a change in a component in the ASC-UI lib.

**Before**
Screenshot 2019-11-04 at 11 08 15

**After**
Screenshot 2019-11-04 at 11 07 36
Screenshot 2019-11-04 at 11 07 44

### DIFF
--- a/src/components/SiteHeader/index.js
+++ b/src/components/SiteHeader/index.js
@@ -90,6 +90,16 @@ const HeaderWrapper = styled.div`
       #header {
         position: static;
 
+        header {
+          height: 160px;
+        }
+
+        @media screen and (max-width: 539px) {
+          header {
+            height: 116px;
+          }
+        }
+
         &:after {
           max-width: 1400px;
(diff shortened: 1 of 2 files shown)


src/components/SiteHeader/index.js
```diff
@@ -90,6 +90,16 @@ const HeaderWrapper = styled.div`
       #header {
         position: static;
 
+        header {
+          height: 160px;
+        }
+
+        @media screen and (max-width: 539px) {
```
spaces-ga: Do we have breakpoint variables for this?

janjaap: No, there aren't any defined for the whole application. Most breakpoints are defined by `@datapunt/asc-ui`'s theme configuration. Some others, where applicable, are defined on component level. Most of the time, that makes most sense.

PR #434, opened 2019-11-05
[SIG-1809] - Update margin history item.

spaces-ga:
Replaces styled divs for styled components. Set history item margins to margin-top: 0 and margin-bottom: 16px


spaces-ga:
Merged with https://github.com/Amsterdam/signals-frontend/pull/436

PR #441, opened 2019-11-06
[SIG-1877] pagination

janjaap:
This PR adds pagination functionality to the users overview page. In order to get it to work,
- the `<List>` component is made simpler so that it's not responsible for navigation actions
- the `<Pager>` component is now able to set the number of items per page
- routes for the `settings` module got a redirect from `/instellingen/gebruikers/` to `/instellingen/gebruikers/page/1`
- the `useFetchUsers` hook is changed so that it can handle the `page_size` param
- the `UsersOverview` container component is set to handle both location and page changes as well navigation actions to user detail pages


src/signals/settings/users/containers/Overview/index.js
```diff
@@ -8,10 +9,51 @@ import Pager from 'components/Pager';
...
+    const { dataset } = e.currentTarget;
+    const { itemId } = dataset;
+
+    if (itemId) {
+      history.push(routes.user.replace(':userId', itemId));
+    }
+  };
 
   return (
     <div className="users-overview-page">
```
jasperswart: the classname is not needed

src/signals/settings/routes.js
```diff
@@ -0,0 +1,7 @@
+const routes = {
+  users: `/instellingen/gebruikers`,
+  usersPaged: `/instellingen/gebruikers/page/:pageNum`,
```
spaces-ga: Could probably be optional 
```js
usersPaged: `/instellingen/gebruikers/page/:pageNum?
```

leaving `pageNum` out will break the route now

src/components/Pager/index.js
```diff
@@ -19,7 +19,7 @@ class Pager extends React.Component {
 
   render() {
     const currentPage = this.props.page;
-    const totalPages = Math.floor(this.props.itemCount / 100) + 1;
+    const totalPages = Math.floor(this.props.itemCount / this.props.pageSize) + 1;
     const hasPrevious = currentPage > 1;
```
spaces-ga: There are some issues with pagination. Pagination is also already done by backend. They return a 'count' and an previous and next link. We could simply use those values to determine what is needed on the frontend. Probably best to have an offline discussion about this.

src/signals/settings/users/containers/Overview/index.js
spaces-ga: This comment is about the next line
```js
<PageHeader title={`Gebruikers (${users.length})`} />
```

If we don't wait until loading is finished this titel will flash from `0` to actual number

I would also like to suggest showing `30 of 70` instead of just the number of items on the current page. Maybe we need to discuss this with UX

janjaap:
This PR has become too big; closing in favour of multiple smaller PRs

PR #446, opened 2019-11-11
[SIG-1564] - Add minimum file size check to image upload

spaces-ga:
This PR adds a minimum file size check to the `FileInput` component

Encountered some out of scope issues and created a new ticket on the backlog https://datapunt.atlassian.net/browse/SIG-1903


src/signals/incident/components/form/FileInput/index.js
```diff
@@ -69,7 +71,9 @@ const FileInput = ({
     }
   };
 
-  const checkFileSize = file => file.size <= meta.maxFileSize;
+  const checkMinFileSize = file => file.size >= meta.minFileSize;
+
+  const checkMaxFileSize = file => file.size < meta.maxFileSize;
```
spaces-ga: NOTE: Updated this check. It used to test `file.size <= meta.maxFileSize` but that would allow files to be equal to maxFileSize while `/src/signals/incident/services/custom-validators/index.js` does not allow for that.

PR #447, opened 2019-11-11
[SIG-1902] Supporting components

janjaap:
This PR adds the following components:
- ### BackLink
  _Component with customisable URL and content_

  Screenshot 2019-11-12 at 08 06 09
  
  Screenshot 2019-11-12 at 08 49 11

- ### FormAlert
  _Notification block that shows a title and, optionally, a message_

  FormAlert, just title

  FormAlert, title and message

  FormAlert, notification style, just title

  FormAlert, notification style, just title

- ### FormFooter
  _Absolute positioned container that can show a cancel button, a reset button and/or a submit button_

FormAlert, notification style, title and message

- ### Input
  _Implementation of the `@datapunt/asc-ui` `<Input />` component with the addition of a hint text_

  Screenshot 2019-11-12 at 09 07 39

### In addition
It also updates the packages `@testing-library/jest-dom` and `@testing-library/react` to their latest stable versions.


src/components/BackLink/index.js
```diff
@@ -0,0 +1,75 @@
...
+  </StyledLink>
+);
+
+BackLink.defaultProps = {
+  className: '',
+};
+
+BackLink.propTypes = {
+  /** The BackLink label */
+  children: PropTypes.node.isRequired,
```
jasperswart: that is incorrect comment isn't it?

janjaap: The `children` prop can be considered the component's label, but I guess the term `label` is a bit ambiguous. I'll change it.

src/components/BackLink/__tests__/Backlink.test.js
```diff
@@ -0,0 +1,20 @@
...
+    const { container, getByText } = render(
+      withAppContext(<BackLink to={to}>{label}</BackLink>)
+    );
+
+    expect(getByText(label)).toBeInTheDocument();
+
+    const link = container.querySelector('a');
+    expect(link).toBeInTheDocument();
+    expect(link.getAttribute('href')).toEqual(to);
+  });
```
jasperswart: there is a test missing for clicking the link

src/components/FormAlert/index.js
```diff
@@ -0,0 +1,70 @@
...
+      margin-bottom: 0;
+    `}
+`;
+
+/**
+ * Component that will display a title and, optionally, a message block. Should be used
+ * for generic feedback that is not related to individual fields.
+ */
+const FormAlert = ({ className, message, title, isNotification }) => (
+  <Div className={className} isNotification={isNotification}>
```
jasperswart: this should be called Alert I think there is also FormError component in DS.

it should have 4 variants:
[https://designsystem.amsterdam.nl/7awj1hc9f/p/901ede ](url)

I don't see those

janjaap: What should be called `Alert`? The Component that is the default export or the `<Div>` component on line 45?

This particular component is not meant as a generic notification component, but is specific to the context of a form. That is why this `FormAlert` component doesn't have the four different states that are in the design system. It is not necessary for the current scope. If needed, we can change the component at a later date, but for now, I'm not going to bother investing time in functionality that we don't yet need.

src/components/BackLink/__tests__/Backlink.test.js
janjaap: What should the test verify in your opinion? The element is an anchor; there is no need that the browser does what it is supposed to be doing.

src/components/Input/index.js
```diff
@@ -0,0 +1,62 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
```
jasperswart: I would wait with this implementation of this component until we have a new form lib.
we should create a wrapper component for all types of form components with a label and a subtitle and the field.
but if it is needed now just add it.

src/components/Input/index.js
```diff
@@ -0,0 +1,62 @@
...
+import styled, { css } from 'styled-components';
+import {
+  Input as Field,
+  Label as FieldLabel,
+  styles,
+  themeColor,
+  Typography,
+  themeSpacing,
+} from '@datapunt/asc-ui';
```
jasperswart: I miss a input placeholder here

janjaap: Not sure what you mean. There is no `Placeholder` component in `@datapunt/asc-ui`. If you mean setting a placeholder on the input element, that is already possible by passing that prop, like so
```jsx
  <Input placeholder="Foo bar baz" />
```

Line 44 takes care of that.

src/components/Input/index.js
janjaap: We're going to need the `asc-ui` implementation anyway, regardless of which lib we choose.

src/components/Input/index.js
jasperswart: yes true but I the test is still <Input placeholder="Foo bar baz" /> missing

src/components/BackLink/__tests__/Backlink.test.js
jasperswart: the link should link to the right place that is part of the component

jasperswart: we should have a talk about the with Carlos about what to test and what not in general

src/components/FormAlert/index.js
janjaap: IMHO, `Notification` is too generic a name for a component that is meant to show feedback related to actions performed in forms.

PR #452, opened 2019-11-14
[SIG-1914] Alert on incident patch 403

janjaap:
This PR shows an alert on the incident detail page whenever a user encounters an error with the  403 error status.

### DIFF
--- a/internals/testing/test-bundler.js
+++ b/internals/testing/test-bundler.js
@@ -13,6 +13,7 @@ import Adapter from 'enzyme-adapter-react-16';
 Enzyme.configure({ adapter: new Adapter() });
 
 global.window.L = L;
+global.window.alert = msg => msg;
 
 /**
  * Element.closest() polyfill
--- a/src/models/incident/saga.js
+++ b/src/models/incident/saga.js
@@ -49,6 +49,13 @@ export function* patchIncident(action) {
     yield put(patchIncidentSuccess({ type: payload.type, incident }));
     yield put(requestHistoryList(payload.id));
   } catch (error) {
+    const { response } = error;
+
+    if (response.status === 403) {
+      const { jsonBody } = response;
+      global.alert(jsonBody.detail);
+    }
+
     yield put(patchIncidentError({ type: payload.type, error }));
   }
 }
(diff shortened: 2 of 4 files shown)


internals/testing/test-bundler.js
```diff
@@ -13,6 +13,7 @@ import Adapter from 'enzyme-adapter-react-16';
 Enzyme.configure({ adapter: new Adapter() });
 
 global.window.L = L;
+global.window.alert = msg => msg;
```
spaces-ga: Do we really need this? Just to test a request value?

src/models/incident/saga.js
```diff
@@ -49,6 +49,13 @@ export function* patchIncident(action) {
     yield put(patchIncidentSuccess({ type: payload.type, incident }));
     yield put(requestHistoryList(payload.id));
   } catch (error) {
+    const { response } = error;
+
+    if (response.status === 403) {
+      const { jsonBody } = response;
+      global.alert(jsonBody.detail);
```
spaces-ga: Would leave this out

src/models/incident/saga.test.js
```diff
@@ -135,11 +135,33 @@ describe('models/incident/saga', () => {
...
+      await expectSaga(patchIncident, action)
         .provide([[matchers.call.fn(authPatchCall), throwError(error)]])
         .put(patchIncidentError({ type, error }))
         .silentRun();
+
+      expect(global.alert).not.toHaveBeenCalled();
+
+      error.response.status = 403;
+
+      await expectSaga(patchIncident, action)
```
spaces-ga: Can't we inspect the return values here instead of spying on a global function that we added just for this purpose?

internals/testing/test-bundler.js
janjaap: Yes we do; JSDOM doesn't have the `alert` method implemented.

src/models/incident/saga.js
janjaap: Why, exactly? Showing an alert window has been discussed with UX and has been approved.

src/models/incident/saga.test.js
janjaap: But the return value doesn't contain anything I need. I just need to know that the alert method has been called.

PR #461, opened 2019-11-20
(chore) New pagination

janjaap:
This PR replaces the `Pager` component with the `Pagination` component.

I've merged `develop` into this branch and, because this PR's base branch (`SIG-1778_users-overview`) didn't have that merge yet, the number of changed files is huge. However, the only commits that matter and that need review, are the four latest commits in this branch.


src/signals/settings/routes.js
```diff
@@ -1,7 +1,7 @@
 const routes = {
```
spaces-ga: could we add a `export const baseUrl = `/instellingen/gebruikers`
and use this in the `routes` object

Then you could also import this base url when you want to push location to users instead of replacing with regex.

```js
import { baseUrl as usersRoute } from './routes'

history.push(`${usersRoute}/page/${page}`));
history.push(`${usersRoute}/${userId}`));
```

src/signals/settings/__tests__/index.test.js
```diff
@@ -0,0 +1,27 @@
...
+import { render } from '@testing-library/react';
+import { withAppContext, withIntlAppContext } from 'test/utils';
+import { isAuthenticated } from 'shared/services/auth/auth';
+
+import SettingsModule from '..';
+import translations from '../../../translations/nl.json';
+
+jest.mock('shared/services/auth/auth');
+
+describe('signals/settings', () => {
```
spaces-ga: I am not sure any of this logic should be part of the settings module...

src/signals/settings/routes.js
```diff
@@ -1,7 +1,7 @@
...
   users: '/instellingen/gebruikers',
   usersPaged: '/instellingen/gebruikers/page/:pageNum(\\d+)',
-  user: '/instellingen/gebruikers/:userId',
+  user: '/instellingen/gebruikers/:userId(\\d+)',
```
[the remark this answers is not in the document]
janjaap: Ah yes, good one! Will do.

src/signals/settings/routes.js
janjaap: Ah yes, I thought about doing that, but somehow I didn't get around to doing that. Will do.

src/signals/settings/routes.js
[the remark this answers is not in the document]
jasperswart: ik zie helemaal geen test voor de routes maar dat is al gezegd

PR #468, opened 2019-11-22
(chore) Error messages

janjaap:
This PR adds a function to the API service that returns a message related to a specific error status. If the status code is isn't registered, a default message is returned.

### DIFF
--- a/src/shared/services/api/api.test.js
+++ b/src/shared/services/api/api.test.js
@@ -7,7 +7,7 @@ import {
   authDeleteCall,
   authPatchCall,
   authPostCall,
-  defaultErrorMessage,
+  errorMessageDictionary,
   generateParams,
   postCall,
   getErrorMessage,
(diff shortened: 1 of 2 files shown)


src/shared/services/api/api.test.js
```diff
@@ -165,4 +167,19 @@ describe('api service', () => {
...
+    it('returns a default error message', () => {
+      expect(getErrorMessage({ status: 415 })).toEqual(defaultErrorMessage);
+      expect(getErrorMessage({ status: 'foo bar' })).toEqual(defaultErrorMessage);
+    });
+
+    it('returns a specific error message', () => {
+      const statuses = [401, 403, 408, 413, 429, 500, 503];
+
+      statuses.forEach(status => {
+        expect(getErrorMessage({ status })).not.toEqual(defaultErrorMessage);
```
spaces-ga: Could we also check that they **DO** receive the correct error message. Not just **NOT** default

janjaap: Done.

PR #484, opened 2019-11-29
[SIG-1972] GlobalError container refactor

janjaap:
This PR contains the refactoring of the `GlobalError` container. Since the component was repurposed, its name was also changed and has been moved from the `containers` folder to the `components` folder.

The new component can render three different types of notifications (see https://projects.invisionapp.com/d/main/#/console/18007596/393242628/inspect for reference) that can have a title and an optional description message. The component disappears automatically after eight seconds (it stays on screen when it's an alert instead of a notice or success message) or when a navigation action occurs or when the close button is clicked. Note that the eight second timer is reset every time the component receives a `mouseWithin` event.


janjaap:
Closing this PR; it has since been split into several smaller PRs

PR #487, opened 2019-12-03
(chore) Notification component

janjaap:
This PR adds a global notification component according to the specs in Invision. The component can render three different variants: error, success and notice and looks (for the success variant) like this:
Screenshot 2019-12-03 at 17 19 39

Timers runs to hide the component automatically after eight seconds unless the mouse pointer is hovered over it. Exceeding the timeout or clicking the close button will slide the component up out of view. 

Apart from the three variants, there are two types (global and local) that determine if the component should be hidden automatically (local) or not (global).

A follow-up pull request will replace the `GlobalError` container with the `Notification` component (and corresponding container).


src/components/Notification/__tests__/Notification.test.js
```diff
@@ -0,0 +1,359 @@
...
+    expect(onClose).toHaveBeenCalled();
+
+    expect(getSetTimeoutCalls()).toHaveLength(2);
+    expect(getClearTimeoutCalls()).toHaveLength(2);
+
+    expect(container.firstChild.classList.contains('slideup')).toEqual(true);
+  });
+
+  it('clears timeouts on unmount', () => {
+    expect(getSetTimeoutCalls()).toHaveLength(0);
```
spaces-ga: Could we move this to after the `act(runAllTimers)` ?

src/components/Notification/__tests__/Notification.test.js
```diff
@@ -0,0 +1,359 @@
...
+    act(() => {
+      fireEvent.mouseOver(getByTestId('notification'));
+    });
+
+    act(() => {
+      jest.runAllTimers();
+    });
+
+    expect(getSetTimeoutCalls()).toHaveLength(2);
+    expect(getClearTimeoutCalls()).toHaveLength(4); // clearTimeout calls on update and unmount
```
spaces-ga: Could we split this in 2 expect lines. 1 here and one between the `mouseOver()` and `runAllTimers()`

src/components/Notification/__tests__/Notification.test.js
```diff
@@ -0,0 +1,359 @@
...
+    expect(onClose).not.toHaveBeenCalled();
+  });
+
+  it('restarts the timer when the component receives a mouse out event', () => {
```
spaces-ga: Could we test for `set` and `clearTimeouts`. And in both cases advance the timers by `ONCLOSE_TIMEOUT + SLIDEUP_TIMEOUT` ?

src/components/Notification/index.js
```diff
@@ -0,0 +1,259 @@
...
+
+    if (hasFocus) {
+      global.clearTimeout(onCloseTimeoutRef.current);
+      global.clearTimeout(slideUpTimeoutRef.current);
+    } else {
+      const slideUpTimeoutId = global.setTimeout(() => {
+        global.clearTimeout(slideUpTimeoutRef.current);
+
+        setSlideUp(true);
+      }, ONCLOSE_TIMEOUT);
```
spaces-ga: I have the feeling these 2 names are mixed...

src/components/Notification/__tests__/Notification.test.js
janjaap: I'll remove the check for `getSetTimeoutCalls`,  because I don't need to be checking for that. Just for `clearTimeout` calls.

src/components/Notification/__tests__/Notification.test.js
janjaap: I'm going to get rid of the timer checks and going to replace them with a check on the `onClose` function, because that is what I actually should be checking for.

src/components/Notification/index.js
janjaap: Yeah, you're right. I've swapped the names.

PR #499, opened 2019-12-06
(chore) Change FormFooter render requirements

janjaap:
This PR alters the `FormFooter` component so that callback function props are no longer required to render any of the buttons. Callback functions can programmatically be set later or by setting a form
s `onReset` or `onSubmit` attributes.


janjaap:
It doesn't; nothing changed in appearance.

PR #505, opened 2019-12-09
Create actions config

janjaap:
This PR adds a Github actions config file that runs `lint`, `test` and `build` in latest Ubuntu on node `v10`.


action.Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM node:10-stretch AS builder
+LABEL maintainer="datapunt@amsterdam.nl"
+
+WORKDIR /app
+
+# Run updates and cleanup
+RUN apt-get update && apt-get install -y \
+      netcat \
```
spaces-ga: Why do we need netcat?

action.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+      git && \
```
spaces-ga: could we put the && on the next line?

```sh
git \
&& rm -rf /var/lib/apt/lists/*
```

action.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+      rm -rf /var/lib/apt/lists/*
+
+#  Changing git URL because network is blocking git protocol...
+RUN git config --global url."https://".insteadOf git://
+RUN git config --global url."https://github.com/".insteadOf git@github.com:
+
+COPY internals /app/internals
+COPY server /app/server
+COPY .gitignore \
```
spaces-ga: Why are we copying all these git files?
Doesn't seem like git is used in this container.

PR #506, opened 2019-12-09
Sig 1920 address format in filter tag list

jasperswart:
In this PR:
- fixed issue with incorrect formatting of address in FilterTagList
- removed support for ignoredTags because it is not needed anymore

Please note: we cannot use Date.parse('weesperstraat 113') to detect valid dates.
it will incorrectly return true.


src/signals/incident-management/containers/FilterTagList/__tests__/FilterTagList.test.js
```diff
@@ -126,7 +110,28 @@ describe('signals/incident-management/containers/FilterTagList', () => {
         ),
       );
 
-      expect(container.querySelectorAll('span')).toHaveLength(6);
+      expect(queryByText('Normaal')).toBeTruthy();
```
janjaap: Using `.toBeInTheDocument()` has the same result, but is more descriptive. I would up opt for using that.

src/signals/incident-management/containers/FilterTagList/__tests__/FilterTagList.test.js
```diff
@@ -126,7 +110,28 @@ describe('signals/incident-management/containers/FilterTagList', () => {
...
+      expect(queryByText('Tevreden')).toBeTruthy();
+      expect(queryByText('februariplein 1')).toBeTruthy();
+      expect(queryByText('Centrum')).toBeTruthy();
+      expect(queryByText('Westpoort')).toBeTruthy();
+      expect(queryByText('Telefoon – Adoptant')).toBeTruthy();
+      expect(queryByText('Telefoon – ASC')).toBeTruthy();
+
+      expect(container.querySelectorAll('span')).toHaveLength(10);
```
janjaap: Why would you check for `<span>` elements to be in the document? Wouldn't it make more sense to look for specific test ids?

src/signals/incident-management/containers/FilterTagList/index.js
```diff
@@ -31,11 +28,11 @@ const renderTag = (key, tagKey, mainCategories, list) => {
 
   let display = (found && found.value) || key;
 
-  if (!display || ignoredTags.includes(tagKey)) {
+  if (!display) {
     return;
   }
 
-  if (isDate(display)) {
+  if (tagKey === 'incident_date') {
```
janjaap: There will be other fields values in the future that will contain dates, take for instance the proposed change to the date field where a range is possible. You should still be checking on the contents of a field otherwise we'll end up with a whitelist that we'll actively have to maintain, which might lead to confusing and unexpected visual presentations of the tags list.

src/signals/incident-management/containers/FilterTagList/index.js
```diff
@@ -64,10 +61,6 @@ export const FilterTagListComponent = props => {
     category_slug: sub,
   };
 
-  if (!tags) {
```
janjaap: Why remove this check? The component shouldn't render anything when there are no tags to render.

jasperswart: tags is in the defaultProps.

src/signals/incident-management/containers/FilterTagList/index.js
jasperswart: I agree, but isDate uses Date.parse('weesperstraat 113'). that will returns a correct date. but it is not a date. how to check for this then?

src/signals/incident-management/containers/FilterTagList/__tests__/FilterTagList.test.js
jasperswart: I agree

src/signals/incident-management/containers/FilterTagList/index.js
janjaap: I think the `isDate` function is too limited. We can replace it with [lodash.isdate](https://www.npmjs.com/package/lodash.isdate)

src/signals/incident-management/containers/FilterTagList/index.js
janjaap: It can still be bypassed. `defaultProps` Is merely a safeguard during development.

src/signals/incident-management/containers/FilterTagList/__tests__/FilterTagList.test.js
jasperswart: gedaan

src/signals/incident-management/containers/FilterTagList/index.js
jasperswart: de lodash isDate is niet geschikt omdat hij  alle string values afkeurt als datum.

volgens mij moeten we toch maar op key de datum formatten, dus dat ga ik er maar terug in bouwen

PR #509, opened 2019-12-10
[SIG-1818] 0 Meldingen notification

janjaap:
This PR:
- replaces the incidents list with a notification when no incidents are returned from the filter request
- makes sure that the `Pagination` component doesn't render any navigation buttons when there isn't more than one page to be displayed


janjaap:
Closing PR; merge gone wrong

PR #515, opened 2019-12-11
[SIG-1818] 0 Meldingen gevonden

janjaap:
This PR:

replaces the incidents list with a notification when no incidents are returned from the filter request
makes sure that the Pagination component doesn't render any navigation buttons when there isn't more than one page to be displayed

### DIFF
--- a/src/components/Pagination/index.js
+++ b/src/components/Pagination/index.js
@@ -58,6 +58,10 @@ const Pagination = ({
 }) => {
   const pagesInRange = pageNumbersList(currentPage, totalPages);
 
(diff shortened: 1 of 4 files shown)


src/components/Pagination/index.js
```diff
@@ -58,6 +58,10 @@ const Pagination = ({
 }) => {
   const pagesInRange = pageNumbersList(currentPage, totalPages);
 
+  if (totalPages <= 1) {
```
spaces-ga: Could probably move this to before `pageNumbersList` since you don't really need that. But that's just optimisation

PR #516, opened 2019-12-11
[SIG-1843] scroll to top overal in de app

jasperswart:
In this PR:
- added useEffect hook in app container that scrolls to top after each url change
- I have added the test as well

### DIFF
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -37,7 +37,7 @@ export const AppContainer = ({ requestCategoriesAction }) => {
   useEffect(() => {
     const unlisten = history.listen(() => {
       defer(() => {
-        window.scrollTo({
+        global.window.scrollTo({
           top: 0,
           left: 0,
         });
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -40,7 +40,6 @@ export const AppContainer = ({ requestCategoriesAction }) => {
         window.scrollTo({
           top: 0,
           left: 0,
-          behavior: 'smooth',
         });
       });
     });
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -47,7 +47,7 @@ export const AppContainer = ({ requestCategoriesAction }) => {
     return () => {
       unlisten();
     };
-  }, [history]);
+  }, [history.push]);
 
   return (
     <ThemeProvider>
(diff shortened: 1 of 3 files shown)


src/containers/App/index.js
```diff
@@ -27,11 +28,27 @@ import { requestCategories } from './actions';
...
   authenticate();
+  const history = useHistory();
 
   useEffect(() => {
     requestCategoriesAction();
   }, []);
 
+  useEffect(() => {
+    const unlisten = history.listen(() => {
+      defer(() => {
```
janjaap: Why would you defer the `scrollTo`?

janjaap: Also, did you test this with both `PUSH` and `POP` actions? It might not be necessary on `POP`, because browsers usually restore scroll position on that action.

jasperswart: that is because of the use  of defer in the wizard form component: 
that is needed for correct rendering of visible and not visible questions

it will bite you in the ass

janjaap: How exactly will it bite me in the ass?

jasperswart: alleen de wizard form heeft de defer nodig.  vreemd. wat zal ik doen?

janjaap: I guess you need to find out why navigating the wizard form requires us to defer the scrolling of the page.

src/containers/App/index.test.js
```diff
@@ -1,13 +1,30 @@
...
 import { REQUEST_CATEGORIES } from './constants';
 
 jest.mock('components/MapInteractive');
 
+jest.useFakeTimers();
+
 describe('<App />', () => {
+  let listenSpy;
+  let spyScrollTo;
+  const history = createMemoryHistory();
```
janjaap: you can use the `history` object provided by the test utils

src/containers/App/index.test.js
```diff
@@ -16,9 +33,31 @@ describe('<App />', () => {
...
+      withAppContext(<AppContainer requestCategoriesAction={() => { }} />),
+    );
+
+    expect(spyScrollTo).not.toHaveBeenCalled();
+
+    act(() => {
+      history.push('/somewhere/else');
+    });
+
+    act(() => {
```
janjaap: you should test that `spyScrollTo` has not been called after pushing

src/containers/App/index.test.js
jasperswart: die zit niet in deze branch  maar wel gedaan

src/containers/App/index.test.js
jasperswart: gedaan

PR #517, opened 2019-12-11
[SIG-1973] Saga exception handling

janjaap:
This PR attaches an exception handler to any saga that is injected into a container component. This allows us to tap into any error that is rethrown from a saga, like so:

```jsx
export function* sagaGeneratorFunc() {
  try {
    // attempt to fetch from a resource
  } catch (error) {
    // do whatever needs to be done when an error is caught

    // important bit
    throw(error);
  }
}
```

or so:

```jsx
export function* sagaGeneratorFunc() {
  throw(error);
}
```


jest.config.js
```diff
@@ -1,10 +1,13 @@
 module.exports = {
   collectCoverageFrom: [
+    'src/*.js',
     'src/**/*.{js,jsx}',
+    '!src/*.test.js',
```
spaces-ga: Is this not already covered by the line beneath?

src/configureStore.test.js
```diff
@@ -29,6 +28,32 @@ describe('configureStore', () => {
...
+      }
+
+      expect(dispatchSpy).not.toHaveBeenCalled();
+
+      store.runSaga(someGenerator);
+
+      expect(dispatchSpy).toHaveBeenCalledWith(
+        expect.objectContaining({
+          payload: {
+            message: expect.any(String),
```
spaces-ga: We know the message and the string at this point right?

jest.config.js
janjaap: No, `configureStore.js`, for instance, didn't show up in the coverage report.

src/configureStore.test.js
janjaap: Yes, but for the test it doesn't really matter which message has been sent, just that a message has been sent.

PR #519, opened 2019-12-11
[SIG-1040] Vrij zoeken

janjaap:
This feature PR contains:
- #511 - [SIG-1815] search endpoint
- #515 - [SIG-1818] 0 Meldingen gevonden
- #552 - [SIG-1816] - Error handling for corrupted elastic index


src/signals/incident-management/containers/IncidentOverviewPage/reducer.js
```diff
@@ -35,6 +38,16 @@ export default (state = initialState, action) => {
...
 
+    case SEARCH_INCIDENTS:
+      return state
+        .set('loading', true)
+        .set('error', false)
+        .set('errorMessage', undefined)
+        .set('searchQuery', action.payload);
+
+    case RESET_SEARCH_INCIDENTS:
+      return state.set('searchQuery', '');
```
spaces-ga: Should this not also set `loading` back to `false`?

spaces-ga: Or does this not cancel any requests?

src/signals/incident-management/containers/IncidentOverviewPage/saga.js
```diff
@@ -61,6 +62,32 @@ export function* fetchIncidents() {
...
+    yield put(requestIncidentsSuccess(incidents));
+  } catch (error) {
+    if (
+      error.response &&
+      error.response.status === 500
+    ) {
+      // Getting an error response with status code 500 from the search endpoint
+      // means that the Elasticsearch index is very likely corrupted. In that
+      // case we simulate a success response without results.
+      yield put(requestIncidentsSuccess({ count: 0, results: [] }));
```
spaces-ga: Does this mean we yield a success first and then an error on line 87?

src/containers/SearchBar/index.js
```diff
@@ -24,19 +23,18 @@ export const SearchBarComponent = ({
...
-    onRequestIncidents();
-  };
+    onSearchIncidents(searchInput);
+  }, [history, onApplyFilter, onSearchIncidents]);
 
-  const onChange = value => {
+  const onChange = useCallback(value => {
     if (value === '') {
-      onSetSearchQuery('');
+      onSearchIncidents('');
```
spaces-ga: Will this fire off an empty search request right before a regular fetchIncidents?

src/signals/incident-management/containers/IncidentOverviewPage/reducer.js
janjaap: Doesn't cancel any requests, no. It clears a part of the state that is used in, for instance, the `PageHeader` container component to show what the user searched for. Setting it to an empty string will clear the `PageHeader` title.

src/signals/incident-management/containers/IncidentOverviewPage/saga.js
janjaap: Yes, for administrative purposes. So that, when we're debugging we can still see that the request was erroneous. The user shouldn't see an error.

src/containers/SearchBar/index.js
janjaap: Hm, yeah it does. I'll change this. Good catch.

PR #531, opened 2019-12-17
[SIG-2032] fixed all subtitles keys

jasperswart:
all subheader keys changed into subtitle


janjaap:
@jasperswart Would be nice to have tests that validate the changes in this PR.

PR #542, opened 2019-12-18
(fix) Clear filter

janjaap:
This PR fixes an issue where, when the filter form was reset, the global store wasn't properly updated and, as a result, rendered the `FilterForm` component with the wrong data.

In addition, this PR:
- contains fixes for tests for the `FilterForm` component
- wraps functions with `useCallback`
- fixes another issue with the `name` field where its contents couldn't be altered after resetting the form (replacing `defaultValue` with `value`)
- makes specific props required instead of leaving them optional

### DIFF
--- a/src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
+++ b/src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
@@ -506,7 +506,8 @@ describe('signals/incident-management/components/FilterForm', () => {
       const { container } = render(
         withAppContext(
           <FilterForm
-            {...formProps, ...handlers}
+            {...formProps}
+            {...handlers}
             filter={{
               name: '',
               options: { incident_date: '1970-01-01' },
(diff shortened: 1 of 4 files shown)


src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
```diff
@@ -522,7 +506,7 @@ describe('signals/incident-management/components/FilterForm', () => {
       const { container } = render(
         withAppContext(
           <FilterForm
-            categories={categories}
+            {...formProps}
```
spaces-ga: could we include handlers here for clarity? 
`{...formProps, ...handlers}`

src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
```diff
@@ -549,7 +545,7 @@ describe('signals/incident-management/components/FilterForm', () => {
...
             {...handlers}
```
spaces-ga: See previous comment

src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -40,128 +40,133 @@ const FilterForm = ({
...
-  };
+    }));
+  }, []);
+
+  const updateFilterDate = useCallback(
+    /* istanbul ignore next */ (prop, dateValue) => {
+      setFilterData(state => ({
+        ...state,
+        options: {
+          ...state.options,
```
spaces-ga: Does `options` always exist?

PR #543, opened 2019-12-18
[SIG-2038] Date field helper components

janjaap:
This PR adds the `CalendarInput` and `CustomInput` components. They both are needed to render a date input field like so:
Screenshot 2019-12-19 at 07 30 33
instead of:
Screenshot 2019-12-19 at 07 31 00

### DIFF
--- a/src/signals/incident-management/components/CalendarInput/index.js
+++ b/src/signals/incident-management/components/CalendarInput/index.js
@@ -1,0 +1,51 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import DatePicker from 'react-datepicker';
+import moment from 'moment';
+import CustomInput from './CustomInput';
+
+const CalendarInput = ({ id, label, name, onSelect, selectedDate }) => (
+  <Fragment>
+    <DatePicker
+      autoComplete="off"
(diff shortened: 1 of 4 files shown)


src/signals/incident-management/components/CalendarInput/index.js
```diff
@@ -0,0 +1,51 @@
...
+import moment from 'moment';
+import CustomInput from './CustomInput';
+
+const CalendarInput = ({ id, label, name, onSelect, selectedDate }) => (
+  <Fragment>
+    <DatePicker
+      autoComplete="off"
+      customInput={<CustomInput label={label} id={id} />}
+      dateFormat="DD-MM-YYYY"
+      id={id}
```
spaces-ga: Will these components have the same `id`?

janjaap: Good catch. The `DatePicker` component doesn't do anything with the `id` prop. I'll remove it.

spaces-ga: It does do something? :)

janjaap: Well, I removed it and then realised I use it in the tests. So I added it back
🤷‍♂

PR #544, opened 2019-12-19
[SIG-2004] Date fields

janjaap:
This PR adds an extra date field to the `Filter` form component. The initial field, which was tied to `incident_date` has been replaced by `created_after` and `created_before` filter fields.
Since the date fields should only show dates (no times) and the API expects and returns a string that contains both date and time, incoming and outgoing data are converted to match those requirements.

The `FilterTagList` container needed to combine the possible selected dates into one tag element, so that it can display three different date range options, like so:
- `Datum: 09-12-2019 t/m nu`
- `Datum: 09-12-2019 t/m 24-12-2019`
- `Datum: t/m 24-12-2019`

### DIFF
--- a/src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
+++ b/src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
@@ -528,8 +528,7 @@ describe('signals/incident-management/components/FilterForm', () => {
       const { container } = render(
         withAppContext(
           <FilterForm
-            {...formProps}
-            {...handlers}
+            {...{...formProps, ...handlers}}
             filter={{
               name: '',
               options: { incident_date: '1970-01-01' },
@@ -567,8 +566,7 @@ describe('signals/incident-management/components/FilterForm', () => {
       const { container } = render(
         withAppContext(
           <FilterForm
-            {...formProps}
-            {...handlers}
+            {...{...formProps, ...handlers}}
             filter={{
               name: 'My filter',
               options: {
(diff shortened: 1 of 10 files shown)


src/signals/incident-management/components/FilterForm/__test__/FilterForm.test.js
```diff
@@ -533,10 +538,22 @@ describe('signals/incident-management/components/FilterForm', () => {
...
       fireEvent.click(container.querySelector('button[type="submit"]'));
 
-      expect(handlers.onSaveFilter).not.toHaveBeenCalled();
       expect(handlers.onSubmit).toHaveBeenCalled();
+      expect(handlers.onSaveFilter).not.toHaveBeenCalled(); // name field is empty
+
+      fireEvent.change(nameField, { target: { value: 'New name' }});
+      fireEvent.click(container.querySelector('button[type="submit"]'));
+
+      expect(handlers.onSaveFilter).toHaveBeenCalled();
```
spaces-ga: Should expect `submitHandler` to have been called 1 more time?

src/signals/shared/filter/parse.js
```diff
@@ -64,6 +65,16 @@ export const parseOutputFormData = form => {
     name, refresh, id, ...options
   } = parsed;
 
+  // before/after params require a specific format that includes the time
+  // since `created_before` means before and including, we add a day
+  if (moment(options.created_before, 'YYYY-MM-DD').toISOString() !== null) {
+    options.created_before = moment(options.created_before).set({ hours: 23, minutes: 59, seconds: 59 }).format('YYYY-MM-DDTkk:mm:ss');
```
spaces-ga: Still not a fan of this.
Before should always mean the same. So if that mean we need to add or remove a day for displaying purposes, we should be doing that in my opinion.

PR #551, opened 2019-12-24
[SIG-2047] Department list page

janjaap:
This PR adds the department overview page. The list of departments is requested from the API and then stored in the global store. Since more pages in the settings module need access to the departments data,  the data retrieval responsibility has been moved to the settings module itself. That also goes for roles and permissions. The `RolesListContainer` component has been simplified and corresponding tests have been removed.


src/models/departments/__tests__/reducer.test.js
```diff
@@ -0,0 +1,63 @@
+import { fromJS } from 'immutable';
+import departmentsJson from 'utils/__tests__/fixtures/departments.json';
+
+import reducer, { initialState } from '../reducer';
+import * as constants from '../constants';
+
+const intermediateState = fromJS({
+  count: 9,
+  error: false,
+  list: departmentsJson.results.slice(0, 9),
```
spaces-ga: Would suggest using a constant here to make sure the slice and the count use the same number

src/signals/settings/__tests__/index.test.js
```diff
@@ -23,6 +24,41 @@ describe('signals/settings', () => {
...
+    expect(typeof containerProps.onFetchDepartments).toEqual('function');
+
+    expect(containerProps.onFetchPermissions).not.toBeUndefined();
+    expect(typeof containerProps.onFetchPermissions).toEqual('function');
+
+    expect(containerProps.onFetchRoles).not.toBeUndefined();
+    expect(typeof containerProps.onFetchRoles).toEqual('function');
+  });
+
+  it('should initiate fetches on mount', () => {
```
spaces-ga: Performing async tasks here. `async ()`

src/signals/settings/__tests__/index.test.js
```diff
@@ -23,6 +24,41 @@ describe('signals/settings', () => {
...
+    render(
+      withAppContext(
+        <Module
+          onFetchDepartments={onFetchDepartments}
+          onFetchPermissions={onFetchPermissions}
+          onFetchRoles={onFetchRoles}
+        />
+      )
+    );
+
```
spaces-ga: Wait to finish async tasks: `await wait();`

src/models/departments/reducer.js
```diff
@@ -0,0 +1,39 @@
...
+      return state
+        .set('count', action.payload.count)
+        .set('error', false)
+        .set('errorMessage', '')
+        .set('list', fromJS(action.payload.results))
+        .set('loading', false);
+
+    case FETCH_DEPARTMENTS_ERROR:
+      return state
+        .set('count', undefined)
```
spaces-ga: Why update `count` if we don't update `list`? Those 2 will get out of sync on error.

src/signals/settings/__tests__/index.test.js
janjaap: Not necessarily; the actual fetch is async, but not the calls that initiate the fetch and that is what this particular test is about. The fetch itself is not the responsibility of the `settings` module, so that is why I didn't test it here.

src/models/departments/reducer.js
janjaap: True. `count` Doesn't have to be updated.

PR #571, opened 2020-01-14
Sig 1931 trigger tel and email software keyboard

jasperswart:
in this PR:
- in meldingsformulier added type of phone and email address fields
- meldingsformulier is now using Input component
- removed snapshot test from meldingsformulier TextInput
- tested in :

iPhone 13: safari
iPad (sim) 13: safari
iPhone 8 plus: safari

Android
Galaxy S10: chrome, firefox and samsung browser
Galaxy S9: chrome, firefox and samsung browser
Pixel 4: chrome, firefox
Pixel 2: chrome, firefox
Galaxy Note 10: chrome, firefox, samsung browser

Windows Phone: Lumia 930


src/signals/incident/definitions/wizard-step-3-telefoon.js
```diff
@@ -18,7 +18,7 @@ export default {
           subtitle: 'Zo kunt u ons helpen het probleem sneller of beter op te lossen.',
           path: 'reporter.phone',
           placeholder: 'Telefoonnummer',
-          type: 'text',
+          type: 'tel',
           autoRemove: /[^ ()\d+-]/g,
```
janjaap: What if someone would input their phone number like +31612345678? Would that result in `31612345678`?
The regex seems a bit strange to me; does it even do anything?

janjaap:
There might be other fields that could benefit form a change in type; I saw, for instance,
- `extra_container_number`
- `extra_personen_overig`
- `extra_straatverlichting_niet_op_kaart_nummer`
- `extra_klok_niet_op_kaart_nummer`
- `extra_verkeerslicht_nummer`

src/signals/incident/definitions/wizard-step-3-telefoon.js
jasperswart: on blur it will remove unwanted characters: "+" is allowed

janjaap: The `autoRemove` option doesn't seem to work. Can you confirm?

jasperswart: again: on blur the invalid characters are removed. please look at telefoonnummer step

https://acc.meldingen.amsterdam.nl/incident/telefoon

PR #572, opened 2020-01-14
(Chore) FF Instellingen fixes

janjaap:
This PR fixes an issue where, in Firefox 43, JSON responses from a RESTful API endpoint weren't parsed by the browser and resulted in empty pages or error notifications about content that could not be retrieved.
The solution was fairly simple; add `Accept: 'application/json'` to the request headers.

### DIFF
--- a/src/shared/services/api/api.js
+++ b/src/shared/services/api/api.js
@@ -37,6 +37,7 @@ export function* authCall(url, params, authorizationToken) {
 
 export function* authCallWithPayload(url, params, method) {
   const headers = {
+    accept: 'application/json',
     'Content-Type': 'application/json',
   };
 
--- a/src/shared/services/api/api.test.js
+++ b/src/shared/services/api/api.test.js
@@ -104,6 +104,7 @@ describe('api service', () => {
       const options = {
         method: 'METHOD',
         headers: {
+          accept: 'application/json',
           'Content-Type': 'application/json',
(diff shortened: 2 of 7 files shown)


src/shared/services/api/api.js
```diff
@@ -37,6 +37,7 @@ export function* authCall(url, params, authorizationToken) {
 
 export function* authCallWithPayload(url, params, method) {
   const headers = {
+    accept: 'application/json',
```
spaces-ga: This will most likely be corrected, but maybe this should be `Accept` for consistency

src/shared/services/api/api.test.js
```diff
@@ -104,6 +104,7 @@ describe('api service', () => {
       const options = {
         method: 'METHOD',
         headers: {
+          accept: 'application/json',
```
spaces-ga: see previous comment

src/shared/services/api/api.test.js
```diff
@@ -118,6 +119,7 @@ describe('api service', () => {
...
```
spaces-ga: See previous comment

src/shared/services/api/api.js
janjaap: Headers are case-**in**sensitive. See https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 for reference.

PR #583, opened 2020-01-15
[SIG-2140] Manage redirects

janjaap:
This PR solves two routing related problems:

- `/manage` URL shows a 404
- Logging out shows an error notification

The first issue was easily solved by putting in another redirect rule in the `App` container.

Second issue was a bit more complicated and required a change in the state after the `logout` action was dispatched and a push to a different URL after the logout was performed so that there were no requests performed that required a user to be authenticated. Also, a location reload has been removed to prevent the page from loading more than necessary.


src/containers/App/index.js
```diff
@@ -52,11 +56,12 @@ export const AppContainer = ({ requestCategoriesAction }) => {
           <Switch>
             <Redirect exact from="/" to="/incident" />
             <Redirect exact from="/login" to="/manage" />
+            <Redirect exact from="/manage" to="/manage/incidents" />
```
spaces-ga: Does this work for `/manage/` ?

janjaap: Yeah, it does. That URL is picked up by `<Route path="/manage" component={IncidentManagementModule} />`

PR #588, opened 2020-01-20
Sig 2057 edit user remove whitespace and subline email

jasperswart:
- in IE11 the user form and role form have grey bodies: setting app height to 100vh makes sure the white background is everywhere in all pages
- removed the hint test form user form email


janjaap:
@jasperswart The ticket number SIG-2057 is an entirely different ticket than what this ticket tries to solve.

PR #589, opened 2020-01-20
[SIG-2138] Loading spinner on searching incidents

janjaap:
This PR fixes an issue where the loading spinner wasn't shown when the search form was submit.


src/signals/incident-management/reducer.js
```diff
@@ -149,11 +149,13 @@ export default (state = initialState, action) => {
         .set('activeFilter', initialState.get('activeFilter'))
         .set('editFilter', initialState.get('editFilter'))
         .set('ordering', initialState.get('ordering'))
+        .set('loading', true)
         .set('page', initialState.get('page'))
         .set('searchQuery', action.payload);
 
     case RESET_SEARCH_QUERY:
       return state
+        .set('loading', true)
```
spaces-ga: Should this set `loading` to `true`?

janjaap: Yeah, it should, because both setting and resetting the search query will trigger an incident fetch. 

Ideally, `loading` should be set to 'true' when the `searchIncidents` action is dispatched, but that will result in an infinite loop, because of the way the saga routes and triggers requests.

PR #594, opened 2020-01-23
SIG-1850 - Add filter by username to Users Overview.

spaces-ga:
Allow to filter users by username

### DIFF
--- a/src/signals/settings/users/containers/Overview/hooks/useFetchUsers.js
+++ b/src/signals/settings/users/containers/Overview/hooks/useFetchUsers.js
@@ -13,7 +13,7 @@ import filterData from './filterData';
  *
  * @returns {FetchResponse}
  */
-const useFetchUsers = ({ page, pageSize } = {}) => {
+const useFetchUsers = ({ page, pageSize, filters } = {}) => {
   const [isLoading, setLoading] = useState(false);
   const [users, setUsers] = useState([]);
   const [error, setError] = useState(false);
@@ -25,16 +25,25 @@ const useFetchUsers = ({ page, pageSize } = {}) => {
     (async function fetchData() {
       setLoading(true);
 
+      const pageParams = [
+        page && `page=${page}`,
+        pageSize && `page_size=${pageSize}`,
+      ];
+
+      const filterParams = Object.entries(filters || {}).reduce(
+        (acc, [filter, value]) => value ? [...acc, `${filter}=${value}`] : acc,
+        []
+      );
+
+      const queryParams = [
+        ...pageParams,
+        ...filterParams,
+      ]
+        .filter(Boolean)
+        .join('&');
+
       try {
-        const params = [
-          page && `page=${page}`,
-          pageSize && `page_size=${pageSize}`,
-        ]
-          .filter(Boolean)
-          .join('&');
-        const url = [configuration.USERS_ENDPOINT, params]
-          .filter(Boolean)
-          .join('?');
+        const url = [configuration.USERS_ENDPOINT, queryParams].filter(Boolean).join('?');
         const response = await fetch(url, {
           headers: {
             ...getAuthHeaders(),
@@ -57,7 +66,7 @@ const useFetchUsers = ({ page, pageSize } = {}) => {
     return () => {
       controller.abort();
     };
-  }, [page, pageSize]);
+  }, [page, pageSize, filters]);
 
   /**
    * @typedef {Object} FetchResponse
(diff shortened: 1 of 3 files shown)


src/signals/settings/users/containers/Overview/hooks/useFetchUsers.js
```diff
@@ -25,16 +25,33 @@ const useFetchUsers = ({ page, pageSize } = {}) => {
     (async function fetchData() {
       setLoading(true);
 
+      const pageParams = [
+        page && `page=${page}`,
+        pageSize && `page_size=${pageSize}`,
+      ];
+
+      const filterParams = Object.keys(filters || {}).reduce((acc, filter) => {
```
janjaap: I think this can be simplified a bit:
```jsx
const filterParams = Object.entries(filters || {}).reduce(
  (acc, [filter, value]) => [...acc, `${filter}=${value}`],
  []
);
```

and maybe even

```jsx
const params = [
  page && `page=${page}`,
  pageSize && `page_size=${pageSize}`,
  Object.entries(filters || {}).reduce(
    (acc, [filter, value]) => [...acc, `${filter}=${value}`],
    []
  ),
]
  .filter(Boolean)
  .join('&');
```

src/signals/settings/users/containers/Overview/hooks/useFetchUsers.js
```diff
@@ -25,16 +25,25 @@ const useFetchUsers = ({ page, pageSize } = {}) => {
     (async function fetchData() {
       setLoading(true);
 
+      const pageParams = [
+        page && `page=${page}`,
+        pageSize && `page_size=${pageSize}`,
+      ];
+
+      const filterParams = Object.entries(filters || {}).reduce(
+        (acc, [filter, value]) => value ? [...acc, `${filter}=${value}`] : acc,
```
janjaap: For readability, I would add a `filter` function, like so:
```jsx
const filterParams = Object.entries(filters || {})
  .filter(([, value]) => Boolean(value))
  .reduce((acc, [filter, value]) => [...acc, `${filter}=${value}`], []);
```
Just a suggestion, though

src/signals/settings/users/containers/Overview/index.js
```diff
@@ -31,11 +32,18 @@ const HeaderButton = styled(Button)`
...
+  }
+`;
+
 export const UsersOverviewContainer = ({ pageSize, userCan }) => {
   const history = useHistory();
   const { pageNum } = useParams();
   const [page, setPage] = useState(1);
-  const { isLoading, users: { list: data }, users } = useFetchUsers({ page, pageSize });
+  const [filters, setFilters] = useState({});
+  const { isLoading, users: { list: data }, users } = useFetchUsers({ page, pageSize, filters });
```
janjaap: This was the construct I was looking for !!!!!1!
Getting both the object as well as the the destructured object 🤦‍♂

src/signals/settings/users/containers/Overview/index.js
```diff
@@ -31,11 +32,18 @@ const HeaderButton = styled(Button)`
...
+    display: none;
+  }
+`;
+
 export const UsersOverviewContainer = ({ pageSize, userCan }) => {
   const history = useHistory();
   const { pageNum } = useParams();
   const [page, setPage] = useState(1);
-  const { isLoading, users: { list: data }, users } = useFetchUsers({ page, pageSize });
+  const [filters, setFilters] = useState({});
```
janjaap: This should actually be a reducer; having an object in the state will always rerender the component, even when not necessary.

src/signals/settings/users/containers/Overview/index.js
```diff
@@ -103,6 +129,14 @@ export const UsersOverviewContainer = ({ pageSize, userCan }) => {
           <Column span={12}>
             <DataView
               headers={columnHeaders}
+              filters={[
+                (<StyledSearchbar
+                  label="Zoek op emailadres"
```
janjaap: I don't see the label anywhere when I load the user overview page. Does it do anything?

src/signals/settings/users/containers/Overview/index.js
```diff
@@ -103,6 +136,13 @@ export const UsersOverviewContainer = ({ pageSize, userCan }) => {
           <Column span={12}>
             <DataView
               headers={columnHeaders}
+              filters={[
+                (<StyledSearchbar
+                  placeholder=""
```
janjaap: For the ticket regarding the untrimmed input; the `Searchbar` component can receive a value for its `value` prop. You can trim the input, store it in the state and use that to populate the `value` prop.

spaces-ga: Tried that, doesn't seem to work out of the box...

PR #595, opened 2020-01-25
chore fix for deselecting checkboxes when nieuw filter has been clicked 

jasperswart:
this PR:
- it has a fix for the deselecting checkboxes when nieuw filter has been clicked
- found when I was trying to deselect the checkboxes when looking for Sig 2136 

PLEASE NOTE: no fix for filter history


spaces-ga:
> This works but it looks like this should be fixed in `src/signals/incident-management/components/CheckboxList/index.js` instead of (just) here.
> Then you could also add a test to `CheckboxList.test`

Discussed with JJ, your PR is good as it is

PR #597, opened 2020-01-27
[react-map]  adds react-maps and geocoder components

janjaap:
This PR replaces the amaps with react-maps


aditudorache:
Closes in favor of #733

PR #598, opened 2020-01-29
[SIG-2136] CheckboxList callback handlers

janjaap:
This PR adds two callback props to the `CheckboxList` component that allow getting the values of the items that have been checked in the component when anything changes.
- `onChange` is called every time an individual checkbox is checked, but only when not all boxes have been checked
- `onToggle` is called when the toggle button is clicked or when all boxes are ticked manually


janjaap:
Closing; superseded by #601

PR #599, opened 2020-01-29
SIG-2197 - Preserve filters state between Users Overview and Details pages.

spaces-ga:


src/signals/settings/users/containers/Overview/index.js
```diff
@@ -49,9 +49,11 @@ const filtersReducer = (state, action) => {
 
 export const UsersOverviewContainer = ({ pageSize, userCan }) => {
   const history = useHistory();
+  const location = useLocation();
+  const filtersInitialState = location.state && location.state.filters || {};
```
janjaap: Wouldn't `location.state.filters || {};` be enough? As far as I can see, the `location` object always has a `state` prop.

spaces-ga: Tried that before, but apparently it's not always there...

PR #609, opened 2020-02-05
[SIG-2176] - Remove setClassification

janjaap:
This PR removes the necessity of having the full list of categories from the public endpoint available in the meldingen form.

In order to accomplish this, this PR:
- removes the `retryFetchClassification` saga from `IncidentContainer`; it wasn't required and added more complexity than necessary
- renames `set-classification` service to `resolveClassification` and removes the requirement of passing a list of subcategories to that service
- has the `createIncident` saga query the public categories endpoint to retrieve the necessary data for the (sub)category that was provided by the prediction service
- removes the `subcategory_link` required field from the `wizard-step-1-beschrijf` form configuration
- adds a fixture (`postIncident.json`)


src/signals/incident/definitions/wizard-step-1-beschrijf.js
```diff
@@ -83,25 +84,14 @@ export default {
...
         },
         render: FormComponents.HiddenInput,
       },
-      subcategory_link: {
-        meta: {
-          label: 'Subcategorie',
-          path: 'category.sub_category',
-          type: 'text',
-          validateAsyncProp: 'incidentContainer.loadingClassification',
-        },
```
jasperswart: die validateAsyncProp zorgt ervoor dat er wordt gewacht met het ophalen van de  classificatie na het drukken op de submit knop. dat werkt niet meer in jouw branch

janjaap: Seems to go smoothly; Redux actions are executed in order and the `incident` scope is updated with the proper category and subcategory on blur or submit:

Screenshot 2020-02-11 at 16 36 49

janjaap: And I took out the `async` behaviour, because it wasn't needed anymore; not retrieving the full list of categories.

PR #635, opened 2020-02-21
[SIG-2049] useFetch hook

janjaap:
# Custom `useFetch` hook
This PR adds a generic `useFetch` hook. The hook contains asynchronous request handlers and can be used as follows

## GET

```jsx
const MyComponent = ({ id }) => {
  const { isLoading, error, data, get } = useFetch();

  useEffect(() => {
    const params = { foo: 'bar' };
    // method get can be given a query parameter object as its 
    // second function parameter
    get(`${ENDPOINT_URL}${id}`, params);
  }, []);

  return isLoading ? (
    <LoadingIndicator />
  ) : (
    // do something with data
  );
```

## PATCH

```jsx
const MyComponent = ({ id }) => {
  const { isLoading, isSuccess, error, data, patch } = useFetch();

  useEffect(() => {
    if (isSuccess) {
      global.alert('Success!!!1!);
    }
  }, [isSuccess]);

  const onSubmit = formData => {
    patch(`${ENDPOINT_URL}${id}`, formData);
  };

  return isLoading ? (
    <LoadingIndicator />
  ) : (
    <FormComponent onSubmit={onSubmit} />
  );
```

## POST

```jsx
const MyComponent = () => {
  const { isLoading, isSuccess, error, data, post } = useFetch();

  useEffect(() => {
    if (isSuccess) {
      // redirect to another page or whatever
    }
  }, [isSuccess]);

  const onSubmit = formData => {
    post(ENDPOINT_URL, formData);
  };

  return isLoading ? (
    <LoadingIndicator />
  ) : (
    <FormComponent onSubmit={onSubmit} />
  );
```

### DIFF
--- a/internals/testing/test-bundler.js
+++ b/internals/testing/test-bundler.js
@@ -8,8 +8,9 @@ import 'jest-localstorage-mock';
 
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import fetchMock from 'jest-fetch-mock';
 
-require('jest-fetch-mock').enableMocks();
+fetchMock.enableMocks();
 
 // React 16 Enzyme adapter
 Enzyme.configure({ adapter: new Adapter() });
--- a/internals/testing/test-bundler.js
+++ b/internals/testing/test-bundler.js
@@ -9,7 +9,7 @@ import 'jest-localstorage-mock';
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
-global.fetch = require('jest-fetch-mock');
+require('jest-fetch-mock').enableMocks();
 
 // React 16 Enzyme adapter
 Enzyme.configure({ adapter: new Adapter() });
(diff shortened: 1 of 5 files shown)


src/hooks/useFetch.js
```diff
@@ -0,0 +1,118 @@
...
+      });
+      /* istanbul ignore else */
+      if (response.ok === false) {
+        throw response;
+      }
+
+      const JSONResponse = await response.json();
+
+      setData(JSONResponse);
+    } catch (e) {
```
jpoppe: Personally I prefer not to use 1 letter variables, error (or event) is much better 'greppable' then 'e', my 2 cents, feel free to ignore!

internals/testing/test-bundler.js
```diff
@@ -9,7 +9,7 @@ import 'jest-localstorage-mock';
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
-global.fetch = require('jest-fetch-mock');
+require('jest-fetch-mock').enableMocks();
```
jpoppe: Could we use an import instead of a require?

src/hooks/useFetch.js
janjaap: Good one; I'll take that into account for future PRs

internals/testing/test-bundler.js
janjaap: Yeah, could definitely do. I'll change it.

PR #657, opened 2020-03-04
[SIG-2329] Fix for inactive categories

janjaap:
In the filter form and on the incident detail page, categories are shown that have been marked as inactive.

This PR contains a fix for that problem.


janjaap:
Closing; taking other branch for this PR. Branchs contains too many changes because of wrong base branch

PR #678, opened 2020-03-16
[SIG-2415] Added banner about availability

jpoppe:
Added warning banner about eventual delays in managing reports.


src/signals/incident/containers/IncidentContainer/index.js
```diff
@@ -24,6 +25,13 @@ import './style.scss';
 
 import IncidentWizard from '../../components/IncidentWizard';
 
+const Alert = styled.div`
+  color: white;
```
aditudorache: how about `themeColor('tint', 'level1');` ?

PR #683, opened 2020-03-17
[SIG-2315] Radio list component

janjaap:
This PR adds a radiobutton list to the incident split page. In addition to that, this PR:

- fixes an error in the `FieldControlWrapper` component that prevented the incident split page from rendering when the split page was accessed directly or the page was refreshed after it had loaded
- removes the unnecessary passing down of props from the `IncidentSplit` container component to its child components
- optimizes the `SplitForm` component by moving the form instantiation into the component (not having this would set previously selected radio button options as default instead of always having 'SIG' as the default) and by using `useCallback` for one of its internal callbacks.
- corrects tests where needed


src/signals/incident-management/containers/IncidentSplitContainer/components/SplitForm/index.js
```diff
@@ -50,56 +50,57 @@ const StyledBottomDisclaimer = styled(StyledDisclaimer)`
   margin: ${themeSpacing(5)} 0;
```
aditudorache: how about `${themeSpacing(5, 0)};`?

janjaap: 🤷‍♂

PR #684, opened 2020-03-17
SIG-1499 add map

aditudorache:
This PR adds the base map component with a base layer and renders a default marker


package.json
```diff
@@ -55,11 +55,13 @@
   ],
   "dependencies": {
     "@babel/polyfill": "^7.8.3",
-    "@datapunt/asc-assets": "^0.18.1",
-    "@datapunt/asc-core": "^0.18.1",
-    "@datapunt/asc-ui": "^0.18.2",
+    "@datapunt/amsterdam-react-maps": "git+https://github.com/Amsterdam/amsterdam-react-maps.git#feature/package-export",
+    "@datapunt/asc-assets": "0.18.2-alpha.29",
```
janjaap: This is a canary version and shouldn't be used in production. Why would we need this version?

package.json
```diff
@@ -55,11 +55,13 @@
   ],
   "dependencies": {
     "@babel/polyfill": "^7.8.3",
-    "@datapunt/asc-assets": "^0.18.1",
-    "@datapunt/asc-core": "^0.18.1",
-    "@datapunt/asc-ui": "^0.18.2",
+    "@datapunt/amsterdam-react-maps": "git+https://github.com/Amsterdam/amsterdam-react-maps.git#feature/package-export",
```
janjaap: Can't we use a specific version reference?

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,68 @@
...
+        <BackgroundLayer />
+      </StyledMap>
+    </MapWrapper>
+  );
+};
+
+MapEditor.propTypes = {
+  location: PropTypes.shape({
+    geometrie: PropTypes.shape({}),
+    address: PropTypes.shape({}),
```
janjaap: Can we more precisely define these shapes?

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,68 @@
...
+          setInstance={setMarker}
+          args={[{
+            lat: 0,
+            lng: 0,
+          }]}
+          options={{
+            icon: markerIcon,
+            opacity: 0,
+          }}
+        />
```
janjaap: Shouldn't the marker only be rendered when there is an actual location available to the Map component?

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,68 @@
...
+  const [marker, setMarker] = useState();
+
+  useEffect(() => {
+    if (!marker || !location.geometrie) return;
+
+    const opacity = 1;
+    const latlng = feature2location(location.geometrie);
+    marker.setLatLng(latlng);
+    marker.setOpacity(opacity);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
janjaap: Why is it necessary to disable the linter for this rule? Seems like `location` should also be `location.geometrie`.

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,68 @@
...
+const StyledMap = styled(Map)`
+  width: 100%;
+  height: 450px;
+`;
+
+const StyledViewerContainer = styled(ViewerContainer)`
+  z-index: 400;
+`;
+
+const MapEditor = ({ location }) => {
```
janjaap: This name is a bit confusing. Reading the component name gives me the idea that this component is an editor instead of a map

src/shared/services/configuration/map-markers.js
```diff
@@ -0,0 +1,14 @@
...
+  iconUrl: 'https://map.data.amsterdam.nl/dist/images/svg/marker.svg',
+  iconSize: [20, 20],
+  iconAnchor: [10, 19],
+});
+
+export const markerIcon = L.icon({
+  iconUrl: 'https://map.data.amsterdam.nl/dist/images/svg/marker.svg',
+  iconSize: [40, 40],
+  iconAnchor: [20, 39],
+});
```
janjaap: Isn't one icon enough? Why would we need a big and a small icon?

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,68 @@
+import React, { memo , useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Map, Marker } from '@datapunt/react-maps';
+import { ViewerContainer } from '@datapunt/asc-ui';
+import { Zoom } from '@datapunt/amsterdam-react-maps/lib/components';
+import styled from '@datapunt/asc-core';
+import MAP_OPTIONS from 'shared/services/configuration/map-options';
```
janjaap: Importing the options here locks the component to those options where they should be configurable.

src/shared/services/configuration/map-options.js
```diff
@@ -0,0 +1,20 @@
+import { getCrsRd } from '@datapunt/amsterdam-react-maps/lib/utils';
+
+export const DEFAULT_MARKER_POSITION = {
```
janjaap: I can't think of a situation where we would have a need for a default marker position.

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,114 @@
...
+export const feature2location = feature => {
+  const { coordinates } = feature;
+  return {
+    lat: coordinates[1],
+    lng: coordinates[0],
+  };
+};
+
+/**
+ * converts the sia location in latlon fromat
```
janjaap: Just the 'sia' location or any location?
fromat should be format

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,114 @@
...
-      ],
-    };
+  if (loc.address) {
+    location.address = address2pdok(loc.address);
   }
 
   return location;
 }
 
+export const pdok2address = pdokAddress => {
```
janjaap: Are we using PDOK services anywhere in the SIA application?

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,114 @@
...
+    huisnummer: `${huisnummer}`,
+    huisletter: huisletter || '',
+    huisnummertoevoeging: huisnummer_toevoeging || '',
+    postcode,
+    woonplaatsnaam: woonplaats,
+  };
+};
+
+/**
+ * converts the geocoder location in sia fromat
```
janjaap: fromat should be format

src/signals/incident/components/form/MapInput/index.js
```diff
@@ -2,17 +2,25 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import mapLocation from 'shared/services/map-location';
-import MapInteractive from 'components/MapInteractive';
+import MapEditor from 'components/MapEditor';
 
 import Header from '../Header';
+import { DEFAULT_MARKER_POSITION } from '../../../../../shared/services/configuration/map-options';
```
janjaap: Can you use `'shared/services/configuration/map-options'` here instead of a relative path?

src/signals/incident/components/form/MapInput/index.js
```diff
@@ -29,7 +37,7 @@ const MapInput = ({
               getError={getError}
             >
               <div className="invoer">
-                <MapInteractive onQueryResult={onQueryResult} location={value} />
+                <MapEditor onLocationChange={onLocationChange} location={value} />
```
janjaap: `onLocationChange` Doesn't seem to be a prop on the map component

package.json
aditudorache: There are some breaking changes in the amsterdam-react-maps that work with this version, will replace it asa theese changes are implemented in the main version

package.json
aditudorache: there are not yet versions defined in this component

src/components/MapEditor/index.js
aditudorache: thanks!

src/shared/services/map-location/index.js
aditudorache: yes, the location is the geometry and addres.

src/shared/services/map-location/index.js
aditudorache: not yet, but will do in the next pr :-)

src/signals/incident/components/form/MapInput/index.js
aditudorache: sure!

src/signals/incident/components/form/MapInput/index.js
aditudorache: Correct, the implementation wil follow in the next pr, if I remove it here, will generate some warnings in this component that I would have to disable.

src/components/MapEditor/index.js
aditudorache: we could implement it like this, but if we do that, we won't be able to 'flyto' in the later stage :-)

src/components/MapEditor/index.js
aditudorache: I don't like it either, I'm glad to change it please make a suggestion

src/components/MapEditor/index.js
aditudorache: moved it to the caller

src/shared/services/configuration/map-markers.js
aditudorache: small icon is used in the incident overview.

src/shared/services/configuration/map-options.js
aditudorache: I used this to show the defautl marker in this PR, Will remove in the next pr when the click functionality is implemented

package.json
janjaap: This notation won't work as Jenkins would need Github authentication tokens. We cannot merge this PR otherwise.

src/components/MapEditor/index.js
janjaap: Why not just `Map`?

src/components/MapEditor/index.js
janjaap: In which cases would we need `flyto`?

src/shared/services/configuration/map-markers.js
janjaap: Hm, can't we use a static image for that? We don't need an interactive map component in the incident preview mode.

src/shared/services/map-location/index.js
janjaap: Then please add only the functionality that we need in this PR. Makes the coming PRs much harder to review otherwise.

package.json
aditudorache: It works, It;s already implemented in another project

src/components/MapEditor/index.js
```diff
@@ -58,11 +56,23 @@ const MapEditor = ({ location }) => {
...
+    address: PropTypes.shape({
+      openbare_ruimte: PropTypes.string,
+      huisnummer: PropTypes.string,
+      huisletter: PropTypes.string,
+      huisnummer_toevoeging: PropTypes.string,
+      postcode: PropTypes.string,
+      woonplaats: PropTypes.string,
+    }),
   }).isRequired,
+  options: PropTypes.shape({ }).isRequired,
```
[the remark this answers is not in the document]
aditudorache: no, is not needed, the options shape is passed to the map and not used explicit here. We are talking about all the the all possible options of leaflet btw

src/components/MapEditor/index.js
aditudorache: extisting functionality: a marker is visible and then the user selects a location from the autosuggest.

src/shared/services/configuration/map-markers.js
aditudorache: we could but is more work than this while the component already does what it should do. The map has to be rendered anyway in the preview.

src/components/MapEditor/index.js
janjaap: Still, you're rendering a marker where that is not needed. If there is a location for a marker, only then the marker should be rendered. And it can be moved to a different point on the map when another location for the marker comes in.

src/components/MapEditor/index.js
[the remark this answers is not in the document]
janjaap: Can you then put in a docblock that has a link to the proper Leaflet documentation? Because, in the current case, there is no way to see which options can be passed in.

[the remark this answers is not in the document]
aditudorache: ok

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,80 @@
...
+  z-index: 400;
+`;
+
+const Map = ({ location, options }) => {
+  const [marker, setMarker] = useState();
+
+  useEffect(() => {
+    if (!marker || !location.geometrie) return;
+    const opacity = 1;
+    const latlng = feature2location(location.geometrie);
```
janjaap: `location.geometrie` Can still be falsy at this point, which means that calling `feature2location` will throw an exception. Please cover that scenario.

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,80 @@
...
+import { markerIcon } from 'shared/services/configuration/map-markers';
+import { feature2location } from 'shared/services/map-location';
+
+const MapWrapper = styled.div`
+  position: relative;
+`;
+
+const StyledMap = styled(MapComponent)`
+  width: 100%;
+  height: 450px;
```
janjaap: What happens when we need an instance of `MapEditor` to cover a different height?

src/shared/services/configuration/map-markers.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable global-require */
+import L from 'leaflet';
+
+export const smallMarkerIcon = global.window.L.icon({
```
janjaap: You're using `global.window.L.icon` here and `L.icon` on line 10. Won't that conflict? I'd use one notation and stick with that.

src/components/MapEditor/__tests__/MapInteractive.test.js
```diff
@@ -0,0 +1,37 @@
...
+      type: 'Point',
+      coordinates: [4, 52],
+    },
+  };
+
+  it('should render the map', () => {
+    const { container, getByText } = render(<Map options={MAP_OPTIONS} location={{}}/>);
+
+    // Map
+    expect(getByText('Leaflet')).toBeInTheDocument();
```
janjaap: Where does this text come from? I don't see any reference of it in the `MapEditor` component. We cannot rely on external dependencies in our test cases.

src/components/MapEditor/__tests__/MapInteractive.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+    // Tile layer
+    expect(container.querySelector('.leaflet-tile-container')).toBeInTheDocument();
```
janjaap: There is no reference to this class name in the `MapEditor` component. If you need to verify element's existence, you can use the `data-testid` attribute on components and elements and assert `getByTestId` or `queryByTestId`

src/components/MapEditor/__tests__/MapInteractive.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+    // Zoom
+    expect(container.querySelector('button[title="Inzoomen"]')).toBeInTheDocument();
+    expect(container.querySelector('button[title="Uitzoomen"]')).toBeInTheDocument();
```
janjaap: The test breaks on this assertion. Check the Jenkins console output for this PR's build.

src/components/MapEditor/__tests__/MapInteractive.test.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it('should render the marker', () => {
+    const { container, rerender } = render(<Map options={MAP_OPTIONS} location={{}}/>);
+
+    expect(container.querySelector('img.leaflet-marker-icon')).not.toBeInTheDocument();
```
janjaap: Please use `data-testid`

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,80 @@
...
+      </StyledMap>
+    </MapWrapper>
+  );
+};
+
+Map.propTypes = {
+  location: PropTypes.shape({
+    geometrie: PropTypes.shape({
+      type: PropTypes.string,
+      coordinates: PropTypes.arrayOf(PropTypes.number),
```
janjaap: Isn't this prop required? The `feature2location` function relies on it.

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
-function mapLocation(loc) {
+export const location2feature = location => ({
```
janjaap: This function doesn't seem to be used anywhere. Please remove if not used.

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
...
+  };
+};
+
+export const address2pdok = address => {
+  const { openbare_ruimte, huisnummer, huisletter, huisnummer_toevoeging, postcode, woonplaats } = address;
+
+  return {
+    straatnaam: openbare_ruimte,
+    huisnummer: `${huisnummer}`,
+    huisletter: huisletter || '',
```
janjaap: Why not also `${huisletter}`?

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
...
+    huisnummertoevoeging: huisnummer_toevoeging ? String(huisnummer_toevoeging) : '',
```
janjaap: Why not also `${huisnummer_toevoeging}`?

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
...
+    lat: coordinates[1],
+    lng: coordinates[0],
+  };
+};
+
+export const address2pdok = address => {
+  const { openbare_ruimte, huisnummer, huisletter, huisnummer_toevoeging, postcode, woonplaats } = address;
+
+  return {
+    straatnaam: openbare_ruimte,
```
janjaap: Is this correct? Can a value for `openbare_ruimte` be considered the same as `straatnaam`?

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
...
+    huisnummertoevoeging: huisnummer_toevoeging ? String(huisnummer_toevoeging) : '',
+    postcode,
+    woonplaatsnaam: woonplaats,
+  };
+};
+
+/**
+ * converts the location from `sia` location format to latlon format
+ */
+export function mapLocation(loc) {
```
janjaap: Please use `const`. Also, why is this function a named export as well as a default export? Please use either one of the two.

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
...
+
+export const feature2location = feature => {
+  const { coordinates } = feature;
+  return {
+    lat: coordinates[1],
+    lng: coordinates[0],
+  };
+};
+
+export const address2pdok = address => {
```
janjaap: This function doesn't seem to be used. Please remove if not used.

src/shared/services/map-location/index.js
```diff
@@ -1,28 +1,60 @@
...
+  }
+
+  if (loc.address) {
+    location.address = address2pdok(loc.address);
   }
 
   return location;
 }
 
+export const formatAddress = address => {
```
janjaap: This function doesn't seem to be used. Please remove if not used.

src/signals/incident/components/form/MapInput/index.js
```diff
@@ -2,17 +2,25 @@ import React from 'react';
...
 import Header from '../Header';
 
 const MapInput = ({
   handler, touched, hasError, meta, parent, getError, validatorsOrOpts,
 }) => {
-  const value = handler().value || {};
+  const value = handler().value || {
+    geometrie: {
+      type: "Point",
+      coordinates: [DEFAULT_MARKER_POSITION.lng, DEFAULT_MARKER_POSITION.lat],
```
janjaap: Why would we need to render a marker when there is no location available? Doesn't seem right.

src/signals/settings/components/PageHeader/index.js
```diff
@@ -20,6 +20,7 @@ const StyledSection = styled.section`
 const StyledHeading = styled(Heading)`
   margin: 0;
   line-height: 44px;
+  font-weight: 400;
```
janjaap: There are several changes in this PR like the one on this line. This seems not to be in scope of this PR and I'm quite certain that those changes are already present in `develop` and `master`. Have you brought your branch up to date with `develop`?

src/components/BackgroundLayer/__tests__/BackgroundLayer.test.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { withAppContext } from 'test/utils';
+import * as reactMaps from '@datapunt/react-maps';
+import BackgoundLayer from '..';
+
+describe('src/components/BackgroundLayer', () => {
+  it('should render the TileLayer', () => {
+    jest.spyOn(reactMaps, 'TileLayer').mockImplementation(() => 'TileLayerMock');
```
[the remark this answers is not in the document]
aditudorache: no, the rendering of the tile layer has effect changing the internal state of leaflet and doesn't render direct a html element. The test of leaflet component is out of scope here.

src/components/MapEditor/__tests__/MapInteractive.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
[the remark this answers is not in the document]
aditudorache: Ok

src/components/BackgroundLayer/__tests__/BackgroundLayer.test.js
```diff
@@ -0,0 +1,26 @@
...
+    expect(reactMaps.TileLayer).toHaveBeenCalledWith(
+      {
+        args: ['https://{s}.data.amsterdam.nl/topo_rd/{z}/{x}/{y}.png'],
+        options: {
+          attribution: 'Kaartgegevens CC-BY-4.0 Gemeente Amsterdam',
+          subdomains: ['t1', 't2', 't3', 't4'],
+          tms: true,
+        },
+      }
+      ,{}
```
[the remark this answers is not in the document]
aditudorache: that is the second parameter for the component, the first are the props and the second is the state which is empty and not null or undefined. I'm not sure why, but this is what it calls in this case.

src/components/MapEditor/index.js
aditudorache: added extra check

src/components/MapEditor/index.js
```diff
@@ -0,0 +1,80 @@
...
+  position: relative;
+`;
+
+const StyledMap = styled(MapComponent)`
+  width: 100%;
+  height: 450px;
+`;
+
+const StyledViewerContainer = styled(ViewerContainer)`
+  z-index: 400;
```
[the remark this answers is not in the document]
aditudorache: 400 is the first elevation that is not covered by the leaflet internal components. 399 and lower wouldn't show the container elements, in this case the Zoom control

src/components/MapEditor/index.js
aditudorache: Good point! I have changed this component to pass the styles from through. It can be used now as a styled-component.

src/shared/services/configuration/map-markers.js
aditudorache: You are right, I have corrected that.

src/components/MapEditor/__tests__/MapInteractive.test.js
aditudorache: Sure thing! I've added a test-id for this

src/components/MapEditor/__tests__/MapInteractive.test.js
aditudorache: I have corrected this in the amsterdam-react-maps, there was the text wrong for the Zoom out button.

src/components/MapEditor/index.js
aditudorache: it is, indeed.

src/shared/services/map-location/index.js
aditudorache: yes!

src/shared/services/map-location/index.js
aditudorache: Right! Changed to const.

src/shared/services/map-location/index.js
aditudorache: This one is used in the mapLocation conversion.

src/shared/services/map-location/index.js
aditudorache: it is used in MapPreview

src/signals/incident/components/form/MapInput/index.js
aditudorache: This is just for testing this pr. Will be removed in the next pr.

src/signals/settings/components/PageHeader/index.js
aditudorache: solved by updating the base branch with `develop`

PR #687, opened 2020-03-18
adds the pdok services and sends the click info

aditudorache:
This PR adds the pdok services and implements the click functionality in the MapEditor


aditudorache:
Closes in favor of #733

PR #698, opened 2020-03-23
[SIG-2414] Autosuggest component

janjaap:
# Autosuggest component

Fully accessible, ARIA compliant, text input with suggestions dropdown list.
Lay-out according to [design system requirements](https://designsystem.amsterdam.nl/7awj1hc9f/p/4902d2-auto-suggest).

## Usage

```jsx
// example response formatter; has to return an array objects with both 
// `id` and `value` props
const formatResponse = data => {
  return data.map(({ id, text }) => ({ id, value: text }));
};

// example number of options determiner function
const numOptionsDeterminer = data => data?.length || 0;

const onSelect = option => {
  // handle option selection
};

<AutoSuggest 
  formatResponse={formatResponse}
  numOptionsDeterminer={numOptionsDeterminer}
  onSelect={onSelect}
  url="https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest?q="
/>
```

## Accessibility

As per https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#kbd_label_listbox, keyboard interaction has been implemented as well as the required ARIA rules and roles.

## PDOK

The underlying geocoding service that is used in the current map implementation is [PDOK](https://www.pdok.nl/). Since this service can be considered to be somewhat stable, this PR also adds a PDOK implementation of the `AutoSuggest` component; `PDOKAutoSuggest`.


src/components/AutoSuggest/SuggestList.js
```diff
@@ -0,0 +1,96 @@
...
+
+const StyledList = styled.ul`
+  border: 1px solid ${themeColor('tint', 'level5')};
+  border-top: 0;
+  padding: 0;
+  margin: 0;
+`;
+
+const Li = styled.li`
+  line-height: 20px;
```
aditudorache: do we want to work with fixed values here? themeSpacing?

src/components/AutoSuggest/SuggestList.js
```diff
@@ -0,0 +1,96 @@
...
+    option => {
+      onSelectOption(option);
+    },
+    [onSelectOption]
+  );
+
+  return (
+    <StyledList className={className} role={role} ref={listRef} {...rest}>
+      {options.map(option => (
+        <Li
```
aditudorache: Can we use here the styled compoennts version of the list?
`https://amsterdam.github.io/amsterdam-styled-components/?path=/story/experimental-atoms-list--default-style-with-chevron-links`

src/components/AutoSuggest/index.js
```diff
@@ -0,0 +1,252 @@
...
+  const delayedCallback = useCallback(
+    debounce(value => {
+      serviceRequest(value);
+    }, INPUT_DELAY),
+    [serviceRequest]
+  );
+
+  const onChange = useCallback(
+    event => {
+      event.persist();
```
aditudorache: why is this needed here?

src/components/AutoSuggest/index.js
```diff
@@ -0,0 +1,252 @@
...
+      const { value } = event.target;
+      delayedCallback(value);
+    },
+    [delayedCallback]
+  );
+
+  const serviceRequest = useCallback(
+    value => {
+      if (value.length >= 3) {
+        get(`${url}${value}`);
```
aditudorache: You are assuming here that the call format of the search service will always match this pattern. Maybe passing an url formatter instead of `url` as a function parameter would give us more flexibility.

src/components/AutoSuggest/SuggestList.js
janjaap: Indeed, good catch. I'll change it.

src/components/AutoSuggest/SuggestList.js
janjaap: I tried, but the styling is different from the design system and I wasn't able to get a ref to the list so that I could get the keyboard navigation right as well as the focus 🤷‍♂

src/components/AutoSuggest/index.js
janjaap: If I don't persist the synthetic event, I cannot get a reference to its `target` prop.

src/components/AutoSuggest/index.js
janjaap: Good suggestion. However, I think that this format will cover all cases that we need for now. If we need a more intricate way of constructing the URL, we can always change it later on.

src/components/AutoSuggest/index.js
```diff
@@ -0,0 +1,252 @@
...
+  const [activeIndex, setActiveIndex] = useState(-1);
+  const wrapperRef = useRef(null);
+  const inputRef = useRef(null);
+
+  const handleKeyDown = useCallback(
+    event => {
+      const numberOfOptions = numOptionsDeterminer(data);
+      if (!show) return;
+
+      switch (event.keyCode) {
```
jpoppe: Why would you not use `key names` here instead of `keyCode`, `key names` would make it more readable IMHO?

See: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

```js
window.addEventListener("keydown", function (event) {
  if (event.defaultPrevented) {
    return; // Do nothing if the event was already processed
  }

  switch (event.key) {
    case "Down": // IE/Edge specific value
    case "ArrowDown":
      // Do something for "down arrow" key press.
      break;
    case "Up": // IE/Edge specific value
    case "ArrowUp":
      // Do something for "up arrow" key press.
      break;
```

src/components/PDOKAutoSuggest/index.js
```diff
@@ -0,0 +1,50 @@
...
+const serviceURL = 'https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest?';
+const numOptionsDeterminer = data => data?.response?.docs?.length || 0;
+const formatResponse = ({ response }) => response.docs.map(({ id, weergavenaam }) => ({ id, value: weergavenaam }));
+
+/**
+ * Geocoder component that specifically uses the PDOK location service to request information from
+ *
+ * @see {@link https://www.pdok.nl/restful-api/-/article/pdok-locatieserver#/paths/~1suggest/get}
+ */
+const PDOKAutoSuggest = ({ gemeentenaam, onSelect }) => {
```
aditudorache: I see no capability to set a value from props as discussed. Would you please add that?

janjaap: I know; forgot about that. Just added it.

src/components/AutoSuggest/index.js
janjaap: Done

aditudorache:
Shell we change the base branch  to `SIG-1499-map-component`? Merging to develop is too soon at this point I think :-)

src/components/AutoSuggest/index.js
jpoppe: Nice!

PR #699, opened 2020-03-24
enforces the linting rules on pre-commit

aditudorache:
This PR triggers the linting on pre-commit and merges two tasks in the scripts section


janjaap:
@aditudorache Why was this change necessary? Linters are already run on pre-commit.

aditudorache:
That's the problem, they are run but **not** with the correct parameters.

aditudorache:
Closing. Looks to be a local problem.

PR #701, opened 2020-03-24
[SIG-2442] Bug on split page

janjaap:
This PR fixes a bug on the incident split page where, if three incident parts were selected, all parts would get the default value for `type` instead of the selected value.

Note that a couple of test cases have been removed. This, because tests were breaking without any particular reason and I also couldn't find the reason why that was. When the `IncidentSplitContainer` component is refactored to remove the `connect` dependency, the tests won't be needed anyway 🤷‍♂

### DIFF
--- a/src/signals/incident-management/containers/IncidentSplitContainer/index.test.js
+++ b/src/signals/incident-management/containers/IncidentSplitContainer/index.test.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { render } from '@testing-library/react';
-import { mount } from 'enzyme';
 import * as reactRouterDom from 'react-router-dom';
 import { store, withAppContext } from 'test/utils';
 import incidentJson from 'utils/__tests__/fixtures/incident.json';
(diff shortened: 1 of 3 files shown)


src/signals/incident-management/containers/IncidentSplitContainer/index.test.js
```diff
@@ -20,15 +20,14 @@ store.dispatch(requestIncidentSuccess(incidentJson));
...
 }));
 
-jest.mock('models/incident/selectors', () =>
-  // eslint-disable-next-line global-require
-  jest.fn(() => require('utils/__tests__/fixtures/incident.json'))
-);
+// jest.mock('models/incident/selectors', () =>
+//   // eslint-disable-next-line global-require
+//   jest.fn(() => require('utils/__tests__/fixtures/incident.json'))
+// );
```
jpoppe: Shall we remove those commented lines?

janjaap: Totes. Already did that, but forgot to push the change.

PR #703, opened 2020-03-25
(Chore) Fix fetch GET

janjaap:
This PR fixes a bug where resources could not be loaded and returned a `401` response from the API until the page was reloaded.

The problem came from the request headers being defined outside the `useFetch` hook. When a user enters the site and the session is expired, the request headers are set without the bearer token.
When that same user authenticates without reloading the site, the request headers still have the bearer token missing. This is only solved by reloading the page so that the bearer token is included in the request headers.

By placing the initialisation of the request headers inside the `useFetch` hook function, the issue is resolved.


janjaap:
@aditudorache That is a suggestion, right? Or are you going to reject the pull request because of an existing construct?

janjaap:
Thanks 😗

PR #708, opened 2020-03-26
Sig 1499 add map context

aditudorache:
This PR
- Adds a map context
- Adds a base map container
- adds a map implementation with a selection marker


aditudorache:
Duplicate of #710. Closing this one

PR #710, opened 2020-03-26
Sig 1499 add map context

aditudorache:
This PR:
- creates a MapBase component that accepts children
- refactors the Map component to use MapBase
- Adds a MapContainer that:
   - uses a context/reducer pattern to manage the internal state 
   - accepts as parameters a value onChange pattern for enabling a controlled component
- Adds a MapImplementation component that:
  - Adds a MapBase with a Marker
  - Implements the marker functionality in the map through the MapContext
  - passes back the selected value to the parent component.

NB. Tests will be implemented when there are no more remarks


src/components/Map/index.js
```diff
@@ -1,97 +1,25 @@
...
-        </div>
-      </div>
-    );
-  }
-}
+const Map = ({ className, lat, lng, icon, ...otherProps }) => (
+  <MapBase className={className} data-testid="map-base" {...otherProps}>
+    {lat && lng && <Marker args={[{ lat, lng }]} options={{ icon }} />}
+  </MapBase>
+);
```
janjaap: As we discussed, remove this component and only leave `MapBase` (which should be called `Map) in the components folder.

src/components/MapBase/__tests__/MapBase.test.js
```diff
@@ -0,0 +1,32 @@
...
+    expect(container.querySelector('button[title="Inzoomen"]')).not.toBeInTheDocument();
+
+    rerender(<MapBase mapOptions={MAP_OPTIONS} hasZoomControls />);
+    expect(container.querySelector('button[title="Inzoomen"]')).toBeInTheDocument();
+  });
+
+  it('should implement isInteractive', () => {
+    /**
+     *  the isInteractive prop uses the intern state of leaflet (Browser.touch) and therefore cannot be tested
+     */
```
janjaap: This functionality can now be tested

src/components/MapBase/index.js
```diff
@@ -0,0 +1,61 @@
...
+import PropTypes from 'prop-types';
+import { ViewerContainer } from '@datapunt/asc-ui';
+import { Zoom } from '@datapunt/amsterdam-react-maps/lib/components';
+import styled from '@datapunt/asc-core';
+import { Map as MapComponent, TileLayer } from '@datapunt/react-maps';
+
+const Wrapper = styled(MapComponent)`
+  height: 450px;
+  width: 100%;
+`;
```
janjaap: This should not be in this component; should be the responsibility of the implementing components

src/components/PDOKAutoSuggest/index.js
```diff
@@ -2,23 +2,25 @@ import React from 'react';
...
   ['fq', 'bron:BAG'],
   ['fq', 'type:adres'],
   // ['fl', '*'], // undocumented; requests all available field values from the API
+  ['fl', 'id, weergavenaam, centroide_ll'],
   ['q', ''],
 ];
 const serviceURL = 'https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest?';
 const numOptionsDeterminer = data => data?.response?.docs?.length || 0;
-const formatResponse = ({ response }) => response.docs.map(({ id, weergavenaam }) => ({ id, value: weergavenaam }));
+const formatResponse = ({ response }) => response.docs.map(({ id, weergavenaam, centroide_ll }) => ({ id, value: weergavenaam, location:  centroide_ll && wktPointToLocation(centroide_ll) }));
```
janjaap: Is this change covered by a test case?

src/containers/MapContext/reducer.js
```diff
@@ -0,0 +1,43 @@
+import PropTypes from 'prop-types';
+import { SET_LOCATION, SET_ADDRESS, SET_VALUES } from './constants';
+
+export const mapValuesType = PropTypes.shape({
+  location: PropTypes.shape({
+    lat: PropTypes.number.isRequired,
+    lng: PropTypes.number.isRequired,
+  }),
+  addressText: PropTypes.string,
+});
```
janjaap: This type should be included in `shared/types`

src/shared/services/map-location/index.js
```diff
@@ -19,6 +24,20 @@ export const address2pdok = address => {
...
+  const coordinate = wktPoint.split('(')[1].split(')')[0];
+  const lat = parseFloat(coordinate.split(' ')[1]);
+  const lng = parseFloat(coordinate.split(' ')[0]);
+
+  return {
+    lat,
+    lng,
+  };
+};
+
```
janjaap: This function doesn't seem to have a test case. Also, the implementation expects a specific object as parameter. Provide a docblock that described the function as well as its params and return value.

src/components/MapBase/__tests__/MapBase.test.js
aditudorache: It looks that is not possible with jest and mocking to trigger the moveStart event in leafled

src/components/PDOKAutoSuggest/index.js
aditudorache: reverted

src/containers/MapContext/reducer.js
aditudorache: reverted

src/shared/services/map-location/index.js
aditudorache: removed

PR #712, opened 2020-03-27
[SIG-2447] Refactor incident detail component

janjaap:
This PR contains a full refactor of the `IncidentDetail` container and its inner components. 

### Prop drilling

Removal of prop drilling from the `IncidentContainer` component down to its most deeply nested child components. All props have been hoisted to the top-most component and actions are dispatched through context.

### Global state exclusion

The `IncidentDetail` container component doesn't make use of the global state any more; it holds it own state and passes its dispatch functions down.

Where the previous implementation looked similar to this:

```jsx
const ChangeValue = ({ onUpdateIncident }) => (
  <Element onClick={onUpdateIncident} />
);

const MetaList = ({ onUpdateIncident }) => (
  <ChangeValue onClick={onUpdateIncident} />
);

const IncidentDetail = ({ onUpdateIncident }) => (
  <MetaList onUpdateIncident={onUpdateIncident} />
);
```

It now looks like this:

```jsx
const ChangeValue = () => {
  const { update } = useContext(IncidentDetailContext);

  return (
    <Element onClick={update} />
  );
};

const IncidentDetail = () => {
  const update = useCallback(() => {}, []);
  return (
    <IncidentDetailContext.Provider value={{ update }}>
      <MetaList />
    </IncidentDetailContext.Provider>
  );
};
```

### useEventEmitter hook

Another notable change is that the `Highlight` component now uses the `useEventEmitter` hook to respond to incident update events that are emitted by the `IncidentDetail` container component.

Handling one-time events through a store is tedious, because the event, once emitted, has to be unset or the corresponding part of the store has to be reset, which requires two-way communication between the context/store and the appropriate component. Using the `useEventEmitter` hook makes it a fire-and-forget action which leads to a much simpler component implementation.

### Incident model
`models/incident` has been removed and only `patchIncidentSuccess` action has been moved to the `incident-management` module.


src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -33,225 +20,283 @@ import AttachmentViewer from './components/AttachmentViewer';
...
   position: relative;
+  z-index: 1;
+  justify-content: flex-start;
 `;
 
-export class IncidentDetail extends React.Component {
-  // eslint-disable-line react/prefer-stateless-function
-  constructor(props) {
-    super(props);
+const reducer = (state, action) => {
```
jpoppe, line 38: There is still a lint warning here (Expected to return a value at the end of arrow function)

PR #713, opened 2020-03-27
[SIG-2447] Incident reset on page navigation

janjaap:
# Incident reset on page navigation

This PR contains changes that fixes an issue where, after an incident had been submitted, the state that held the incident's data, was not cleared.

The `IncidentContainer` contained a listener for changes to the browser's location, but a crucial change was not picked up properly and that was the navigation away from the last page in the incident submission process.

By moving the listener to the `App` container this problem was solved.

### DIFF
--- a/src/signals/incident/containers/IncidentContainer/saga.js
+++ b/src/signals/incident/containers/IncidentContainer/saga.js
@@ -111,15 +111,24 @@ export function* getPostData(action) {
     handling_message,
   };
 
+  const invalidFields = ['incident_date', 'incident_time_minutes', 'subcategory', 'datetime', 'incident_time_hours', 'description', 'status'];
   const authenticatedOnlyFields = ['priority', 'source', 'type'];
 
   // function to filter out values that are not supported by the public API endpoint
   const filterSupportedFields = ([key]) =>
     isAuthenticated() || (!isAuthenticated() && !authenticatedOnlyFields.includes(key));
 
+  const filterInvalidFields = ([key]) => {
+    const isInvalid = invalidFields.includes(key);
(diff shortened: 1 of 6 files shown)


src/signals/incident/containers/IncidentContainer/saga.js
```diff
@@ -111,15 +111,24 @@ export function* getPostData(action) {
     handling_message,
   };
 
+  const invalidFields = ['incident_date', 'incident_time_minutes', 'subcategory', 'datetime', 'incident_time_hours', 'description', 'status'];
```
aditudorache: Is this a part of the PR? I can't figure it out...

src/signals/incident/containers/IncidentContainer/saga.test.js
```diff
@@ -145,6 +145,40 @@ describe('IncidentContainer saga', () => {
   });
 
   describe('getPostData', () => {
+    it('returns data that is valid', () => {
```
aditudorache: Is this a part of the PR? I can't figure it out...

src/signals/incident/containers/IncidentContainer/saga.js
janjaap: It is, otherwise the fix can't be tested, because of the limitations currently in the API (throwing `400` responses because of props that are part of the payload that should not be sent along).

janjaap:
How is a global interceptor of route changes different from what is this PR now? The interceptor is global, because it's at the app level. Ideally, we wouldn't need such a thing, but the way the state is constructed now, forces us solve it in this way. At a later point in time, the state needs to be refactored drastically so that it only holds what it should hold and that it is reset at the correct moment. The implementation in this PR is a temporary workaround.

PR #729, opened 2020-04-06
[SIG-2389] Checkbox list toevoegen

jpoppe:
## Add role selection checkboxes to the user overview page

This pull request implements roles selection checkboxes to the user overview page

## ~~TODO~~

~~Since the `user.json` has been updated two tests fail:
`./node_modules/.bin/jest --watch src/containers/App/selectors.test.js`~~

~~It fails because `getPermissionCount` returns a wrong?? number.~~
~~
- ~~To be able to test everything properly I have added a `userNew.json` fixture since the current `user.json` fixture seems to be a bit outdated, some old tests gave weird issues when using the new `user.json` fixture which were not obvious to fix and out of scope of for this pull request~~

- ~~There is a bug in the `CheckBoxList` component when all `checkboxes` are selected one less will be sent, the bug is triggered in this if statement:~~

~~https://github.com/Amsterdam/signals-frontend/blob/8f609242c4f9d1ec80c46ba51397cfa33df7cdad/src/signals/incident-management/components/CheckboxList/index.js#L204~~


src/signals/settings/users/Detail/components/UserForm/__tests__/UserForm.test.js
```diff
@@ -136,13 +189,33 @@ describe('signals/settings/users/containers/Detail/components/UserForm', () => {
     expect(radio2.checked).toBe(true);
   });
 
-  it('should check an unchecked checkbox', () => {
+  it('should check an unchecked role checkbox', async () => {
```
janjaap, line 192: `async` Is not necessary here.

src/signals/settings/users/Detail/components/UserForm/index.js
```diff
@@ -146,14 +167,25 @@ const UserForm = ({ data, onCancel, onSubmit, readOnly }) => {
...
 
+          <FieldGroup>
+            <CheckboxList
+              defaultValue={state.roles}
+              disabled={readOnly}
+              groupName="roles"
+              name="roles"
+              options={inputRoles}
+              onChange={(field, value) => { onChange(field, value); }}
+            />
```
janjaap, line 178: Group is missing a `<Label />`

src/signals/settings/users/Detail/components/UserForm/__tests__/UserForm.test.js
jpoppe: Oops! Normally ESLint warns me about this, will take a look later to enable this in our code base ;)

src/signals/settings/users/Detail/components/UserForm/index.js
jpoppe: Fixed, thanks!

PR #734, opened 2020-04-07
[SIG-2504] Update asc-ui to 0.19.1

janjaap:
This PR:
- updates `@datapunt/asc-ui` to `0.19.1`
- fixes tests and components because of breaking changes


package.json
```diff
@@ -60,7 +60,7 @@
     "@datapunt/amsterdam-react-maps": "0.1.2-alpha.1",
     "@datapunt/asc-assets": "^0.18.1",
     "@datapunt/asc-core": "^0.18.1",
```
aditudorache, line 62: shouldn't this be removed?

janjaap: `asc-core` is still used in the application

PR #736, opened 2020-04-08
[SIG-1505] Map toggle buttons

janjaap:
This PR adds a section to the incident overview page that has a tabbed list of links that can toggle the list and the map on that page.

Screenshot 2020-04-08 at 09 26 38

Next to that, this PR:
- adjusts the margins of the `PageHeader` component to make the tabbed list fit
- sets the overflow of the `body` to 'scroll' so that the page scrollbar is always visible and the page never jumps when navigating from a long page to a short page
- moves `IncidentOverview` styled components into their own file

### DIFF
--- a/src/signals/incident-management/containers/IncidentOverviewPage/components/SubNav/index.js
+++ b/src/signals/incident-management/containers/IncidentOverviewPage/components/SubNav/index.js
@@ -1,0 +1,100 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Row, Column, themeSpacing, themeColor, Heading } from '@datapunt/asc-ui';
+import { Link } from 'react-router-dom';
+
+import { MAP_URL, INCIDENTS_URL } from '../../../../routes';
+
+export const Wrapper = styled(Row)`
+  margin-bottom: ${themeSpacing(5)};
+`;
+
+export const MapHeading = styled(Heading).attrs({
+  forwardedAs: 'h2',
+})`
+  margin: 0;
+  font-size: 16px;
+`;
+
+export const TabContainer = styled.div`
+  border-bottom: 2px solid ${themeColor('tint', 'level3')};
+  display: flex;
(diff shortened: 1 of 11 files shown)


src/signals/incident-management/containers/IncidentOverviewPage/components/SubNav/index.js
```diff
@@ -0,0 +1,100 @@
...
+`;
+
+const SubNav = ({ showsMap }) => (
+  <Wrapper data-testid="subNav">
+    <Column span={{ small: 1, medium: 1, big: 3, large: 6, xLarge: 6 }}>
+      {showsMap && <MapHeading>Afgelopen 24 uur</MapHeading>}
+    </Column>
+
+    <TabWrapper span={{ small: 1, medium: 1, big: 3, large: 6, xLarge: 6 }}>
+      <TabContainer>
```
aditudorache, line 67: Not a fan of this construction, can we create a ticket for refactoring later on?

src/signals/incident-management/containers/IncidentOverviewPage/index.test.js
```diff
@@ -243,10 +243,10 @@ describe('signals/incident-management/containers/IncidentOverviewPage', () => {
     expect(props.pageChangedAction).toHaveBeenCalledWith(pagenum);
   });
 
-  it('should render a map', () => {
+  it('should render a map', async () => {
```
aditudorache, line 246: this is more of a integration than an unit test :-)

src/signals/incident-management/containers/IncidentOverviewPage/components/SubNav/index.js
janjaap: Which construct?

src/signals/incident-management/containers/IncidentOverviewPage/index.test.js
janjaap: Not really, it tests that specific parts are rendered in this component
🤷‍♂

PR #761, opened 2020-04-14
[SIG-2545] Fix for extra_properties object

janjaap:
This PR fixes an issues where, for older incidents, the `ExtraProperties` component would throw an error. A while ago, the API expected a different structure for extra properties for incidents, namely an object where it's keys were the properties' labels and its values, well, the values. Currently, the API expects an array of objects. The current implementation of the `ExtraProperties` components expects an array which leads to the incidents that still have objects for values to throw an exception in the incident detail page.

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/Detail/components/ExtraProperties/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/Detail/components/ExtraProperties/index.js
@@ -7,7 +7,7 @@ import { extraPropertiesType } from 'shared/types';
 
 const getValue = answer => {
   if (isArray(answer)) {
-    return answer.map(item => isObject(item) ? item.label : item).join(', ');
+    return answer.map(item => (isObject(item) ? item.label : item)).join(', ');
   }
   if (isObject(answer)) {
     if (isBoolean(answer.value)) {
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/containers/IncidentDetail/components/Detail/components/ExtraProperties/index.js
```diff
@@ -19,14 +19,18 @@ const getValue = answer => {
   return answer;
 };
 
-const ExtraProperties = ({ items }) =>
-  items &&
-  items.map(item => (
+const ExtraProperties = ({ items }) => {
+  // Some incidents have been stored with values for their extra properties that is incompatible with the current API
+  // We therefore need to check if we're getting an array or an object
+  const itemList = Array.isArray(items) ? items : Object.entries(items);
```
aditudorache, line 25: I would mention here that the `good` last version of the interface is the array. It is said in the PR description but later on will be forgotten.

janjaap: The prop types already mention that the type of `items` needs to be an array, so no need to be more explicit than that.

PR #762, opened 2020-04-14
[SIG-2529] Fix for address string parse

janjaap:
This PR addresses an issue where selecting an option in the autosuggest dropdown would show `undefined` in the address field, like so: `President Kennedyplantsoen 1A, undefined Amsterdam`. This was causes by a function that didn't take into account that specific address values could be undefined.

Next to that, this PR increases test coverage for the `map-location` service and makes sure that when an option in the `AutoSuggest` component is selected, the parent form isn't submitted.


src/shared/services/map-location/index.js
```diff
@@ -63,29 +46,46 @@ export const mapLocation = loc => {
   return value;
 };
 
-export const formatMapLocation = loc => {
+const getAddressText = ({ openbare_ruimte, huisnummer, huisnummer_toevoeging, postcode, woonplaats }) =>
+  [
+    [openbare_ruimte, huisnummer, huisnummer_toevoeging && `-${huisnummer_toevoeging}`],
```
aditudorache, line 51: Wouldn't we get here between huisnummer and huisnummer_toevoeging an extra space?

janjaap: I don't think so. Why do think that?

src/shared/services/map-location/index.js
```diff
@@ -70,28 +78,14 @@ export const formatMapLocation = location => {
...
 };
 
-export const formatAddress = address => {
-  const toevoeging = address.huisnummer_toevoeging ? `-${address.huisnummer_toevoeging}` : '';
-  const display = address.openbare_ruimte
-    ? `${address.openbare_ruimte} ${address.huisnummer}${address.huisletter}${toevoeging}, ${address.postcode} ${address.woonplaats}`
-    : '';
-  return display;
-};
+export const formatAddress = address => getAddressText(address);
```
aditudorache, line 88: Can this function been dropped?

janjaap: It's used somewhere else and I didn't want to touch that part of the application just yet.

src/shared/services/map-location/index.js
aditudorache: I see the unittest  'should render the address name' is failing

janjaap: Ah, I see what you mean. I'm asking around what the correct notation of an address should be.

janjaap: Fixed

src/shared/services/map-location/index.js
```diff
@@ -63,35 +46,51 @@ export const mapLocation = loc => {
   return value;
 };
 
-export const formatMapLocation = loc => {
+const getAddressText = ({ openbare_ruimte, huisnummer, postcode, woonplaats }) =>
+  [
+    [openbare_ruimte, huisnummer],
```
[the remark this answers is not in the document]
janjaap: Cool call; I missed that. I partially reverted the previous change.

src/shared/services/map-location/index.js
```diff
@@ -46,9 +46,9 @@ export const mapLocation = loc => {
   return value;
 };
 
-const getAddressText = ({ openbare_ruimte, huisnummer, postcode, woonplaats }) =>
+const getAddressText = ({ openbare_ruimte, huisnummer, huisletter, huisnummer_toevoeging, postcode, woonplaats }) =>
   [
-    [openbare_ruimte, huisnummer],
+    [openbare_ruimte, `${huisnummer || ''}${huisletter || ''}${huisnummer_toevoeging || ''}`],
```
aditudorache, line 51: Correct would be here een dash `-` between the `huisletter` and `toevoeging`. This is the standard notation and I would keep that.

src/shared/services/map-location/index.test.js
```diff
@@ -122,10 +131,12 @@ describe('formatMapLocation', () => {
 
     const result = {
       location: { lat: 52, lng: 4 },
-      addressText: 'Keizersgracht 666, 1016EJ Amsterdam',
+      addressText: 'Keizersgracht 666D3, 1016EJ Amsterdam',
```
aditudorache, line 134: Missing a dash '-' here, correct notation of the address should be `666D-3`

src/shared/services/map-location/index.js
janjaap: Since we're getting the full house number from the geocoder service now and since that address already has the correct notation, it's not necessary to format the address like that anymore. That's why I left the dash out.

PR #765, opened 2020-04-14
Feature/1 update npm dependencies

boris-arkenaar:
# Before

Run `npm audit` and it will tell you
```
found 38419 vulnerabilities (38360 low, 30 moderate, 28 high, 1 critical)
```

# Change

What I then did was

```
npm audit fix
npm audit fix
npm audit fix
npm remove leaflet-headless
npm i --save-dev leaflet-headless
npm remove build
```

# Result

With the following result after running `npm audit` again

```
found 3 vulnerabilities (1 low, 1 moderate, 1 high)
```

## Vulnerabilities left

First of all, all three vulnerabilities left are in dev dependencies.

Two vulnerabilities left are a moderate and a high one with `js-yaml`, indirect dependency of `babel-plugin-inline-react-svg`. These are [discussed here](https://github.com/airbnb/babel-plugin-inline-react-svg/issues/45) where they say that these are actually false positives since no yaml input is being used. Unfortunately not easy to fix they say. We could only see for ourselves if we could upgrade to a different library which solves the same problem.

Then there is only one low vulnerability left with `minimist`, indirect dependency of `webpack-pwa-manifest`. This should be [fixed by this PR](https://github.com/arthurbergmz/webpack-pwa-manifest/pull/122), but that hasn't been merged yet.

## Testing

To make sure everything still works after all these changes I ran all the tests (`npm test`) and ran the client (`npm start`) making sure the page seems to load correctly in the browser.

Please make sure as well if everything is still working. Maybe I'm missing some features that I haven't tested and might be broken.


janjaap:
@boris-arkenaar Thanks for the PR. That cleans the warning up nicely. Can you resolve the conflicts?

boris-arkenaar:
Fixes https://github.com/Signalen/frontend/issues/1

janjaap:
Replaced by #787

PR #773, opened 2020-04-15
Sig 2532 zoom to incident

aditudorache:
This PR persitsts the selections in het MapInput and MapSelect when coming back from the incident preview

### DIFF
--- a/src/components/MapInput/__tests__/MapInput.test.js
+++ b/src/components/MapInput/__tests__/MapInput.test.js
@@ -133,26 +133,26 @@ describe('components/MapInput', () => {
       movestart: movestartSpy,
     };
 
-    const withMapContextState = Component => state =>
+    const withMapContextState = state => Component =>
       withAppContext(<context.Provider value={{ state, dispatch: () => {} }}>{Component}</context.Provider>);
 
     const state = { ...initialState };
 
     const { rerender } = render(
-      withMapContextState(<MapInput mapOptions={MAP_OPTIONS} value={{}} events={mapEvents} />)(state)
+      withMapContextState(state)(<MapInput mapOptions={MAP_OPTIONS} value={{}} events={mapEvents} />)
     );
 
     expect(movestartSpy).not.toHaveBeenCalled();
 
-    rerender(withMapContextState(<MapInput mapOptions={MAP_OPTIONS} value={testLocation} events={mapEvents} />)(state));
+    rerender(withMapContextState(state)(<MapInput mapOptions={MAP_OPTIONS} value={testLocation} events={mapEvents} />));
 
     expect(movestartSpy).toHaveBeenCalled();
     movestartSpy.mockClear();
 
     const stateWithLocation = { ...state, ...testLocation };
     rerender(
-      withMapContextState(<MapInput mapOptions={MAP_OPTIONS} value={testLocation} events={mapEvents} />)(
-        stateWithLocation
+      withMapContextState(stateWithLocation)(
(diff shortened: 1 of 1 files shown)


src/components/MapInput/index.js
```diff
@@ -115,10 +116,18 @@ const MapInput = ({ className, value, onChange, mapOptions, ...otherProps }) =>
...
-    if (!value?.geometrie?.coordinates && !value?.addressText) return;
+    if (!map) return;
+    if (!value?.location && !value?.addressText) return;
+
+    if (value?.location?.lat > 0) {
+      if (map && !isEqual(value?.location, state.location)) {
+        const currentZoom = map.getZoom();
+        map.flyTo(value?.location, currentZoom);
+      }
+    }
```
janjaap, line 127: This code block isn't covered by tests

src/components/MapInput/index.js
```diff
@@ -115,10 +116,18 @@ const MapInput = ({ className, value, onChange, mapOptions, ...otherProps }) =>
   }, [marker, location, hasLocation]);
 
   useEffect(() => {
-    if (!value?.geometrie?.coordinates && !value?.addressText) return;
+    if (!map) return;
+    if (!value?.location && !value?.addressText) return;
+
+    if (value?.location?.lat > 0) {
+      if (map && !isEqual(value?.location, state.location)) {
```
janjaap, line 123: The check for `map` is redundant; this code cannot be reached whenever `map` is falsy (see line 119).

src/components/MapInput/index.js
```diff
@@ -115,10 +116,18 @@ const MapInput = ({ className, value, onChange, mapOptions, ...otherProps }) =>
...
+        const currentZoom = map.getZoom();
+        map.flyTo(value?.location, currentZoom);
```
janjaap, line 125: No need for optional chaining here; we know that `value` has a prop `location` at this point.

src/components/MapInput/index.js
```diff
@@ -115,10 +116,18 @@ const MapInput = ({ className, value, onChange, mapOptions, ...otherProps }) =>
...
   useEffect(() => {
-    if (!value?.geometrie?.coordinates && !value?.addressText) return;
+    if (!map) return;
+    if (!value?.location && !value?.addressText) return;
+
+    if (value?.location?.lat > 0) {
+      if (!isEqual(value?.location, state.location)) {
+        const currentZoom = map.getZoom();
+        map.flyTo(value.location, currentZoom);
+      }
```
janjaap, line 126: This clause still doesn't seem to be properly covered by a test case.

src/components/MapInput/index.js
```diff
@@ -115,10 +116,18 @@ const MapInput = ({ className, value, onChange, mapOptions, ...otherProps }) =>
   }, [marker, location, hasLocation]);
 
   useEffect(() => {
-    if (!value?.geometrie?.coordinates && !value?.addressText) return;
+    if (!map) return;
+    if (!value?.location && !value?.addressText) return;
```
janjaap, line 120: I don't see `value.addressText` being used in this function. Does it need to be present?

src/components/MapInput/index.js
aditudorache: Done

src/components/MapInput/index.js
aditudorache: removed

aditudorache:
closed through #779

PR #775, opened 2020-04-15
(Chore) Exception message assignment

janjaap:
This PR fixes an issue where the application would throw error unrelated to pending actions. This was due to the fact that in the `useFetch` custom hook, the error message was set on a freezed object.

### DIFF
--- a/src/hooks/__tests__/useFetch.test.js
+++ b/src/hooks/__tests__/useFetch.test.js
@@ -4,7 +4,7 @@ import { getErrorMessage } from 'shared/services/api/api';
 
 import useFetch from '../useFetch';
 
-jest.mock('shared/services/api/api');
+// jest.mock('shared/services/api/api');
 
 const URL = 'https://here-is-my.api/someId/6';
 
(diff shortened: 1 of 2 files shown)


janjaap:
What do you mean by immutable? You example doesn't show anything immutable.

aditudorache:
Bad choice of words, I meant create a new instance of the object and pass it through.

janjaap:
Ah, like that. Well, that would make testing significantly harder as I cannot assets that a thrown error is the actual error that is available after hook update.

janjaap:
I've added a couple of assertions that verify that the message has been set on the thrown error.

src/hooks/__tests__/useFetch.test.js
```diff
@@ -4,7 +4,7 @@ import { getErrorMessage } from 'shared/services/api/api';
 
 import useFetch from '../useFetch';
 
-jest.mock('shared/services/api/api');
+// jest.mock('shared/services/api/api');
```
aditudorache: forgot to remove?

janjaap: No, it's not there any more.

PR #776, opened 2020-04-15
[SIG-2540] Remove marker on input clear

janjaap:
This PR applies changes to the `AutoSuggest` and `MapInput` components so that the marker can be removed from the map whenever the auto suggest input field is emptied.

### DIFF
--- a/src/components/MapInput/__tests__/MapInput.test.js
+++ b/src/components/MapInput/__tests__/MapInput.test.js
@@ -227,7 +227,7 @@ describe('components/MapInput', () => {
     expect(container.querySelector(`.${markerIcon.options.className}`)).toBeInTheDocument();
   });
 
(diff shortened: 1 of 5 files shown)


src/components/MapInput/__tests__/MapInput.test.js
```diff
@@ -225,4 +226,41 @@ describe('components/MapInput', () => {
 
     expect(container.querySelector(`.${markerIcon.options.className}`)).toBeInTheDocument();
   });
+
+  it.only('should clear location and not render marker', async () => {
```
aditudorache, line 230: forgot to remove `only`?

janjaap: Whoops

PR #777, opened 2020-04-15
[SIG-2549] Remove translations

janjaap:
This PR removes translation files and all translation references from the code. These include:
- component/container generators
- the Dockerfile
- hot module reloading


src/hooks/useFetch.js
```diff
@@ -105,10 +109,13 @@ export default () => {
 
       setData(responseData);
       setSuccess(true);
-    } catch (e) {
-      e.message = getErrorMessage(e);
+    } catch (exception) {
+      Object.defineProperty(exception, 'message', {
```
aditudorache, line 113: Why is this construction needed here? Is it related to the translations?

janjaap: Not related; added it because otherwise the tests would break. Requires #775 to be merged first

PR #781, opened 2020-04-16
[SIG-2557] Fix selection of all checkboxes when the onToggle prop has not been set

jpoppe:
## Description

With this pull request the 'CheckBoxlist' will also submit all selected boxes when every box is checked and the `onToggle` prop has not been set.


src/signals/incident-management/components/CheckboxList/index.js
```diff
@@ -200,8 +200,8 @@ const CheckboxList = ({
         setToggled(allOptionsChecked);
 
         // in case that a list of options contains of only one item, we need to call the `onToggle`
-        // callback function instead of the `onChange` callback
-        if (allOptionsChecked) {
+        // callback function instead of the `onChange` callback unless the `onToggle` is undefined
+        if (onToggle && allOptionsChecked) {
```
janjaap, line 204: I don't see a test that covers this change. Can you confirm?

jpoppe: Yes, and I have been thinking about it, let me ping you on Slack ;)

PR #785, opened 2020-04-20
[SIG-2538] adds Avenir Next to the MapSelect controls

aditudorache:
This PR fixes the font choice in the MapSelect component

### DIFF
--- a/src/components/MapSelect/style.scss
+++ b/src/components/MapSelect/style.scss
@@ -30,7 +30,6 @@
   .zoom-control {
     background-color: #FEC813;
     color: $ams-zwart;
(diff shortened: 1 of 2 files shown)


src/components/MapSelect/style.scss
```diff
@@ -30,6 +30,7 @@
   .zoom-control {
     background-color: #FEC813;
     color: $ams-zwart;
+    font-family: Avenir Next LT W01, arial, sans-serif;
```
janjaap, line 33: Wouldn't it make more sense to apply this style rule to the `StyledMap` component in `index.js`?

aditudorache: Good point thanks!

PR #788, opened 2020-04-20
[ADR004] Initial draft graphql layer between frontend and backend

jpoppe:
## ADR-004

See Files changed for all the info


jpoppe:
Thanks for your feedback @aditudorache :) I  just saw your comment and will talk to `dataportaal` soon (was already in the pipeline :))

jpoppe:
@aditudorache the ADR has been discussed with `dataportaal` and is ready for review :partying_face:

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
+# GraphQL layer between backend and frontend
+
+Date: 2020-04-30
+
+## Related tickets (not available for public)
+
+SIG-2568
+
```
janjaap, line 8: Status is missing. Should have

> ## Status
> 
> Accepted

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+## Disclaimer
+
+Since this proposal is quite big the original author (jpoppe) has taken the liberty make this
+proposal a bit more opiniated then it normally should be. (One of the reasons is that the original
+author has done this implementation and fine tuned it a couple of times)
```
janjaap, line 13: No need to include this section

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Context
+
+Since our current backend only serves data through REST and there is no possibility to generate
+generate types/interfaces there is no automated way to validate data on the frontend which is coming
```
janjaap, line 18: Duplicate 'generate'

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+from the backend.
+
+Besides of that there are too many inconstencies between different backend calls and the current
```
janjaap, line 21: inconsistensies

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+backend API documentation is not always correct and/or up to date because it's written manually
+and not generated from code.
+
+Because of this our `React components` are over complicated since we need to mangle data all the
```
janjaap, line 25: 'overcomplicated' instead of 'over complicated'
Also, we don't mangle data, we transform API output to a format that the API can consume and that leads to bloat and too much client-side logic.

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+time to be able to send a POST/PUT request from data which got from a GET request.
+
+## Proposal
+
+We would like to introduce a thin layer between the backend and our frontend which should expose
+backend data through GraphQL Queries and should expose backend PUT/POST/DELETE actions via GraphQL
+mutations.
+
+We will also need to implement DTO's (Data Transfer Objects) so we can create/use data contracts
```
janjaap, line 34: Might be good to include a link to DTO definition (Stackoverflow or Wikipedia?).

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+between backend and frontend.
+
+## How
+
+Since this proposal will take quite some time/impact to implement I propose to start a POC with one
+endpoint which we could make acccessible through GraphQL.
```
janjaap, line 40: No need to mention what the workflow will look like.

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+- [https://lerna.js.org/](Lerna): A tool for creating and managing mono repos
+- [https://www.apollographql.com/](Apollo GraphQL): The industry standard GraphQL implementation
+- [https://typegraphql.com](type-graphql): A library which allows you to write GraphQL definitions in TypeScript
+- Express or Fastify: Middleware to serve our GraphQL layer
+- [https://nestjs.com](NestJS): Glue between many modern NodeJS backend implementations and tools
+
+# Frontend tools proposal
+
+- [https://graphql-code-generator.com/](GraphQL Code Generator): Generate GraphQL types/clients/interfaces, etc.
+- [https://github.com/apollographql/eslint-plugin-graphqlhl](eslint-plugin-graphql): Live GraphQL validation on the frontend
```
janjaap, line 66: What about urql?

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+- The POC focuses on the GraphQL backend implementation but the frontend should implement
+  at least one query and two mutations on the frontend
+- The POC will have the test boiler plate in place
+- The POC will have simple and concise documentation about how it has been setup
+- The POC should be able to run in a Docker container
+
+# Notes
+
+- We should be careful with caching strategies and be sure we won't cache sensitive data
+- As suggested by Roy we should take a look to DataLoader when this POC has been implemented:
```
janjaap, line 103: No need to mention sources by name

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Frontend
+
+- queries/mutations: GraphQL queries and mutations should live in one central place, for example:
+  *src/graphql/roles.queries.js* and *src/graphql/roles.mutations.js* or
+  *src/graphql/roles.graphql.js*
+
+# Proof of Concept
+
+Like already said, since this proposal is quite big it would be smart to start with a proof of concept with
```
janjaap, line 89: I would take "Like already said" out

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,104 @@
...
+mutations.
+
+We will also need to implement DTO's (Data Transfer Objects) so we can create/use data contracts
+between backend and frontend.
+
+## How
+
+Since this proposal will take quite some time/impact to implement I propose to start a POC with one
+endpoint which we could make acccessible through GraphQL.
+
```
janjaap, line 41: I would include a section 'Prerequisites' in which we describe what needs to be in place (for instance, the mono repo), before we can set up this infrastructure.

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: It's there already :smile:

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: I added a link to a StackOverFlow page which explains them related to 'NestJS'

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -2,42 +2,36 @@
...
 ## Context
 
 Since our current backend only serves data through REST and there is no possibility to generate
-generate types/interfaces there is no automated way to validate data on the frontend which is coming
-from the backend.
+types and or interfaces, there is no also no automated way to validate data on the frontend
+which is coming from the backend.
 
-Besides of that there are too many inconstencies between different backend calls and the current
+Besides of that there are too many inconstensies between different backend calls and the current
```
janjaap, line 19: Inconsistencies

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -2,42 +2,36 @@
...
 backend API documentation is not always correct and/or up to date because it's written manually
 and not generated from code.
 
-Because of this our `React components` are over complicated since we need to mangle data all the
-time to be able to send a POST/PUT request from data which got from a GET request.
+Because of this our `React components` are overcomplicated since we need to transform thr API output to a
```
janjaap, line 23: 'the' instead of 'thr'

docs/adr/0003-graphql-layer-between-backend-frontend.md
janjaap: No list of related technical implementations (mono-repo et al.) that need to be in place?

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: Thanks, will enable spell checking on my MarkDown documents so it won't happen anymore ;)

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: Thanks, will enable spell checking on my MarkDown documents so it won't happen anymore ;)

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: Yes we discussed that, did not find time yet, will do this tomorrow ;)

janjaap:
@jpoppe You need to rename the ADR file; there is already an ADR who's filename starts with 0003

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Related tickets (not available for public)
+
+SIG-2568
+
+## Context
+
+Since our current backend only serves data through REST and there is no possibility to generate
+types and or interfaces, there is no also no automated way to validate data on the frontend
+which is coming from the backend.
```
bartjkdp, line 17: I think there is an [OpenAPI 3 specification](https://api.data.amsterdam.nl/signals/swagger/openapi.yaml) for the backend. Maybe [openapi-generator](https://github.com/OpenAPITools/openapi-generator) could be used to generate types and interfaces based on this specification?

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
...
+
+Besides of that there are too many inconstencies between different backend calls and the current
+backend API documentation is not always correct and/or up to date because it's written manually
+and not generated from code.
```
bartjkdp, line 21: Django Rest Framework offers the ability to automatically generate an OpenAPI specification from the code with `python manage.py generateschema`. Maybe we can use that?

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
...
+
+Because of this our `React components` are overcomplicated since we need to transform the API output to a
+format that the API can consume and that leads to bloat and too much client-side logic.
+
+## Proposal
+
+We would like to introduce a thin layer between the backend and our frontend which should expose
+backend data through GraphQL Queries and should expose backend PUT/POST/DELETE actions via GraphQL
+mutations.
```
bartjkdp, line 30: I think that introducing a "thin layer" between the frontend and backend does actually solve the problems, but only moves the problem to the "thin layer". I am not sure how this helps anyone.

I think it would be better to solve the problems of the inconsistent API specification in the backend and/or implement a GraphQL interface directly in the backend with [Graphene](https://docs.graphene-python.org/projects/django/en/latest/) or [Ariadne](https://ariadnegraphql.org/).

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Technical requirements / proposals
+
+Our GraphQL layer:
+
+- Should be written in TypeScript so we can reuse as much as possible of our data definitions
+  and will limit duplication as much as possible
+- Should live in a Lerna mono repository so we will also be prepared to split our frontend in the
+  future and can keep all frontend related code in one place
+- Should take away as much logic from our components as possible so we could keep our
```
bartjkdp, line 44: Maybe it is good to add some references to parts in the frontend code that are currently smelly and show how the new structure will solve this?

bartjkdp:
I really like the idea of strengthening the contract between frontend and backend and use tools to automatically generate types and interfaces, but I think it is very important that this contract is implemented end-to-end and not only between the frontend and a new middleware.

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
...
+
+SIG-2568
+
+## Context
+
+Since our current backend only serves data through REST and there is no possibility to generate
+types and or interfaces, there is no also no automated way to validate data on the frontend
+which is coming from the backend.
+
+Besides of that there are too many inconstencies between different backend calls and the current
```
vanbuiten, line 19: We need to identify the issues that the FE is facing when talking to the API. If we know what these issues are we can determine steps to take and see if we can fix them in the current REST API

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
...
+
+Because of this our `React components` are overcomplicated since we need to transform the API output to a
+format that the API can consume and that leads to bloat and too much client-side logic.
+
+## Proposal
+
+We would like to introduce a thin layer between the backend and our frontend which should expose
+backend data through GraphQL Queries and should expose backend PUT/POST/DELETE actions via GraphQL
+mutations.
+
```
vanbuiten, line 31: Agree with @bartjkdp also like I said before we need to identify which problems the FE is facing and how we can tackle these problems.

Also if we decide to implement GraphQL this must live in the API and not a "thin layer" that exists between the FE and API.

docs/adr/0003-graphql-layer-between-backend-frontend.md
```diff
@@ -0,0 +1,98 @@
+# GraphQL layer between backend and frontend
+
+Date: 2020-04-30
+
+## Status
+
+**Accepted**
```
vanbuiten, line 7: Not accepted yet

docs/adr/0003-graphql-layer-between-backend-frontend.md
janjaap: Can you elaborate a bit more on why you think that the BFF should live on the back-end side of the architecture?

docs/adr/0003-graphql-layer-between-backend-frontend.md
janjaap: True. In light of recent discussions, this has not yet been accepted.

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: The `thin layer` will not solve `problems` on the frontend, however it will give us a much better `separation of concerns` which should lead to simpler and cleaner code.

It will also allow us to replace quite some custom code/logic with code/logic which is provided and maintained by the GraphQL communities, think for example about using Axios or Fetch, less selectors, all data transforming could find place in the layer, we could reuse the response objects to make mutations, etc. and we can can organize our queries and mutations in one place.

We will also get a decent amount of 'tooling' for free which will make our life easier (Think about query/mutation validation in your editor, one place where you could test or build all queries and mutations in an easy and simple way, great type/interface integration, tab completion on queries and mutations and much more.

You are correct that REST does not give us the flexibility we would like. *Graphene* or *Ariadne* would also be a `solution`. I have talked about GraphQL with the backend team a couple of times, from what I understood, it is not possible to implement it on short term because of various reasons, I think we could implement this relatively fast with this proposed ADR.

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: Great info, thanks!

I would love to have access to those schema's since it could help to make the thin `layer` described in this `ADR` better integrated with the backend.

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: Great feedback, thanks!

I have used `openapi-generator` before and at least 3 or 4 tools like that, I even attempted and actually started writing my own and did some pull requests on some of comparable projects.

In the end they kept feeling like `bandages` and `incomplete hacks` and I IMHO they only work great for kind of simple projects or a project which is frozen. The moment I solved 'my challenges' like this with a `GraphQL` layer like in this proposal I never looked back.

Even if we would use one of those tools we still have to do lots of additional work on the frontend which I think we could solve cleaner and faster with this proposed `ADR`.

docs/adr/0003-graphql-layer-between-backend-frontend.md
jpoppe: I could work on this, the issue is to describe this properly will take me quite some time, I would prefer to do this after I implemented the `POC` since that will make it easier for me. If there is really the desire to this before a POC I will need to get some time reserved and approved to work on this.

docs/adr/0003-graphql-layer-between-backend-frontend.md
bartjkdp: As the idea is to improve the quality of the frontend code, did you consider refactoring the code within the frontend codebase instead of introducing a new service?

jpoppe: > I really like the idea of strengthening the contract between frontend and backend and use tools to automatically generate types and interfaces in the frontend.
> 
> In my opinion it is very important that this contract is implemented end-to-end and not only between the frontend and a new middleware, otherwise we will probably make things worse.

Agree, but from what I understood this would be hard to achieve this (and would take lots of time) in our current backend because of some design decisions in the past. 

> I think with the current technology in place we can already strengthen the contract and automatically generate types and interfaces.

I am open for that and could even use them in this proposed `thin layer`, still they won't give us the power and simplicity of GraphQL.

> If we choose a new technlogy: GraphQL is a great candidate but we can also consider alternatives such as [gRPC(-web)](https://github.com/grpc/grpc-web).

I have worked with `gRPC` before on a extreme large scale, while it is beautiful for what it is designed for I don't think you can compare this to `GraphQL`.

GraphQL is a specification which could theoretically be used over `gRPC` as far as I know some people even started working on this. But without going to far out of scope, I would compare `gRPC` to `websockets` which is something supported by any major browser and GraphQL works through websockets and are 'normally' used for GraphQL subscriptions.

vanbuiten: @jpoppe we still need to talk about a solution. Before we can talk about solutions we need to know what issues are the FE team is facing. 

Also you mentioned GraphQL a couple of times, and this was/is not a main priority for the team. It is still to be determined if GraphQL will be the solution or that this is a nice to have in the future.

docs/adr/0003-graphql-layer-between-backend-frontend.md
bartjkdp: I agree the tooling in the GraphQL ecosystem is superior to something like openapi-generator.

docs/adr/0003-graphql-layer-between-backend-frontend.md
bartjkdp: Think I can find some time next week to create a pull request for this.

docs/adr/0003-graphql-layer-between-backend-frontend.md
bartjkdp: I would not require a detailed refactoring plan. Just a list of references to the code with some remarks would work.

I think the POC will be more succesful if we define a clear goal for the POC.

docs/adr/0003-graphql-layer-between-backend-frontend.md
bartjkdp: Did you consider refactoring the code within the current frontend codebase and add some layering there, instead of introducing a new middleware?

We are currently refactoring some code as well in the backend with Signalen/backend#7. This refactoring makes it easier to introduce a GraphQL interface in the backend.

Adding a new middleware between frontend and backend will make debugging much harder, especially without the required logging and tracing tooling in place.

bartjkdp:
I think it is a good idea to do a POC and see how GraphQL will improve the workflow and code in the frontend. For a production implementation I would still suggest to implement the GraphQL interface directly in the backend.

docs/adr/0003-graphql-layer-between-backend-frontend.md
tcoenen: @janjaap it lets us skip one transformation:
- DB -> business logic / REST -> BFF -> FE  (as proposed in this ADR)
vs
- DB -> business logic -> GraphQL -> FE
- DB -> business logic -> REST -> external partners

The business logic can than be shared between the REST interface we have to maintain and the new GraphQL interface.

tcoenen: As a PS: 
- Implementing a GraphQL interface in the backend also avoids a network hop and removes the need for caching in the BFF layer.

jpoppe:
I will pick this ADR up again as soon there is time allocated, lot's has changed since writing this ADR, for now I parked it here:
 https://github.com/Amsterdam/signals-frontend/wiki/Unfinished-GraphQL-ADR

PR #789, opened 2020-04-20
[SIG-2562] moves the `searchQuery` string to the global state

aditudorache:
This PR

- Moves the searchQuery to the global state so it becomes available in the IncidentContainer
- Navigates to the manage/incidents from the IncidentContainer when a incident search action is performed


src/containers/App/constants.js
```diff
@@ -18,3 +18,6 @@ export const UPLOAD_REQUEST = 'sia/App/UPLOAD_REQUEST';
 export const UPLOAD_PROGRESS = 'sia/App/UPLOAD_PROGRESS';
 export const UPLOAD_SUCCESS = 'sia/App/UPLOAD_SUCCESS';
 export const UPLOAD_FAILURE = 'sia/App/UPLOAD_FAILURE';
+
+export const SET_SEARCH_QUERY = 'sia/incidentManagement/SET_SEARCH_QUERY';
+export const RESET_SEARCH_QUERY = 'sia/incidentManagement/RESET_SEARCH_QUERY';
```
janjaap, line 23: Naming is incorrect

src/containers/App/reducer.js
```diff
@@ -16,7 +13,9 @@ import {
   UPLOAD_PROGRESS,
   UPLOAD_SUCCESS,
   UPLOAD_FAILURE,
-} from './constants';
+  SET_SEARCH_QUERY,
+  RESET_SEARCH_QUERY,
+} from 'containers/App/constants';
```
janjaap, line 18: I'd use `from './constants';` here to signify that it's coming from the same folder

PR #793, opened 2020-04-21
[TEST] Added test to trigger the notorious 'cannot update bug'

jpoppe:
This test should trigger the following error:
- Cannot update a component (`FilterForm`) while rendering a different component (`CheckboxList`). To locate the bad setState() call inside `CheckboxList`, follow the stack trace as described in https://fb.me/setstate-in-render


janjaap:
Closing, since this PR doesn't add any value

PR #794, opened 2020-04-21
SIG-2538 corrects the Avenir Next font name

aditudorache:
This PR corrects the Avenir Next Regular font name. Because the name didn't match the arial font was used in MapSelect and PaginationItem components

### DIFF
--- a/src/components/MapSelect/index.js
+++ b/src/components/MapSelect/index.js
@@ -26,7 +26,7 @@ const Wrapper = styled.div`
 const StyledMap = styled(Map)`
   height: 450px;
   width: 100%;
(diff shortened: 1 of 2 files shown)


src/components/MapSelect/index.js
```diff
@@ -26,7 +26,7 @@ const Wrapper = styled.div`
 const StyledMap = styled(Map)`
   height: 450px;
   width: 100%;
-  font-family: Avenir Next LT W01, arial, sans-serif;
+  font-family: "AvenirNextLTW01-Regular", arial, sans-serif;
```
janjaap, line 29: Quotes aren't necessary.

PR #796, opened 2020-04-21
[SIG-2540] Map improvements

janjaap:
This pull request tackles a number of problems:
- When the auto suggest input field is cleared, the marker isn't always removed from the map
- The `MapInput` component's `events` prop wasn't applied
- `MapInput` component didn't listen for changes in its `value` prop which made it not possible to correctly set the marker on mount or rerender
- The `Map` component allowed passing through an untestable amount of props to the '@datapunt/react-maps' `Map` component

### DIFF
--- a/src/components/MapInput/__tests__/MapInput.test.js
+++ b/src/components/MapInput/__tests__/MapInput.test.js
@@ -178,8 +178,6 @@ describe('components/MapInput', () => {
     });
   });
 
-  // it('should ')
-
   it('should handle click when a location has no address', async () => {
     const onChange = jest.fn();
     const noneFoundResponse = {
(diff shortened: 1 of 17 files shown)


src/components/MapInput/index.js
```diff
@@ -47,15 +47,17 @@ const StyledAutosuggest = styled(PDOKAutoSuggest)`
...
   const { state, dispatch } = useContext(MapContext);
   const [map, setMap] = useState();
   const [marker, setMarker] = useState();
   const { location, addressText: addressValue } = state;
-  const hasLocation = location && location?.lat !== 0 && location.lng !== 0;
+  const hasLocation = Boolean(location) && location?.lat !== 0 && location?.lng !== 0;
 
   const clickFunc = useCallback(
     async event => {
+      map.flyTo(event.latlng);
```
aditudorache, line 59: Is the intended functionality to center the map on the clicked location? The UX experience is not that pretty when the user will click on several locations after each other.

janjaap: I'll let our UX guy judge the change in behaviour.

src/components/MapInput/__tests__/MapInput.test.js
```diff
@@ -206,6 +178,8 @@ describe('components/MapInput', () => {
     });
   });
 
+  // it('should ')
```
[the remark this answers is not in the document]
janjaap: Whoops

PR #799, opened 2020-04-22
(Chore) Fix checkboxlist state update warning

janjaap:
This PR fixes a warning that was shown when the `CheckboxList` component was updated by a parent component. The underlying issue was a matter of timing. What happened was:
1. a checkbox is checked
2. `CheckboxList` internal state is updated
3. the parent component's `onChange` handler is called
4. parent component updates its internal state and sends the updated state to the `CheckboxList` component
5. `CheckboxList` internal state is updated if the incoming values differ from the internal state

Steps 4 and 5 happen short after one another which renders the parent component while the child component is rerendering. This caused the warning to show up.

The solution was to move the `onChange` handler call from the call stack to the event loop.

### DIFF
--- a/src/signals/settings/departments/Detail/components/CategoryLists/index.js
+++ b/src/signals/settings/departments/Detail/components/CategoryLists/index.js
@@ -4,11 +4,7 @@ import styled from 'styled-components';
 import { Row, themeSpacing } from '@datapunt/asc-ui';
 import isEqual from 'lodash.isequal';
 
-import {
-  ControlsWrapper,
-  Fieldset,
-  Form,
-} from 'signals/incident-management/components/FilterForm/styled';
+import { ControlsWrapper, Fieldset, Form } from 'signals/incident-management/components/FilterForm/styled';
 import Label from 'components/Label';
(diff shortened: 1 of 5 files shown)


src/signals/settings/departments/Detail/components/CategoryLists/index.js
```diff
@@ -4,11 +4,7 @@ import styled from 'styled-components';
 import { Row, themeSpacing } from '@datapunt/asc-ui';
 import isEqual from 'lodash.isequal';
 
-import {
-  ControlsWrapper,
-  Fieldset,
-  Form,
-} from 'signals/incident-management/components/FilterForm/styled';
+import { ControlsWrapper, Fieldset, Form } from 'signals/incident-management/components/FilterForm/styled';
```
aditudorache, line 7: Is anything changed in this file? I can't figure out.

janjaap: No, I guess not. At first, I made changes to this file, but I reverted them. Only formatting seems to be left.

PR #806, opened 2020-04-28
(Chore) Config setup

janjaap:
This PR contains a proposal for a setup that will allow injecting configuration into the web application at build time.

The CI/CD process will be responsible for providing the application with the necessary data to initialise it and to make settings context dependable.

Two files are required to be provided by the CI/CD:
- `site_settings.json`; application-wide settings, like URLs, titles, paths and the like
- `texts.json`; hardcoded string that appear throughout the application


bartjkdp:
Seems to have overlap with #807

janjaap:
Closing; creating separate PR for ADR

PR #821, opened 2020-05-05
[SIG-2447] AttachmentViewer keyboard nav

janjaap:
This PR adds keyboard navigation to the incident detail `AttachmentViewer` component and removes the callback function, because the component handles its own state now.

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/AttachmentViewer/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/AttachmentViewer/index.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/containers/IncidentDetail/components/AttachmentViewer/index.js
```diff
@@ -39,39 +39,73 @@ const Img = styled.img`
...
 
-const AttachmentViewer = ({ href, attachments, onShowAttachment }) => {
-  const index = attachments.findIndex(item => item.location === href);
+const AttachmentViewer = ({ href, attachments }) => {
+  const [currentHref, setCurrentHref] = useState(href);
+  const index = attachments.findIndex(item => item.location === currentHref);
   const previous = index > 0 ? attachments[index - 1].location : false;
   const next = index < attachments.length - 1 ? attachments[index + 1].location : false;
 
+  const handleKeyDown = useCallback(
```
aditudorache, line 48: Shouldn't this only work when this control has the focus? Now it would react when it is rendered and could interfere with any other control that uses keyboard events.

janjaap: Good point, I'll check what happens when the attachment viewer is open and the focus is within an input element.

janjaap: Adjusted. Thanks for pointing the issue out.

PR #822, opened 2020-05-05
[SIG-2447] Location highlight

janjaap:
This PR contains changes to inner components of the `IncidentDetail` container component and specifically to the detail `Location` component. The component is wrapped in a `HighLight` component so that, when the location is edited and saved, the change is reflected in the incident detail page with a yellow background that lights up and fades away to indicate to the user that the value has been updated.

Note that PR also contains a new component; `EditButton`. This component will be used in subsequent PRs to replace the existing buttons on the incident detail page.


janjaap:
@aditudorache You're right that in the current code the highlight isn't visible. It will be when I also push the refactoring of the incident detail container component. I split a fairly large PR into multiple smaller PRs.

PR #828, opened 2020-05-06
(Chore) Remove Jenkins stages

janjaap:
This PR removes the test and lint stages from the `Jenkinsfile` as well as test and lint container settings in the `docker-compose.yml`. The stages aren't needed since they are covered by Github actions.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,25 +10,3 @@ services:
     ipc: host
     ports:
       - "3001:80"
-
-  lint-container:
-    build:
-      cache_from:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -10,25 +10,3 @@ services:
     ipc: host
     ports:
       - "3001:80"
-
```
aditudorache, line 13: I wouldn't remove these from the docker-compose while this can be handy for local testing/solving containerization problems. No caching should be used though.

janjaap: We're running tests as well as the linter directly on the command line. The settings in `docker-compose` have only been used by Jenkins. In which situations do you see a use for these config settings here?

PR #835, opened 2020-05-07
[SIG-2565] fix backlink when navigating from the map overview

aditudorache:
This PR ensures that the back link in the incident detail always navigate to the incidents overview excepting the situation when the incident detail is opened from the overview map. In the last case the back link will return to the overview map of incidents

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
@@ -88,7 +88,9 @@ const DetailHeader = ({ incident, baseUrl, onPatchIncident }) => {
     },
   };
 
-  const referrer = location.referrer || `${baseUrl}/incidents`;
+  const mapOverviewLocation = '/manage/incidents/kaart';
+
+  const referrer = location.referrer === mapOverviewLocation ? mapOverviewLocation : `${baseUrl}/incidents`;
 
   return (
     <Header className="detail-header">
--- a/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
@@ -89,7 +89,7 @@ const DetailHeader = ({ incident, baseUrl, onPatchIncident }) => {
     },
   };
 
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
```diff
@@ -88,7 +88,9 @@ const DetailHeader = ({ incident, baseUrl, onPatchIncident }) => {
     },
   };
 
-  const referrer = location.referrer || `${baseUrl}/incidents`;
+  const mapOverviewLocation = '/manage/incidents/kaart';
```
janjaap, line 91: There is a named export for this route in `routes.js` in the `IncidentManagement` module.

src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
```diff
@@ -88,7 +88,9 @@ const DetailHeader = ({ incident, baseUrl, onPatchIncident }) => {
...
+
+  const referrer = location.referrer === mapOverviewLocation ? mapOverviewLocation : `${baseUrl}/incidents`;
```
janjaap, line 93: What if the referrer isn't exactly the route that you're checking for, but, for instance, contains a hash in the URL?

src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
aditudorache: Awesome! Thanks!

src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
aditudorache: It would navigate to the incidents list.

janjaap: Exactly. And that wouldn't be advisable. We still have a ticket on the board that describes having all filter, pagination and sorting params in the URL. With the current rigid check, that would fail.

aditudorache: Ok, I will check only on the containing of the path then. Didn't know about that ticket.

PR #839, opened 2020-05-07
[SIG-2514] adds an incident subtitle to the incident form

aditudorache:
This PR adds extra information to the incident form

### DIFF
--- a/src/signals/incident/definitions/wizard-step-1-beschrijf.js
+++ b/src/signals/incident/definitions/wizard-step-1-beschrijf.js
@@ -57,6 +57,7 @@ export default {
       description: {
         meta: {
           label: 'Waar gaat het om?',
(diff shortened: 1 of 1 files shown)


src/signals/incident/definitions/wizard-step-1-beschrijf.js
```diff
@@ -57,6 +57,7 @@ export default {
       description: {
         meta: {
           label: 'Waar gaat het om?',
+          subtitle: 'Typ geen persoonsgegevens in deze omschrijving, dit wordt apart gevraagd',
           path: 'text',
           placeholder: 'Beschrijf uw melding',
```
janjaap, line 62: Shouldn't the placeholder be removed?

aditudorache: Good catch! Thanks!

PR #844, opened 2020-05-11
Cypress tests for SIG-2380, change signal after submit

arjandata:
Cypress tests for SIG-2380, change signal after submit. Scenario's:
* Change Location
* Change Status
* Change Urgency
* Change Type
* Change Category

### DIFF
--- a/cypress/integration/changeSignalAfterSubmit.spec.js
+++ b/cypress/integration/changeSignalAfterSubmit.spec.js
@@ -1,0 +1,324 @@
+// <reference types="Cypress" />
+import * as createSignal from '../support/commandsCreateSignal';
+import { CREATE_SIGNAL } from '../support/selectorsCreateSignal';
+import { CHANGE_CATEGORY, CHANGE_LOCATION, CHANGE_STATUS, CHANGE_TYPE, CHANGE_URGENCY, SIGNAL_DETAILS } from '../support/selectorsSignalDetails';
+import { MANAGE_SIGNALS, OVERVIEW_MAP } from '../support/selectorsManageIncidents';
+describe('Change signal after submit', () => {
+  describe('Create signal graffitie', () => {
+    before(() => {
+      cy.server();
+      cy.defineGeoSearchRoutes();
+      cy.getAddressRoute();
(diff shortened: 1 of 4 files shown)


cypress/integration/changeSignalAfterSubmit.spec.js
```diff
@@ -0,0 +1,324 @@
...
+    before(() => {
+      cy.server();
+      cy.defineGeoSearchRoutes();
+      cy.getAddressRoute();
+
+      cy.visitFetch('incident/beschrijf');
+    });
+
+    it('Should search for an address', () => {
+    // Search on address
```
jpoppe, line 17: Not blocking but I would add 2 space before '//' so your comments are consistent, I think I have an ESLint rule for this, so feel free to ignore, will fix this when I will add the rule ;)

cypress/integration/changeSignalAfterSubmit.spec.js
```diff
@@ -0,0 +1,324 @@
...
+
+      cy.clickButton('Verstuur');
+   
+      cy.wait('@postSignalPublic');
+
+      cy.url().should('include', '/incident/bedankt');
+    
+      // Capture signal id
+      cy.get('.bedankt').first().then($signalLabel => {
+      // Get the signal id
```
jpoppe, line 73: Same here..

cypress/integration/changeSignalAfterSubmit.spec.js
```diff
@@ -0,0 +1,324 @@
...
+  describe('Create signal graffitie', () => {
+    before(() => {
+      cy.server();
+      cy.defineGeoSearchRoutes();
+      cy.getAddressRoute();
+
+      cy.visitFetch('incident/beschrijf');
+    });
+
+    it('Should search for an address', () => {
```
janjaap, line 16: There is a lot of repetition in tests. Take searching for an address, entering a description, phone number, email address and such. If the structure of the form changes (and it will), then all of those tests need to be revised and fixed. I would extract any repeating patterns into functions where possible.

PR #846, opened 2020-05-11
[SIG-2662] disable the `Locatie opslaan` button while waiting for the location api calls.

aditudorache:
This PR disables the `Locatie opslaan` button while waiting for the location api call. The `LocationInput` had to be extracted as a separate component in order to comply to the `rules of hooks`.


src/containers/MapContext/reducer.js
```diff
@@ -28,6 +29,7 @@ export default (state, action) => {
       return {
         ...state,
         location: action.payload,
+        loading: true,
```
janjaap, line 32: Shouldn't `RESET_LOCATION` set `loading` back to 'false'?

src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -199,19 +184,18 @@ export class IncidentDetail extends React.Component {
...
                     )}
 
                     {previewState === 'editLocation' && (
-                      <LocationForm
-                        incidentId={incident.id}
-                        location={incident.location}
-                        onPatchIncident={onPatchIncident}
-                        onClose={this.onCloseAll}
-                      />
+                      <MapContext>
```
janjaap, line 191: Isn't it possible to wrap the `LocationForm` component's `render()` function return value with `MapContext`? Seems a bit odd to have to wrap at at this level.

src/test/utils.js
```diff
@@ -98,3 +99,6 @@ export const userObjects = (users = usersJSON) =>
  * @returns {Promise} Resolved promise
  */
 export const resolveAfterMs = timeMs => new Promise(resolve => setTimeout(resolve, timeMs));
+
+
+export const withMapContext = Component => withAppContext(<MapContext>{Component}</MapContext>);
```
janjaap, line 104: 👍

src/containers/MapContext/reducer.js
```diff
@@ -39,6 +41,7 @@ export default (state, action) => {
     case SET_VALUES:
       return {
         ...state,
+        loading: false,
```
janjaap, line 44: What if `setLocationAction` is called, but `setValuesAction` isn't? There isn't an indicator anywhere in the code that tells me that I need to do that in order to set the `loading` prop. Also, you can set the location just by calling `setValuesAction`, which bypasses the `loading` prop altogether.
I think it would be better to have a separate action to set and unset the `loading` prop instead of hiding it in another action.

src/signals/incident-management/containers/IncidentDetail/index.js
aditudorache: Good point. I'll change that.

src/containers/MapContext/reducer.js
aditudorache: I agree. Is better to have it separate.

src/containers/MapContext/reducer.js
aditudorache: I have decided to move the loading event to a separate action and remove it here.

src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -171,12 +160,7 @@ export class IncidentDetail extends React.Component {
               {incident && (
                 <Row>
                   <Column span={12}>
-                    <DetailHeader
-                      incidentId={incident.id}
-                      status={incident?.status?.state}
-                      links={incident?._links}
-                      onPatchIncident={onPatchIncident}
-                    />
+                    <DetailHeader incident={incident} baseUrl="/manage" onPatchIncident={onPatchIncident} />
```
janjaap, line 163: This change reverts a change that I applied in a previous PR. This won't work.

aditudorache: Good point. Changed.

PR #807, opened 2020-04-28
Basic config

boris-arkenaar:
This will fix Signalen/frontend#2, allowing us to configure the logo, colors, and some texts on the website.

* Configuration for these settings can be injected through a JSON file during deployment of the docker container. A `start.sh` has been added to facilitate injecting these settings into the `index.html`.
* For the development server an optional `config.json` in the root of the project will be used to inject into the `index.html`. This is done in the Webpack configuration.
* The existing configuration service is being used. All config from the injected JSON file will be added to the configuration in this service.
* An extra property `all` is added to the configuration service to retrieve all config as an object (except for all the endpoints provided through getters).

- [x] Configuration is possible through a JSON file
- [x] Configuration is injected into index.html (dev server)
- [x] Configuration is injected into index.html (docker deploy)
- [x] Logo is configurable
- [x] Theming variables, as accepted by [Amsterdam/amsterdam-styled-components](https://github.com/Amsterdam/amsterdam-styled-components), is configurable
- [x] Site's footer text is configurable

Before opening a pull request, please ensure:

- [x] double-check your branch is based on `develop` and targets `develop` 
- [x] Pull request has tests (we are going for 100% coverage!)
- [x] Code is well-commented, linted and follows project conventions
- [x] Internal code generators and templates are updated (if necessary)

Be kind to code reviewers, please try to keep pull requests as small and focused as possible :)

PR #853, opened 2020-05-14
Move Matomo and Sentry settings to configuration

boris-arkenaar:
Moved the settings for Matomo and Sentry to the configuration (`environment.conf.json`).
The `siteId` to use for Matomo is now based on the environment (`13` in `environment.conf.prod.json`, `14` otherwise) instead of the hostname. We need to make sure this works correctly.

In case the settings are not present in the configuration, the corresponding service will silently not be initialized.

This PR is based no #807 so it should only be merged after that one has been merged.


src/app.js
```diff
@@ -24,13 +25,15 @@ import './polyfills';
 import configureStore from './configureStore';
 
 const environment = process.env.NODE_ENV;
+const dsn = configuration.sentry && configuration.sentry.dsn;
```
janjaap, line 28: You can use optional chaining:
```
const dsn = configuration?.sentry?.dsn;
```

src/app.js
```diff
@@ -40,13 +43,15 @@ const MOUNT_NODE = document.getElementById('app');
...
 // Setup Matomo
-const hostname = window && window.location && window.location.hostname;
-const MatomoInstance = new MatomoTracker({
-  urlBase: 'https://analytics.data.amsterdam.nl/',
-  siteId: hostname === 'meldingen.amsterdam.nl' ? 13 : 14,
-});
-
-MatomoInstance.trackPageView();
+const urlBase = configuration.matomo && configuration.matomo.urlBase;
+const siteId = configuration.matomo && configuration.matomo.siteId;
```
janjaap, line 47: Optional chaining here as well.

src/app.js
```diff
@@ -24,13 +25,15 @@ import './polyfills';
...
 const environment = process.env.NODE_ENV;
+const dsn = configuration.sentry && configuration.sentry.dsn;
 const release = process.env.GIT_COMMIT;
-
-Sentry.init({
-  environment,
-  dsn: 'https://3de59e3a93034a348089131aa565bdf4@sentry.data.amsterdam.nl/27',
-  release,
-});
+if (dsn) {
```
janjaap, line 30: When this condition doesn't meet, then the Sentry module also shouldn't be imported. That also goes for Matomo. We could go for a `noop` module replacement in that case. That should then be applied to the Webpack settings.

janjaap: Also, in multiple places in the code the `captureException` method of the Sentry instance is called, so that shouldn't break either.

boris-arkenaar: While not the cleanest solution, it doesn't pose a problem to import Sentry and call `captureException` when it's not initialized.

boris-arkenaar: The `MatomoTracker` is only called in `app.js` within that if-statement.

PR #858, opened 2020-05-15
Improve security and configurability of OpenID

bartjkdp:
## Improvements

- Light refactor of the authentication code.
- Make OpenID settings more flexible so the frontend can be configured with other Identity Providers as well.
- Add logic for validating the nonce in the token. In order to mitigate replay attacks we should do this. This is also required by the [spec](https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes).

## Todo
- [x] Rename `AUTH_ROOT` to `OIDC_AUTH_ENDPOINT`, so the solution works for Identity Providers with different authorization endpoints.
- [x] Make `OIDC_CLIENT_ID` configurable.
- [x] Audit all security aspects of OpenID authentication response validation, some parts are still missing.
- [x] Test with [Dex](https://github.com/dexidp/dex).
- [x] Test if changes work in acceptance and production environment of Amsterdam.

## Breaking changes

This change breaks existing deployments. Make sure to:

- Set `OIDC_CLIENT_ID`, `OIDC_RESPONSE_TYPE` and `OIDC_SCOPE`.
- Update `AUTH_ROOT` to `OIDC_AUTH_ENDPOINT`.

Closes Signalen/frontend#17


bartjkdp:
Considered using a library for implementing OpenID Connect, but decided to stick with the current implementation as there are some Amsterdam-specific additions that are not covered in the OpenID Connect spec.

src/shared/services/auth/auth.test.js
```diff
@@ -306,18 +331,15 @@ describe('The auth service', () => {
...
-      savedReturnPath = '/path/leading/back';
+      savedStateToken = 'random-string';
 
       initAuth();
-      expect(getReturnPath()).toBe(savedReturnPath);
     });
 
     it('returns an empty string when the callback was unsuccessful', () => {
       initAuth();
-      expect(getReturnPath()).toBe('');
```
boris-arkenaar, line 320: Not testing here anything anymore. Implicitly testing that it doesn't throw? You could make that explicit?

bartjkdp: Removed the whole function `getReturnPath()` from the codebase as it is not used any more. This part of 314a25a3491e8fd6eaf8fc12bd98cd3137c62364

aditudorache:
Hi @bartjkdp, I can't get it working locally with the Amsterdam configuration. Can you please assist?

bartjkdp:
@aditudorache sure! Just sent you an invite to our Slack.

bartjkdp:
@janjaap @aditudorache thanks for testing. This will probably require some changes at the side of the Identity Provider or some more configurable options in the code. I will do some experimenting with authz and submit some updates.

bartjkdp:
I did some research on [authz](https://github.com/Amsterdam/authz). It currently does not support the id_token response type so I decided to make `OIDC_RESPONSE_TYPE` and `OIDC_SCOPE` configurable.

This update adds two more parameters that need to be added to the deployments:

- OIDC_RESPONSE_TYPE
- OIDC_SCOPE

aditudorache:
@janjaap, it works now for both authentication methods `autz` and `grip`. Can you check this as well before merging?

bartjkdp:
@janjaap @aditudorache please do not forget to add the following settings to acceptance and production:

- OIDC_RESPONSE_TYPE
- OIDC_SCOPE

PR #859, opened 2020-05-18
[SIG-2692] role checkbox

aditudorache:
This PR fixes a bug that prevents unchecking of the roles in the role settings page

### DIFF
--- a/src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
+++ b/src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
@@ -99,7 +99,7 @@ describe('/signals/settings/roles/components/RoleForm', () => {
     expect(props.onPatchRole).toHaveBeenCalledWith({
       id: 2,
       name: 'Behandelaar',
-      permission_ids: [permissionsJson[0].id, 110, 164, 162, 163, 161, 111],
+      permission_ids: [permissionsJson[0].id, ...rolesJson[0].permissions.map(p => p.id)],
     });
 
     act(() => {
@@ -115,7 +115,7 @@ describe('/signals/settings/roles/components/RoleForm', () => {
     expect(props.onPatchRole).toHaveBeenCalledWith({
       id: 2,
       name: 'Behandelaar',
-      permission_ids: [110, 164, 162, 163, 161, 111],
+      permission_ids: [...rolesJson[0].permissions.map(p => p.id)],
     });
   });
 
--- a/src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
+++ b/src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, act } from '@testing-library/react';
 import { withAppContext } from 'test/utils';
 import * as reactRouterDom from 'react-router-dom';
 
@@ -28,14 +28,11 @@ describe('/signals/settings/roles/components/RoleForm', () => {
   });
 
   it('should render correctly', () => {
-    const { container, queryByTestId } = render(
-      withAppContext(<RoleForm {...props} />)
-    );
+    const { container, queryByTestId } = render(withAppContext(<RoleForm {...props} />));
 
     expect(queryByTestId('rolesFormFieldName')).toHaveValue('Behandelaar');
 
-    expect(container.querySelectorAll('input[type="checkbox"]').length)
-      .toBe(permissionsJson.length);
+    expect(container.querySelectorAll('input[type="checkbox"]').length).toBe(permissionsJson.length);
 
     expect(container.querySelectorAll('input[type="checkbox"]:checked').length).toBe(6);
   });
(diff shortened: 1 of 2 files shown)


janjaap:
@aditudorache Can you fix conflicts?

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
```diff
@@ -84,9 +77,28 @@ describe('/signals/settings/roles/components/RoleForm', () => {
     };
 
     expect(props.onPatchRole).not.toHaveBeenCalled();
-
     fireEvent.change(getByTestId('rolesFormFieldName'), event);
```
janjaap, line 80: This event should also be wrapped in `act`

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
```diff
@@ -84,9 +77,28 @@ describe('/signals/settings/roles/components/RoleForm', () => {
...
 
+    act(() => {
+      fireEvent.click(container.querySelectorAll('input[type="checkbox"]')[0]);
+    });
+
+    await findByTestId('rolesFormFieldName');
+
+    fireEvent.click(getByTestId('submitBtn'), { preventDefault: jest.fn() });
```
janjaap, line 88: Event should be wrapped in `act`

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
```diff
@@ -84,9 +77,28 @@ describe('/signals/settings/roles/components/RoleForm', () => {
...
+      permission_ids: [permissionsJson[0].id, 110, 164, 162, 163, 161, 111],
+    });
+
+    act(() => {
+      fireEvent.click(container.querySelectorAll('input[type="checkbox"]')[0]);
+    });
+
+    await findByTestId('rolesFormFieldName');
+
     fireEvent.click(getByTestId('submitBtn'), { preventDefault: jest.fn() });
```
janjaap, line 102: Event should be wrapped in `act`

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
```diff
@@ -84,9 +77,28 @@ describe('/signals/settings/roles/components/RoleForm', () => {
...
+    });
+
+    await findByTestId('rolesFormFieldName');
+
+    fireEvent.click(getByTestId('submitBtn'), { preventDefault: jest.fn() });
+
+    expect(props.onPatchRole).toHaveBeenCalledWith({
+      id: 2,
+      name: 'Behandelaar',
+      permission_ids: [permissionsJson[0].id, 110, 164, 162, 163, 161, 111],
```
janjaap, line 93: Please refrain from using hardcoded values. Instead, use references to props in the fixture. This prevents the test from breaking in the future when the fixture data is replaced or updated.

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/index.js
```diff
@@ -27,6 +27,17 @@ const StyledInput = styled(Input)`
 export const RoleForm = ({ role, permissions, onPatchRole, onSaveRole, readOnly }) => {
   const formRef = useRef(null);
   const { isValid, validate, errors, event } = useFormValidation(formRef);
+  const [rolePermissions, setRolePermissions] = useState(role.permissions);
+
+  const handleChange = useCallback(id => onChangeEvent => {
+    if (!onChangeEvent) return;
```
janjaap, line 33: Can there be such an occasion when there is `onChangeEvent` value? If so, can you add a test case for it?

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/index.js
```diff
@@ -27,6 +27,17 @@ const StyledInput = styled(Input)`
 export const RoleForm = ({ role, permissions, onPatchRole, onSaveRole, readOnly }) => {
   const formRef = useRef(null);
   const { isValid, validate, errors, event } = useFormValidation(formRef);
+  const [rolePermissions, setRolePermissions] = useState(role.permissions);
```
janjaap, line 30: Since the state value `rolePermissions` contains an array of objects, wouldn't it be better to use `useReducer` for this instead of `useState`?

aditudorache: While the array is used as a whole and there is no different way of modifying the data but replacing (no different event-state relation), I don't see the point.

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/__test__/RoleForm.test.js
aditudorache: done

src/signals/settings/roles/containers/RoleFormContainer/components/RoleForm/index.js
aditudorache: Actualy not, good catch. I removed that!

PR #863, opened 2020-05-18
[SIG-2724] showing the correct questions

janjaap:
This PR:
- Removes all the options listed in the wizard samenvatting. Some of the entries were incorrect, not used or didn't have a label which caused them not to be shown in the summary. Instead, all control options are imported from their corresponding files.
- Separately exports the controls for the wizard's step 2 options, so that they can be listed in the summary without having to hard code each and every one of them
- Adds the `shortLabel` prop to controls where they were missing
- Adds configuration for the `wonen-overig` category list of radio buttons so that, when one of the options is selected, the incident state will be (partially) cleared. This will prevent questions from being shown that should not be shown.

Mind you, there's only twelve files changed, but the files have a lot of line, so the overall impact of this PR is bigger than might seem at first glance.


janjaap:
Apologies for another review request; had to fix some issues and merge conflicts.

PR #874, opened 2020-05-26
[SIG-2706] Display the departments in the subcategory name 

aditudorache:
This PR Changes the MetaList subcategory so it displays the departments for each subcategory. When the text is too long the text overflow is marked with ellipsis. 
Mind that in edit mode the Select stylecomponent used is not working with ellipsis yet because o a bug in the component, this change will be introduced as soon as this will be fixed in `asc`.


janjaap:
@aditudorache The title of the PR is confusing; the code doesn't contain changes to a selector.
Wouldn't it be better to have the subcategories selector return an enriched dataset where the department names are appended to the category name as an extra prop?

src/signals/incident-management/containers/IncidentDetail/components/MetaList/components/ChangeValue/index.js
```diff
@@ -22,6 +22,14 @@ const EditButton = styled(Button)`
...
 `;
 
+const DisplayValue = styled.span`
+    display: inline-block;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    max-width: calc(100% - ${themeSpacing(10)});
+`;
+
```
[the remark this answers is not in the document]
aditudorache: Good catch!

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
```diff
@@ -44,9 +44,21 @@ const EditButton = styled(Button)`
   padding: ${themeSpacing(0, 1.5)};
 `;
 
+const getCategoryName = ({ name, departments }) => {
+  const departmensStringList = departments?.length ? ` (${departments.map(({ code }) => code).join(',')})` : '';
+  return `${name}${departmensStringList}`;
+};
+
```
janjaap, line 51: Can you add a test case that verifies the addition of the concatenated department names to the subcategory?

aditudorache: done

PR #884, opened 2020-05-27
[SIG-2573] QA-hooks

janjaap:
This PR adds following QA-hooks by setting `data-testid` attributes on HTML elements. The following have been added:
- `mapZoom`; wrapper around map zoom controls. It wasn't possible to target the individual buttons.

- `markerClusterIcon`; icons of clustered incidents.

- `meta-list-type-definition`, `meta-list-type-value`, `meta-list-priority-definition`, `meta-list-priority-value`, `meta-list-status-definition` and `meta-list-status-value` in incident detail meta data column

- `editTypeButton`, `editPriorityButton`, `editSubcategoryButton` and `editStatusButton` in incident detail meta data column for individual meta values. `editLocationButton` In detail section.

- `submitTypeButton`, `cancelTypeButton`, `submitPriorityButton`, `cancelPriorityButton`, `editSubcategoryButton` and `cancelSubcategoryButton` for individual meta value forms.

- All checkboxes in lists have gotten a unique `data-testid` attribute value in the form of `checkbox-{objectType}_{objectId}` where `objectType` is a string and `objectId` a numeric value  (except for subcategories where it is a string, since that is the PK were working with)


src/signals/incident-management/containers/IncidentDetail/components/MetaList/components/ChangeValue/index.test.js
```diff
@@ -10,22 +10,18 @@ import ChangeValue from './index';
...
 
-  expect(valuePath).toBeInTheDocument();
   expect(editButton).toBeInTheDocument();
   expect(queryByTestId('changeValueForm')).not.toBeInTheDocument();
 };
 
 const expectEditState = async ({ queryByTestId, findByTestId }) => {
-  const editButton = queryByTestId('editButton');
-  const valuePath = queryByTestId('valuePath');
+  const editButton = queryByTestId('editMockTypeButton');
```
aditudorache, line 21: The construction of the mock name is difficult to understand at the first glance because the name is constructed dynamically. I would choose a more straight forward solution where the test-id name is either explicit set 
on the button, or it can be selected through the meta-list selector.

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.test.js
```diff
@@ -70,13 +70,13 @@ describe('<MetaList />', () => {
 
     it('should call onPatchIncident', async () => {
       const { getAllByTestId } = render(withAppContext(<MetaList {...props} />));
-      const editButtons = getAllByTestId('editButton');
+      const editButtons = getAllByTestId('editPriorityButton');
```
aditudorache, line 73: Same here. Cost some time to figure out where the test-id comes from. We should find a way to easy figure out the id element which is selected

src/signals/incident-management/containers/IncidentDetail/components/MetaList/components/ChangeValue/index.test.js
janjaap: The issue here is that when I use a generic test id value, there is no way to distinguish between the elements in e2e tests. That is why their values are dynamically generated based on their context. In the case of the `ChangeValue` component, that is the incoming data.

PR #888, opened 2020-05-27
(Chore) amsterdam-stijl Removal helper components

janjaap:
This PR adds components that will replace specific styling, now supplied by the `amsterdam-stijl` package. In a follow-up pull request, these components will be applied.

### DIFF
--- a/src/components/ButtonBar/index.js
+++ b/src/components/ButtonBar/index.js
@@ -1,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { themeSpacing } from '@datapunt/asc-ui';
+
+const Bar = styled.div`
+  & > * + * {
+    margin-left: ${themeSpacing(2)};
+  }
+`;
+
+/**
+ * Button container that merely adds margin between children
+ */
+const ButtonBar = ({ className, children }) => <Bar className={className}>{children}</Bar>;
+
+ButtonBar.defaultProps = {
+  className: '',
+};
+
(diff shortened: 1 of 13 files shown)


src/components/Checkbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+import { Checkbox as AscCheckbox, themeSpacing } from '@datapunt/asc-ui';
+
+const Wrapper = styled.span`
+  display: inline-block;
+  background: white;
+  padding-right: ${themeSpacing(2)};
+  background-clip: content-box;
+`;
+
+const StyledBox = styled(AscCheckbox)`
```
aditudorache, line 13: `CheckboxStyle` would be more readable

src/components/ButtonBar/index.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { themeSpacing } from '@datapunt/asc-ui';
+
+const Bar = styled.div`
```
aditudorache, line 6: `ButtonBarStyle`  would be more readable

src/components/Checkbox/index.js
janjaap: But it's not style, it's a component.

src/components/ButtonBar/index.js
janjaap: Same here, the result of a call to `styled` is a component, not a list of styled, which the name `ButtonBarStyle` applies

PR #891, opened 2020-05-27
Feature/16 questions

boris-arkenaar:
# Dynamic complementary questions

Any complementary questions for a category/subcategory will be fetched from the backend, directly after the prediction results are in.

Fixes Signalen/Frontend#16

## To test it

- Add some questions to the backend via the Jango Admin.
- Turn on the `fetchQuestionsFromBackend` feature flag in `environment.conf.development.json`.
- On the website type the right keywords in the 'Waar gaat het om' textarea to trigger the category you've added the questions to.

## What it does

- The result from the backend (array with objects) is converted to an expected format for the frontend (object with keys) in [`src/signals/incident/containers/IncidentContainer/services.js`](https://github.com/Amsterdam/signals-frontend/pull/891/files#diff-111ed26cd77926d56ccfcc6f5a34e33b).
- The resolved questions object is saved to the store. This means that the `render` attribute only has a string naming the component that should render, and the Validators are also still just strings instead of validator functions.
- In [`src/signals/incident/definitions/wizard-step-2-vulaan.js`](https://github.com/Amsterdam/signals-frontend/pull/891/files#diff-08e0571f4c1f8182692400733f905b80) the configuration is expanded to replace the `render` strings with actual components and validator strings with actual validator functions. This function (`expandQuestions`) is `memoize`d (using lodash) with a custom cache key based on category and subcategory. This is needed for it to work properly with `react-reactive-form`.

## Additional remarks

- ~Backend has yet to be implemented. For now, no fetch call is actually done. Only a temporary function is called, which will return a static results object for testing purposes. (`fetchQuestions` in [`src/signals/incident/containers/IncidentContainer/saga.js`](https://github.com/Amsterdam/signals-frontend/pull/891/files#diff-a33b056fbcfb297a2413a7185b0178e0R40).)~
- Fully backward compatible with the current hardcoded categories of Amsterdam by the use of a feature flag (`fetchQuestionsFromBackend`) that's turned off by default.
- It's not possible anymore to pass on [complex component hierarchies](https://github.com/Amsterdam/signals-frontend/blob/develop/src/signals/incident/definitions/wizard-step-2-vulaan/overlast-van-dieren.js#L27-L51) to the `PlainText` form component, only strings. The suggestion is to create a markdown form field which can render markdown. That way formatting can easily be created and transferred.

## What still needs to be done

- [x] Implement call to questions endpoint.
- [x] An exhaustive [mapping to all form components and validator functions](https://github.com/Amsterdam/signals-frontend/pull/891/files#diff-111ed26cd77926d56ccfcc6f5a34e33bR1-R13).

## SIA

- [x] double-check your branch is based on `develop` and targets `develop` 
- [x] Internal code generators and templates are updated (if necessary)


src/signals/incident/definitions/wizard-step-2-vulaan.js
```diff
@@ -1,4 +1,8 @@
+import { Validators } from 'react-reactive-form';
+import memoize from 'lodash/memoize';
```
janjaap, line 2: Did you come across https://dev.to/nioufe/you-should-not-use-lodash-for-memoization-3441? Maybe we'd better replace if with `defaultMemoize` from `reselect`. A package we're also using in this application.

src/signals/incident/containers/IncidentContainer/services.js
```diff
@@ -0,0 +1,64 @@
...
+  min_length: 'minLength',
+  pattern: 'pattern',
+  required: 'required',
+  required_true: 'requiredTrue',
+};
+
+const mapValidatorWithArgs = ([key, ...args]) => [validatorMap[key], ...args];
+const mapValidator = key => (Array.isArray(key) ? mapValidatorWithArgs(key) : validatorMap[key]);
+
+export const resolveQuestions = questions => ({
```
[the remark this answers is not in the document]
boris-arkenaar: The default export always confuses me. This file could potentially have more service functions, so then you would need to export it like this.

src/signals/incident/containers/IncidentContainer/reducer.js
```diff
@@ -38,6 +42,7 @@ export const initialState = fromJS({
     },
   },
   loadingClassification: false,
+  loadingQuestions: false,
```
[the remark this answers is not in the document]
boris-arkenaar: True

src/signals/incident/definitions/wizard-step-2-vulaan.js
boris-arkenaar: I don't see how I can create a custom cache key with `defaultMemoize` from reselect, like I do with `memoize` from lodash. Memoizing on the `questions` argument doesn't work since it changes all the time. I need to check for the combination of category/subcategory.
Besides that, I don't expect a memory leak since no unique IDs are being used for instance. There will only ever be a very limited set of category/subcategory pairs.

src/signals/incident/containers/IncidentContainer/reducer.js
```diff
@@ -107,15 +106,13 @@ export default (state = initialState, action) => {
       return state.set('incident', getIncidentWithoutExtraProps(state.get('incident'), action.payload));
 
     case GET_QUESTIONS:
-      return state.set('loadingQuestions', true);
+      return state;
```
janjaap: This case doesn't need to be here

src/signals/incident/containers/IncidentContainer/reducer.js
```diff
@@ -107,15 +106,13 @@ export default (state = initialState, action) => {
...
 
     case GET_QUESTIONS_SUCCESS:
-      return state
-        .set('loadingQuestions', false)
-        .set('incident', state.get('incident').set('questions', action.payload.questions));
+      return state.set('incident', state.get('incident').set('questions', action.payload.questions));
 
     case GET_QUESTIONS_ERROR:
-      return state.set('loadingQuestions', false);
+      return state;
```
janjaap: Same here; case is not needed

src/signals/incident/containers/IncidentContainer/reducer.test.js
```diff
@@ -275,20 +275,19 @@ describe('signals/incident/containers/IncidentContainer/reducer', () => {
...
+    it('does nothing', () => {
       expect(
         incidentContainerReducer(fromJS({ incident: {} }), {
           type: GET_QUESTIONS,
         }).toJS()
       ).toEqual({
         incident: {},
-        loadingQuestions: true,
       });
     });
```
janjaap: Safe to remove; test case doesn't verify any thing

src/signals/incident/containers/IncidentContainer/reducer.test.js
```diff
@@ -309,13 +308,12 @@ describe('signals/incident/containers/IncidentContainer/reducer', () => {
             key1: {},
           },
         },
-        loadingQuestions: false,
       });
     });
   });
 
   describe('GET_QUESTIONS_ERROR', () => {
```
janjaap, line 315: Test case can be removed since it doesn't make any valuable assertions

PR #912, opened 2020-06-04
[SIG-1741] minor fix for empty predictions in the incident form

aditudorache:
This PR prevents that an unnecessary api prediction call is made when the decription text area is left empty in the IncidentForm


src/signals/incident/components/form/DescriptionWithClassificationInput/index.js
```diff
@@ -7,7 +7,7 @@ import Header from '../Header';
 import './style.scss';
 
 function get(e, meta, parent) {
-  parent.meta.getClassification(e.target.value);
+  if(e.target.value) parent.meta.getClassification(e.target.value);
```
jpoppe, line 10: A space is missing after `if`, will add an ESLint rule for this later.

PR #917, opened 2020-06-04
[SIG-2452] delete dashboard container

aditudorache:
This PR removes the unused DashboardContainer component together with the related packages that are not used anymore.


janjaap:
There is more to be removed:  `DASHBOARD_ENDPOINT` in the configuration. Both in this repo and `signals-amsterdam` and `signals-weesp`

environment.conf.json
```diff
@@ -19,19 +19,35 @@
...
-        [52.50536, 5.10737]
+        [
+          52.25168,
+          4.64034
+        ],
+        [
+          52.50536,
+          5.10737
+        ]
+      ],
```
jpoppe, line 46: I think the none expanded `arrays` are more readable.

PR #926, opened 2020-06-05
[SIG-2775] update asc packages

aditudorache:
Thie PR updates the `asc` packages to 0.20.0 and solves a breaking change around the the site Logo. A SIA variant of the Amsterdam Logo is added to this project because the default logo provided by asc is not according to the design of SIA.

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -4,6 +4,7 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
 const devMode = process.env.NODE_ENV !== 'production';
+const __rootdir = process.cwd();
 
 module.exports = options => ({
   mode: options.mode,
@@ -17,7 +18,7 @@ module.exports = options => ({
   // eslint-disable-next-line prefer-object-spread
   output: Object.assign(
     {
-      path: path.resolve(process.cwd(), 'build'),
+      path: path.resolve(__rootdir, 'build'),
       publicPath: '/',
     },
     options.output
@@ -138,6 +139,10 @@ module.exports = options => ({
     modules: ['node_modules', 'src'],
     extensions: ['.js', '.jsx', '.react.js'],
     mainFields: ['browser', 'jsnext:main', 'main'],
+    alias: {
+      '@datapunt/asc-ui': path.resolve(__rootdir, 'node_modules/@datapunt/asc-ui/lib'),
+      '@datapunt/asc-assets': path.resolve(__rootdir, 'node_modules/@datapunt/asc-assets/lib'),
+    },
   },
   devtool: options.devtool,
   target: 'web', // Make web variables accessible to webpack, e.g. window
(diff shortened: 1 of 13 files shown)


src/components/AmsterdamLogo/index.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { LogoShort } from '@datapunt/asc-assets';
+import styled, { css } from 'styled-components';
+import { breakpoint, themeSpacing, themeColor } from '@datapunt/asc-ui';
+
+export const srOnlyStyle = () => ({ srOnly }) =>
+  srOnly
```
janjaap, line 7: I would be enough to do
```jsx
({ srOnly }) => srOnly && css``;
```

src/components/AmsterdamLogo/index.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { LogoShort } from '@datapunt/asc-assets';
+import styled, { css } from 'styled-components';
+import { breakpoint, themeSpacing, themeColor } from '@datapunt/asc-ui';
+
+export const srOnlyStyle = () => ({ srOnly }) =>
```
janjaap, line 6: What is `srOnly` and why should there be a check for it when the prop `srOnly` is always true?

src/components/AmsterdamLogo/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const LogoTitleStyle = styled.span`
+  ${srOnlyStyle()}
+`;
+
+const AmsterdamLogoStyle = styled.a`
+  display: inline-block;
+  height: 30px;
+  width: 68px;
+  flex-shrink: 0;
```
janjaap, line 34: `flex-shrink` Doesn't have any effect when the element isn't a flex element. In this case it's `inline-block`, not `flex` or `inline-flex`.

internals/webpack/webpack.base.babel.js
```diff
@@ -138,6 +139,10 @@ module.exports = options => ({
     modules: ['node_modules', 'src'],
     extensions: ['.js', '.jsx', '.react.js'],
     mainFields: ['browser', 'jsnext:main', 'main'],
+    alias: {
+      '@datapunt/asc-ui': path.resolve(__rootdir, 'node_modules/@datapunt/asc-ui/lib'),
+      '@datapunt/asc-assets': path.resolve(__rootdir, 'node_modules/@datapunt/asc-assets/lib'),
+    },
```
janjaap, line 145: Why wouldn't we make use of esModules? When I apply them, I'm getting a reduction in production bundle size. Not significant, but still a reduction.

src/components/AmsterdamLogo/index.js
aditudorache: indeed

src/components/AmsterdamLogo/index.js
aditudorache: The screen readers variant of the logo. I removed that.

src/components/AmsterdamLogo/index.js
aditudorache: correct, I missed that,removed.

internals/webpack/webpack.base.babel.js
aditudorache: I didn't know we don't support IE11 anymore.

PR #927, opened 2020-06-08
[SIG-2670] Cypress tests wreck

arjandata:


cypress/integration/createSignalWrak.spec.js
```diff
@@ -0,0 +1,144 @@
...
+      );
+      createSignal.setDateTime('Nu');
+
+      cy.contains('Volgende').click();
+    });
+
+    it('Should enter specific information', () => {
+      createSignal.checkSpecificInformationPage();
+
+      cy.contains('Voor het Velodrome ligt een fietswrak, het lijkt een beetje op die van Tom Dumoulin.').should(
```
jpoppe, line 30: Is it on purpose you use the same string, or  would it be nicer to use a variable? I see the `Voor het Velodtrome` string a couple of times.

PR #937, opened 2020-06-08
[SIG-2754] Add configuration for showing the extra questionary

aditudorache:
This PR
- adds a new property in the application configuration: `hasExtraProps`
- when this property si set to false, no extra questionary wil be presented to the user.
- this property will be used to switch off the extra questionaries for the Wesp implementation

Note that the specific configuration of the build repositories for Weesp and Amsterdam wil be done after this PR is Approved

### DIFF
--- a/environment.conf.json
+++ b/environment.conf.json
@@ -54,6 +54,7 @@
       "gemeentenaam": "amsterdam"
     }
   },
+  "hasExtraProps": true,
   "ROOT": "http://localhost:3001/",
   "AUTH_ROOT": "https://acc.api.data.amsterdam.nl/",
   "API_ROOT_MAPSERVER": "https://map.data.amsterdam.nl/",
--- a/environment.conf.json
+++ b/environment.conf.json
@@ -54,7 +54,7 @@
       "gemeentenaam": "amsterdam"
     }
   },
-  "hasExtraProps": true,
+  "showVulaanControls": true,
   "ROOT": "http://localhost:3001/",
   "AUTH_ROOT": "https://acc.api.data.amsterdam.nl/",
   "API_ROOT_MAPSERVER": "https://map.data.amsterdam.nl/",
--- a/src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
+++ b/src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
@@ -1,0 +1,19 @@
+import configuration from 'shared/services/configuration/configuration';
+import vullaan from '../wizard-step-2-vulaan';
+
+describe('wizard-step-2-vullaan', () => {
+  it('should return controls when hasExtraProps is true', () => {
+    configuration.hasExtraProps = true;
+    expect(vullaan.formFactory({ category: "afval" }).controls).not.toEqual({ });
+  });
+
+  it('should return empty controls when hasExtraProps is false', () => {
+    configuration.hasExtraProps = false;
+    expect(vullaan.formFactory({ category: "afval" }).controls).toEqual({ });
+  });
+
+  it('should return empty controls when hasExtraProps is true but no category is found', () => {
+    configuration.hasExtraProps = false;
+    expect(vullaan.formFactory({ category: "not-existing-category" }).controls).toEqual({ });
+  });
+});
--- a/src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
+++ b/src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
@@ -1,19 +1,19 @@
 import configuration from 'shared/services/configuration/configuration';
-import vullaan from '../wizard-step-2-vulaan';
+import vulaan from '../wizard-step-2-vulaan';
 
 describe('wizard-step-2-vullaan', () => {
-  it('should return controls when hasExtraProps is true', () => {
-    configuration.hasExtraProps = true;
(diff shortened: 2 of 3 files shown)


environment.conf.json
```diff
@@ -54,6 +54,7 @@
       "gemeentenaam": "amsterdam"
     }
   },
+  "hasExtraProps": true,
```
janjaap, line 57: I think the name of this prop is too generic. Should be something like `showVulaanControls`. At least use naming that has already been used in the application.

src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
```diff
@@ -0,0 +1,19 @@
+import configuration from 'shared/services/configuration/configuration';
+import vullaan from '../wizard-step-2-vulaan';
+
+describe('wizard-step-2-vullaan', () => {
```
janjaap, line 4: One 'l' too many

src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
```diff
@@ -0,0 +1,19 @@
+import configuration from 'shared/services/configuration/configuration';
+import vullaan from '../wizard-step-2-vulaan';
```
janjaap, line 2: One 'l' too many

src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+describe('wizard-step-2-vullaan', () => {
+  it('should return controls when hasExtraProps is true', () => {
+    configuration.hasExtraProps = true;
+    expect(vullaan.formFactory({ category: "afval" }).controls).not.toEqual({ });
```
janjaap, line 7: Should be no space in `{ }`

src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
```diff
@@ -0,0 +1,19 @@
...
+  });
+
+  it('should return empty controls when hasExtraProps is false', () => {
+    configuration.hasExtraProps = false;
+    expect(vullaan.formFactory({ category: "afval" }).controls).toEqual({ });
```
janjaap, line 12: Should be no space in `{ }`

src/signals/incident/definitions/__tests__/wizard-step-2-vulaan.test.js
```diff
@@ -0,0 +1,19 @@
...
+  });
+
+  it('should return empty controls when hasExtraProps is true but no category is found', () => {
+    configuration.hasExtraProps = false;
+    expect(vullaan.formFactory({ category: "not-existing-category" }).controls).toEqual({ });
```
janjaap, line 17: Should be no space in `{ }`

src/signals/incident/definitions/wizard-step-2-vulaan.js
```diff
@@ -15,6 +16,9 @@ export default {
   previousButtonClass: 'action startagain',
   formAction: 'UPDATE_INCIDENT',
   formFactory: ({ category }) => {
+    const noExtraProps = { controls: {} };
+    if(!configuration.hasExtraProps) return noExtraProps;
```
janjaap, line 20: Missing space after `if`

environment.conf.json
aditudorache: shure thing!

src/signals/incident/definitions/wizard-step-2-vulaan.js
```diff
@@ -17,7 +17,7 @@ export default {
   formAction: 'UPDATE_INCIDENT',
   formFactory: ({ category }) => {
     const noExtraProps = { controls: {} };
-    if(!configuration.hasExtraProps) return noExtraProps;
+    if (!configuration.showVulaanControls) return noExtraProps;
```
janjaap, line 20: The configuration object with throw an exception when a prop isn't there. Might be better to wrap this in a try/catch

aditudorache: Added extra check but be aware that this extra check is not don for any of the configuration object properties...

janjaap: You're right. In the case of querying for props of the `configuration` object, we can actually do two things:

__try/catch__
```jsx
let nestedProp;
try {
  nestedProp = configuration.map.options.nestedProp;
} catch (e) {
 //
}
```

__conditional chaining__
```
const nestedProp = configuration?.map?.options?.nestedProp;
```

In your case, conditional chaining would be fine.

bartjkdp:
This change breaks our deployments that do not contain the `showVulAan` configuration:

image

janjaap:
@bartjkdp In what way? Not having that prop in the configuration should gracefully degrade because of the conditional chaining in `wizard-step-2-vulaan`

bartjkdp:
@janjaap not having this key in the configuration throws an exception that is not caught. I created an issue for this 👍

janjaap:
@bartjkdp Absolutely right; can't believe we missed that. PR: https://github.com/Amsterdam/signals-frontend/pull/959

bartjkdp:
@janjaap no problem :) thanks fix fixing this so quickly!

PR #949, opened 2020-06-10
(Fix) import/no-unresolved

janjaap:
This PR fixes an issue where VS Code wasn't able to resolve module located in the `src` folder. As it turned out, the lint configuration that is used by VS Code needs a `node` setting where running the linter from the command line does not. The latter requires the Webpack config to be able to resolve module properly.

### DIFF
--- a/.eslintrc.cli.js
+++ b/.eslintrc.cli.js
@@ -1,0 +1,10 @@
+module.exports = {
+  extends: '.eslintrc.js',
+  settings: {
+    'import/resolver': {
+      webpack: {
+        config: './internals/webpack/webpack.dev.babel.js',
+      },
+    },
+  },
+};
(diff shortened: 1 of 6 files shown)


.eslintrc.cli.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
aditudorache, line 1: I think we can merge the 2 `.eslintrc.*.js` files by adding this: 
```json
 settings: {
    'import/resolver': {
      webpack: {
        config: './internals/webpack/webpack.dev.babel.js',
      },
      node: {
        extensions: ['.js', '.jsx', '.react.js'],
        moduleDirectory: ['node_modules', './src'],
      },	      
    },
```

it's cleaner and easyer to read

janjaap: We can't, because the CLI only needs the Webpack config and VS Code only needs the `node` config. Putting in both in the same file will still give `import/no-unresolved` warnings in VS Code.

PR #954, opened 2020-06-11
resets the search query when a filter is applied

aditudorache:
This PR fixes the issue that when a filter is applied, the search term was not reseted in the search bar.

### DIFF
--- a/src/containers/App/reducer.js
+++ b/src/containers/App/reducer.js
@@ -2,6 +2,7 @@ import { fromJS } from 'immutable';
 
 import { VARIANT_DEFAULT, TYPE_DEFAULT } from 'containers/Notification/constants';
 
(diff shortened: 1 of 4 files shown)


src/containers/App/reducer.js
```diff
@@ -82,6 +83,9 @@ function appReducer(state = initialState, action) {
     case RESET_SEARCH_QUERY:
       return state.set('searchQuery', initialState.get('searchQuery'));
 
+    case APPLY_FILTER:
+      return state.set('searchQuery', initialState.get('searchQuery'));
+
```
janjaap, line 88: Combine the two cases?

aditudorache: Sure!

PR #958, opened 2020-06-15
(Chore) update packages 

aditudorache:
This PR updates the packages to the latest version as preparation of update of the react-datepicker package.

Note that after the updates, there are 3 linting warnings remaining as result of the update of the `prettier` package to version `^2.0.`. These will be corrected in a future PR.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sia",
-  "version": "1.0.0",
+  "version": "1.16.0",
```
janjaap, line 3: Change isn't really necessary since we're not publishing this package

src/shared/services/auth/services/query-string-parser/query-string-parser.js
```diff
@@ -10,20 +10,20 @@
...
+      keyValue.shift();
+      const value = keyValue.join('=');
+      return {
+        ...params,
+        [decodeURIComponent(key)]: decodeURIComponent(value),
+      };
+    }, {})
+  : null;
+
+export default queryStringParser;
```
janjaap, line 29: Why is this change necessary?

src/shared/services/auth/services/state-token-generator/state-token-generator.js
```diff
@@ -9,10 +9,10 @@
  * @returns {string} 16 random Ascii characters, empty in case the
  * `crypto` library is not available.
  */
-export default function () {
+const stateTokenGenerator = () => {
```
janjaap, line 12: Same here; why is this required? Looks like there is a change in the rules in the linter configuration; I would tackle the issue there instead of changing the code.

package.json
aditudorache: The version was not reflecting the release number which is confusint.

src/shared/services/auth/services/query-string-parser/query-string-parser.js
aditudorache: yes, the default settings of prettier are a bit changed

src/shared/services/auth/services/state-token-generator/state-token-generator.js
aditudorache: yes, the default settings of prettier are a bit changed

src/signals/incident-management/containers/IncidentDetail/components/ChildIncidents/index.js
```diff
@@ -12,9 +12,10 @@ const Title = styled(Heading)`
 `;
 
 const ChildIncidents = ({ incident }) => {
+  const incidentLinks = incident?._links;
   const children = useMemo(
     () =>
-      incident?._links?.['sia:children']?.map(({ href }) => {
+      incidentLinks?.['sia:children']?.map(({ href }) => {
```
janjaap, line 18: What's wrong with the conditional chaining in this case?

aditudorache: It looks like prettier, doesn't accept this (2 nullables) when the incident is in the depentency array.

janjaap: I would adjust the linter config so that the code stays the same. What do you think @jpoppe ?

jpoppe: @aditudorache I agree with @JanJaap if we will start using GraphQL, conditional chaining of multiple levels is a quite common thing.

PR #960, opened 2020-06-15
Make all meta data configurable

boris-arkenaar:
Allow the necessary metadata in `index.html` to be configurable through the `environment.conf.json`.

- `WebpackPwaManifest` has been removed
- `manifest.json` will be injected with environment variables upon launch of the container, like `index.html`

closes Signalen/frontend#28


environment.conf.json
```diff
@@ -1,4 +1,12 @@
 {
+  "meta": {
+    "androidIcon": "/icon_192x192.png",
+    "backgroundColor": "#ffffff",
+    "favicon": "/favicon.png",
+    "iosIcon": "/icon_180x180.png",
+    "statusBarStyle": "default",
+    "themeColor": "#ec0000"
+  },
```
janjaap, line 9: Since this is all manifest data, I would consider renaming `meta` to `manifest`.

environment.conf.json
```diff
@@ -14,6 +22,8 @@
     "siteId": 14
   },
   "language": {
+    "title": "Signalen Informatievoorziening Amsterdam",
+    "shortTitle": "SIA",
```
janjaap, line 26: Shouldn't this prop be a child of the `meta`/`manifest` prop?

src/manifest.json
```diff
@@ -1,15 +1,21 @@
...
+      "src": "$SIGNALS_ANDROID_ICON",
+      "sizes": "192x192",
       "type": "image/png"
     }
   ],
-  "splash_pages": null
+  "name": "$SIGNALS_PWA_TITLE",
+  "orientation": "portrait",
+  "short_name": "$SIGNALS_PWA_SHORT_TITLE",
+  "start_url": ".",
```
janjaap, line 19: Shouldn't the start URL be `/`, instead of `.`, which could be any page?

src/manifest.json
```diff
@@ -1,15 +1,21 @@
...
   "icons": [
     {
-      "src": "favicon.png",
+      "src": "$SIGNALS_IOS_ICON",
+      "sizes": "180x180",
+      "type": "image/png"
+    },
+    {
+      "src": "$SIGNALS_ANDROID_ICON",
+      "sizes": "192x192",
```
janjaap, line 12: The icon seems skewed. I would suggest to get the icons from https://mijn.amsterdam.nl/favicons/android-chrome-192x192.png and https://mijn.amsterdam.nl/favicons/android-chrome-256x256.png; those have the correct dimensions.

environment.conf.json
boris-arkenaar: It is not, though. `favicon` goes into a `meta` tag and so does `iosIcon`.

environment.conf.json
boris-arkenaar: I've considered it, but thought it would be more convenient to have all language together.

environment.conf.json
janjaap: Okay, not all props are meta tag values, but some are and some are not. So, still, the `meta` parent prop doesn't seem like a good container name.

boris-arkenaar: That's right. Meta, as in the general meaning of the word, would be fitting, but since the `meta` tag in the head is called the same it's confusing. I'll see if I can think of another name.

boris-arkenaar: How about `head`, as in the `head` tag? Or `htmlHead`?

PR #961, opened 2020-06-16
[SIG-2829] Optional label

janjaap:
This PR fixes an issue in the incident form where the 'optional' indicator would not be shown. Existing tests for the `Header` component have been replaced to remove snapshot tests.


src/signals/incident/components/form/Header/index.test.js
```diff
@@ -1,93 +1,122 @@
...
+    expect(queryByText(`U kunt maximaal ${requiredLength} tekens invoeren.`)).not.toBeInTheDocument();
 
-      expect(wrapper).toMatchSnapshot();
-    });
+    rerender(withAppContext(<Header hasError={hasError} getError={getError} touched />));
 
-    it('should render email error', () => {
-      hasError.mockImplementation(type => type === 'email');
-      wrapper.setProps({ touched: true });
+    expect(queryByText(`U kunt maximaal ${requiredLength} tekens invoeren.`)).toBeInTheDocument();
```
boris-arkenaar, line 97: You could define the error message string in a `const` before, so you're absolutely sure you use the same string in both `expect` statements.

src/signals/incident/components/form/Header/index.test.js
```diff
@@ -1,93 +1,122 @@
...
+    const hasError = prop => prop === 'required';
+    const touched = false;
 
-    it('should render optional field message', () => {
-      wrapper.setProps({ options: { validators: [] } });
+    const { queryByText } = render(withAppContext(<Header hasError={hasError} touched={touched} />));
 
-      expect(wrapper).toMatchSnapshot();
-    });
+    expect(queryByText('Dit is een verplicht veld')).not.toBeInTheDocument();
```
boris-arkenaar, line 120: Where do you test the positive version?

janjaap: In the 'should render required error' test case

src/signals/incident/components/form/Header/index.test.js
janjaap: I thought about that, but it's only two occurrences 🤷‍♂️

boris-arkenaar: I know, but if you make a typo in the negative (`not`) test case, you won't notice it while your test isn't testing properly. So, still valuable, even if it's only two occurrences.

src/signals/incident/components/form/Header/index.test.js
boris-arkenaar: In this case as well, put the string in a `const`, to make sure the negative test tests what it should.

src/signals/incident/components/form/Header/index.test.js
janjaap: Done

PR #963, opened 2020-06-16
[SIG-2758] Modify holiday rentals question text

jpoppe:


src/signals/incident/definitions/wizard-step-2-vulaan/wonen.js
```diff
@@ -634,7 +634,7 @@ const vakantieverhuur = {
       type: 'caution',
       value: [
         'Bel nu met 14 020',
-        'Vraag naar team Vakantieverhuur. U wordt direct doorverbonden met een medewerker. Handhaving gaat, indien mogelijk, meteen langs.',
+        'Vraag naar team Vakantieverhuur. U wordt direct doorverbonden met een medewerker. Handhaving gaat, indien mogelijk, binnen 24 uur langs.',
```
boris-arkenaar, line 637: Ik zou zeggen: "Handhaving *komt* [...] langs".

jpoppe: I agree, but that's not up to me ;) Will forward it.

janjaap:
@jpoppe We should verify if this change should be part of the upcoming release before merging this branch. Because, if so, it should be merged in the release branch instead of in `develop`

PR #974, opened 2020-06-22
[SIG-2072] Update react-datepicker

aditudorache:
This PR 
- updates the reac-datepicker to the last version
- removes the moment package from the project


src/shared/services/date-utils/index.js
```diff
@@ -0,0 +1,14 @@
...
+    `${String(date.getFullYear())}-${String(
+      date.getMonth() + 1
+    ).padStart(2, '0')}-${String(date.getDate()).padStart(2, '0')}`
+);
+
+export const dateToString = date => (
+  date && format(date, 'dd-MM-yyyy', date)
+);
+
+export const capitalizeFirstLetter = value => value && (value[0].toUpperCase() + value.substring(1));
```
janjaap, line 14: Or `capitalize`

src/shared/services/date-utils/index.js
```diff
@@ -0,0 +1,14 @@
+import format from 'date-fns/format';
+
+export const dateToISOString = date => (
+  date &&
+    `${String(date.getFullYear())}-${String(
+      date.getMonth() + 1
+    ).padStart(2, '0')}-${String(date.getDate()).padStart(2, '0')}`
+);
+
+export const dateToString = date => (
```
janjaap, line 10: This function doesn't seem to be covered by a test. Can you add one?

src/shared/services/date-utils/index.js
janjaap: Function also not covered by a test

src/shared/services/string-parser/index.js
```diff
@@ -0,0 +1,16 @@
+import { dateToString } from 'shared/services/date-utils';
+import format from 'date-fns/format';
+
+export const string2date = value => {
+  if (!value) return '';
```
janjaap, line 5: This condition doesn't seem to be tested

src/shared/services/string-parser/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+  const date = new Date(value);
+  return date ? dateToString(date) : `[${value}]`;
+};
+
+export const string2time = value => {
+  if (!value) return '';
```
janjaap, line 12: This condition doesn't seem to be tested

src/app.js
```diff
@@ -27,6 +28,8 @@ import './polyfills';
 
 import configureStore from './configureStore';
 
+setDefaultLocale('nl');
```
janjaap, line 31: Shouldn't this line be in the `CalendarInput` component? That's where it is used.
Also, wouldn't `registerLocale('nl', nl);` be enough to use? The `react-datepicker` docs don't mention having to set both, as far as I can see.

src/signals/incident/components/IncidentPreview/components/DateTime/index.js
```diff
@@ -9,12 +12,12 @@ const getValue = (value, incident) => {
...
   }
 
-  const time = moment(`${incident.incident_time_hours}:${incident.incident_time_minutes}`, 'H:m').format('H:mm');
+  const time = `${incident.incident_time_hours}:${String(incident.incident_time_minutes).padStart(2, '0')}`;
   if (incident.incident_date === 'Vandaag') {
     return `Vandaag, ${time}`;
   }
 
-  return `${moment(incident.incident_date).format('dddd D MMMM')}, ${time}`;
+  return `${capitalizeFirstLetter(format(parseISO(incident.incident_date), 'EEEE d MMMM', { locale: nl }))}, ${time}`;
```
janjaap, line 20: The `capitalizeFirstLetter` function can be handled by CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform#:~:text=The%20text%2Dtransform%20CSS%20property,help%20improve%20legibility%20for%20ruby.

aditudorache: Indeed! I have tried that but in this case the different value types are renderd from an array and it would make the change more complex than this.

src/signals/incident/components/IncidentPreview/components/DateTime/index.js
```diff
@@ -17,7 +17,7 @@ const getValue = (value, incident) => {
     return `Vandaag, ${time}`;
   }
 
-  return `${capitalizeFirstLetter(format(parseISO(incident.incident_date), 'EEEE d MMMM', { locale: nl }))}, ${time}`;
+  return `${capitalize(format(parseISO(incident.incident_date), 'EEEE d MMMM', { locale: nl }))}, ${time}`;
```
janjaap, line 20: Please solve this with CSS instead of with JS

src/signals/incident/components/form/DateTimeInput/index.js
```diff
@@ -16,7 +16,7 @@ const formatDate = (offset, type = 'value') => {
   }
 
   const date = subDays(new Date(), offset);
-  return capitalizeFirstLetter(format(date, dateFormat, { locale: nl }));
+  return capitalize(format(date, dateFormat, { locale: nl }));
```
janjaap, line 19: Same here; solve with CSS

PR #977, opened 2020-06-22
Bug/49 service worker

boris-arkenaar:
# Remove Service Worker

The service worker is giving caching issues in combination with our current setup.

A version name for the service worker is defined during the Webpack build process.
On every deploy of the docker container, with a potentially different configuration, the resulting `index.html` would be different. This difference is not picked up (correctly) by the service worker because the version string is the same for every start of the container based on one docker image.

Since the service worker is not being used and not useful for the Signals application since it doesn't function offline anyway, the easiest solution is to remove the service worker altogether.

Some server configuration files have been removed as well. I believe these are not being used. Please comment in case I'm wrong.

fixes Signalen/frontend#49


janjaap:
@boris-arkenaar The serviceworker implementation is used in signals-frontend. What exactly gave you the idea that it wasn't?
Wouldn't it be better to make sure that `index.html` isn't cached by the serviceworker?

boris-arkenaar:
@janjaap The fact that the application is not very useful when offline made me think the service worker isn't very useful either. Caching can also be done by the browser, without a service worker. What exactly is the service worker being used for?
If we want to keep the service worker I'll see if I can make sure to exclude index.html from caching.

janjaap:
@boris-arkenaar I'm inclined to go with your suggestion to remove the serviceworker implementation. The site benefits from offline storage, but since there's no offline strategy in place (for instance keeping event sources in memory till the network is online again), relying on browser cache is also a safe bet.

Your thoughts @aditudorache, @jpoppe ?

janjaap:
@boris-arkenaar There is definitely a benefit in having the serviceworker in place. We've experience issues in the past where, because of updates on the load balancers, that the deployed front-end was not available, yet the API was. Users were still able to use the website. We'd like to keep the serviceworker.

I think we have two options here:
1. Patch the serviceworker's file list so that caching issues aren't occuring any more or
2. Make the instantiation of the serviceworker configurable so that we can decide for ourselves if we want it or not.

My preference would be the first option, because of the obvious benefits that a serviceworker offers.

bartjkdp:
I would prefer the first option as well. I think the service worker should invalidate the cache of index.html when settings change. Maybe we can use something similar to [this](https://dev.to/flexdinesh/cache-busting-a-react-app-22lk) and compare the hashes of `window.CONFIG` to see if the settings changed.

jpoppe:
I discussed this offline with @janjaap last week the outcome is/was I also vote for option 1 since the service worker saved us 'downtime' in the past.

boris-arkenaar:
A very valid point to keep the service worker! I will look into implementing option one.

boris-arkenaar:
How about the following approach?

Normally the webpack offline plugin creates a version string for the cache during build time. I've moved this to the moment where the docker container will be created. So now, instead of a new version on each new build, there will be a new version on each deploy of the docker image, where you can potentially have a different configuration and thus a different `index.html`.

janjaap:
@boris-arkenaar Why not exclude `index.html` from being cached?
See https://github.com/NekR/offline-plugin/blob/master/docs/options.md#excludes-arraystring--globs_pattern

boris-arkenaar:
@janjaap because then the whole offline-mode/frontend-server-not-available-situation becomes useless. Without an `index.html` in your cache, you have no application to run.

PR #983, opened 2020-06-24
[SIG-2853] Fix various FilterForm reset (and other) bugs 

jpoppe:


src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
```diff
@@ -258,7 +258,10 @@ describe('signals/incident-management/components/FilterForm', () => {
 
     expect(submitButton.textContent).toEqual(DEFAULT_SUBMIT_BUTTON_LABEL);
 
-    act(() => { fireEvent.change(nameField, { target: { value: 'My filter' } }); });
+    act(() => {
+      fireEvent.change(nameField, { target: { value: 'My filter' } });
+      fireEvent.blur(nameField);
+    });
```
janjaap: If both `onChange` and `onBlur` events have to stay in place, then please separate their behaviour into different test cases instead of combining them into the same `act` call. It's impossible now to determine which handler updated the state.
That also goes for tests below.

src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
```diff
@@ -282,13 +285,9 @@ describe('signals/incident-management/components/FilterForm', () => {
 
     const addressField = container.querySelector('input[type="text"][name="address_text"]');
 
-    await act(async () => {
-      fireEvent.change(addressField, { target: { value: 'Weesperstraat 113/117' } });
-    });
+    await act(async () => { fireEvent.change(addressField, { target: { value: 'Weesperstraat 113/117' } }); });
```
janjaap, line 288: Does the `async` actually work? The closure doesn't return a value/promise

src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
```diff
@@ -282,13 +285,9 @@ describe('signals/incident-management/components/FilterForm', () => {
...
 
-    await act(async () => {
-      fireEvent.blur(addressField);
-    });
+    await act(async () => { fireEvent.blur(addressField); });
```
janjaap, line 290: Same here.

src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
jpoppe: Yes, without the test will break, this seems to be some `testing-library` magic

src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
jpoppe: Yes, without the test will break, this seems to be some `testing-library` magic

src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
jpoppe: Thanks, we indeed don't need the `onchange fireEvents` anymore :)

PR #984, opened 2020-06-24
[SIG-2873] Fix tests

janjaap:
This PR contains fixes for several tests that used deprecated functions from `@testing-library/react`. All tests have been run with the `--runInBand --detectOpenHandles` options to detect memory leaks. Where necessary, tests have been modified to prevent leaks.

Next to that, a small modification has been added to the linter configuration that prevented formatting files without generating  unnecessary changes.

For a couple of components where `useCallback` is used in conjunction with Lodash' `debounce` function, the linter has been disabled for the `react-hooks/exhaustive-deps` rule.

### DIFF
--- a/src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
+++ b/src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
@@ -43,22 +43,19 @@ const StyledButton = styled(Button)`
 
 const DEFAULT_TEXT_FIELDS = 15;
 
+const fields = [...new Array(DEFAULT_TEXT_FIELDS).keys()].reduce(
+  (acc, key) => ({
+    ...acc,
+    [`item${key}`]: FormBuilder.group({
+      title: [''],
+      text: [''],
+    }),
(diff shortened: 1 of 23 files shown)


src/signals/incident-management/containers/DefaultTextsAdmin/components/DefaultTextsForm/index.js
```diff
@@ -43,22 +43,19 @@ const StyledButton = styled(Button)`
...
-          title: [''],
-          text: [''],
-        }),
-      }),
-      {}
-    );
-
-    return FormBuilder.group(fields);
-  }, []);
+const fields = [...new Array(DEFAULT_TEXT_FIELDS).keys()].reduce(
```
aditudorache, line 46: Why do we need to do this global? Does useMemo make any sense then? How about using an object factory?

janjaap: It's not a global value; it's a var that is limited to the module's scope. `useMemo` Is needed to ensure that the `form` const doesn't change between renders.

PR #985, opened 2020-06-24
Bug/36 map search

boris-arkenaar:
Allow an array of municipalities for the (previously named) `gemeentenaam` param of the `PDOKAutoSuggest`.

## Changes

- The param `gemeentenaam` has been renamed to `municipality` in `PDOKAutoSuggest` and `environment.conf.json`.
- The `municipality` param accepts either a string or an array of municipalities.
- The back office is updated to also use the `municipality` config param.

closes Signalen/frontend#36


.eslintrc.js
```diff
@@ -63,6 +63,7 @@ module.exports = {
     'newline-per-chained-call': 0,
     'no-confusing-arrow': 0,
     'no-console': 1,
+    'no-nested-ternary': 0,
```
janjaap, line 66: This will lead to unreadable code constructs. I'm not in favor of introducing this.
@aditudorach, @jpoppe; your thoughts?

src/components/PDOKAutoSuggest/__tests__/PDOKAutoSuggest.test.js
```diff
@@ -8,13 +8,35 @@ import { formatPDOKResponse } from 'shared/services/map-location';
...
+  const result = render(withAppContext(<PDOKAutoSuggest onSelect={onSelect} {...props} />));
+  const input = result.container.querySelector('input[aria-autocomplete]');
+
+  input.focus();
+
+  act(() => {
+    fireEvent.change(input, { target: { value } });
+  });
+
+  await wait(() => resolveAfterMs(INPUT_DELAY));
```
janjaap, line 28: `wait` Is deprecated in the upcoming version of `testing-library`. It's been replaced by `waitFor`.

src/components/PDOKAutoSuggest/index.js
```diff
@@ -55,10 +55,13 @@ PDOKAutoSuggest.propTypes = {
...
-   * Value that determines to which municipality the search query should be applied
-   * Can be a single name, like amsterdam, or a combination, like (amsterdam OR weesp)
+   * Value that determines to which municipality the search query should be
+   * applied.
+   *
+   * Can be a single name, like amsterdam, or an array of one or more names,
+   * which will be combined using a logical OR
    */
-  gemeentenaam: PropTypes.string,
+  municipality: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
```
janjaap, line 64: `municipality` Isn't a required prop, so it should have a default value.

src/components/PDOKAutoSuggest/index.js
```diff
@@ -55,10 +55,13 @@ PDOKAutoSuggest.propTypes = {
   className: PropTypes.string,
   fieldList: PropTypes.arrayOf(PropTypes.string),
   /**
-   * Value that determines to which municipality the search query should be applied
-   * Can be a single name, like amsterdam, or a combination, like (amsterdam OR weesp)
+   * Value that determines to which municipality the search query should be
+   * applied.
+   *
+   * Can be a single name, like amsterdam, or an array of one or more names,
+   * which will be combined using a logical OR
```
janjaap, line 62: There isn't a construct in the code that point to the use of the logical OR. Is it worth mentioning this?

src/components/PDOKAutoSuggest/index.js
janjaap: On second thought; strike that. `undefined` Is also handled properly without generating warnings.

environment.conf.json
```diff
@@ -101,4 +101,4 @@
   "SEARCH_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/private/search",
   "TERMS_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/private/terms/categories/",
   "USERS_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/private/users/"
-}
+}
```
aditudorache, line 104: Here is the newline missing, is this intended?

.eslintrc.js
aditudorache: I agree with JJ. Le't keep this out.

environment.conf.json
```diff
@@ -67,7 +67,7 @@
       "zoom": 10,
       "maxZoom": 14,
       "minZoom": 8,
-      "gemeentenaam": "amsterdam"
+      "municipality": "amsterdam"
```
janjaap, line 70: I'm documenting the configuration properties and I think that we should take this prop out of the `options` context and should put it directly under `map` instead of `map.options`. The `options` prop should only contain Leaflet Map configuration options (https://leafletjs.com/reference-1.6.0.html#map-factory)

src/components/PDOKAutoSuggest/index.js
boris-arkenaar: To make clear that that's the way multiple municipalities will be logically combined, with OR, not AND. But maybe you're right, wouldn't make sense to combine them with AND, so unnecessary to mention it.

environment.conf.json
boris-arkenaar: This was VSCode's intention, not sure why. Some linting rules have changed?

.eslintrc.js
boris-arkenaar: It added it because of this line: https://github.com/Amsterdam/signals-frontend/pull/985/files#diff-a1eeb871238ab0872eae54894bc0d3e4R26
I see your point, but it hurts me if I would have to write that differently.

environment.conf.json
boris-arkenaar: Fixed it.

src/components/PDOKAutoSuggest/__tests__/PDOKAutoSuggest.test.js
boris-arkenaar: Replaced it. Had to update `@testing-library/react`. Removed `@testing-library/dom` since it's a dependency of `@testing-library/react` and not used anywhere directly.

.eslintrc.js
jpoppe: I also **strongly** agree we should avoid nested ternary expressions (JOKE: we are writing Javascript and not Perl :laughing: )

boris-arkenaar: OK. I will take it out!

boris-arkenaar:
Ready for a final review @janjaap @aditudorache @jpoppe

janjaap:
Both https://github.com/Amsterdam/signals-weesp/pull/20 and https://github.com/Amsterdam/signals-amsterdam/pull/18 need to be merged before this PR can be merged.

PR #986, opened 2020-06-24
[SIG-2854] Set logo max-height CSS value

jpoppe:


src/components/Logo/__tests__/Logo.test.js
```diff
@@ -22,30 +24,30 @@ describe('components/Logo', () => {
...
 
   it('should render correctly', () => {
-    const { container } = render(withAppContext(<Logo />));
+    const { container, getByTestId } = render(withAppContext(<Logo tall />));
 
     expect(container.querySelector(`a[href="${configuration.links.home}"]`)).toBeInTheDocument();
     expect(container.querySelector(`img[src="${configuration.logo.url}"]`)).toBeInTheDocument();
-    expect(container.querySelector(`img[height="${configuration.logo.height}"]`)).toBeInTheDocument();
-    expect(container.querySelector(`img[width="${configuration.logo.width}"]`)).toBeInTheDocument();
+    expect(getByTestId('logo')).toHaveStyleRule('height', configuration.logo.height.toString());
```
janjaap, line 31: This assertion will check for style rule `height: 80`, instead of `height: 80px`. Is that indended?

src/components/Logo/__tests__/Logo.test.js
```diff
@@ -22,30 +24,30 @@ describe('components/Logo', () => {
...
-    const { container } = render(withAppContext(<Logo tall />));
+    const { container, getByTestId } = render(withAppContext(<Logo tall />));
 
     expect(container.querySelector(`a[href="${configuration.links.home}"]`)).toBeInTheDocument();
     expect(container.querySelector(`img[src="${configuration.logo.url}"]`)).toBeInTheDocument();
-    expect(container.querySelector(`img[height="${configuration.logo.height}"]`)).toBeInTheDocument();
-    expect(container.querySelector(`img[width="${configuration.logo.width}"]`)).toBeInTheDocument();
+    expect(getByTestId('logo')).toHaveStyleRule('height', configuration.logo.height.toString());
+    expect(getByTestId('logo')).toHaveStyleRule('width', configuration.logo.width.toString());
   });
```
janjaap, line 42: This test case is equal to the previous one. Maybe remove it?

src/components/Logo/__tests__/Logo.test.js
```diff
@@ -22,30 +24,30 @@ describe('components/Logo', () => {
...
 
   it('should render correctly', () => {
-    const { container } = render(withAppContext(<Logo />));
+    const { container, getByTestId } = render(withAppContext(<Logo tall />));
```
janjaap, line 27: `tall` Defaults to 'true'. No need to provide it.

src/components/Logo/index.js
```diff
@@ -1,16 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+
+import styled from 'styled-components';
+
 import configuration from 'shared/services/configuration/configuration';
 
+const StyledLogo = styled.img`
+  width: ${props => props.tall === true ? configuration.logo.width : configuration.logo.smallWidth};
```
janjaap, line 9: Or 
```jsx
width: ${({ tall }) => tall ? configuration.logo.width : configuration.logo.smallWidth};
```

janjaap: I don't think so. You could potentially set `100%` as value in the configuration.

src/components/Logo/__tests__/Logo.test.js
jpoppe: IMHO this is more explicit, but will remove.

src/components/Logo/__tests__/Logo.test.js
jpoppe: This is how the test was before, in the end we are only checking if the `configuration` value is set, I agree 80px is nicer, will change.

src/components/Logo/__tests__/Logo.test.js
jpoppe: Same here, it was there, will remove.

src/components/Logo/index.js
boris-arkenaar: Of course, I see. Than we need to make sure to add that to the configuration file.

jpoppe: Will check if `jest-styled-components` will understand this and adjust it if so.

jpoppe: It does :)

src/components/Logo/index.js
```diff
@@ -6,9 +6,9 @@ import styled from 'styled-components';
 import configuration from 'shared/services/configuration/configuration';
 
 const StyledLogo = styled.img`
-  width: ${props => props.tall === true ? configuration.logo.width : configuration.logo.smallWidth};
-  height: ${props => props.tall === true ? configuration.logo.height : configuration.logo.smallHeight};
-  max-height: ${props => props.tall === true ? '56px' : '32px'};
+  width: ${({ tall }) => tall ? configuration.logo.width : configuration.logo.smallWidth};
+  height: ${({ tall }) => tall ? configuration.logo.height : configuration.logo.smallHeight};
+  max-height: ${({ tall }) => tall === true ? '56px' : '32px'};
```
janjaap, line 11:
```
max-height: ${({ tall }) => tall ? '56px' : '32px'};
```

src/components/Logo/index.js
boris-arkenaar: Do we want this flexibility though? It was pixels up till now and that is enough I would say.
Otherwise you move css syntax to the configuration.

jpoppe: @boris-arkenaar I don't really understand what you mean, before my changes there was no `px` value hard coded in the template strings, this is the same after this changes, since this value is a string we could use pixels or percentage.

Or I missed something ;)

boris-arkenaar: Before the value was used in the `width` and `height` attributes of the `img` tag which assumes pixel values by default. Now you use them in css attributes which mean you would need to add `px`.

So if you add `px` to the code, no need to change the config.

If you don't add `px` to the code, the configuration needs to be updated. The configuration becomes more powerful, but does it need to?

jpoppe: @boris-arkenaar check, I missed something! I did not realize (or forgot) that the `pixels` were set automatically on a `HTML attribute`, whoops!, thanks for clarifying!

That said, we are using percentages for those values so would :heart: to keep the **flexibility**, hope you can live with that.

src/components/Logo/index.js
jpoppe: Removed explicit **true**, will add linting rules for this soon®

src/components/Logo/index.js
janjaap: I agree with @jpoppe on this one; we indeed don't want our CSS rules to be reflected by the application's configuration, but having [length](https://developer.mozilla.org/en-US/docs/Web/CSS/length) values instead of numerical values greatly increases the flexibility for the positioning of the application's logo.

src/components/Logo/__tests__/Logo.test.js
janjaap: Did you?

PR #1001, opened 2020-07-01
[SIG-2900] LocationMarker component

janjaap:
This PR adds the `LocationMarker` component. This component doesn't actually render anything, because it taps directly into the Leaflet API to add vector layers to the map instance.

I tried using both `Circle` and `CircleMarker` from [Amsterdam/react-maps](https://github.com/Amsterdam/amsterdam-react-maps), but I ran into issues with the zoom buttons not being clickable or the `Circle` vector scaling unpredictably when zooming in or out.

### DIFF
--- a/src/components/LocationMarker/__tests__/LocationMarker.test.js
+++ b/src/components/LocationMarker/__tests__/LocationMarker.test.js
@@ -1,4 +1,4 @@
-import L from 'leaflet';
+import Leaflet from 'leaflet';
 import React from 'react';
 import { render } from '@testing-library/react';
 import { withAppContext } from 'test/utils';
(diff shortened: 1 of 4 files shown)


src/components/LocationMarker/__tests__/LocationMarker.test.js
```diff
@@ -0,0 +1,63 @@
+import L from 'leaflet';
```
jpoppe, line 1: Any reason to use a one letter variable name here?

src/components/LocationMarker/index.js
```diff
@@ -0,0 +1,60 @@
+import { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import L from 'leaflet';
```
jpoppe, line 3: Same here

src/components/LocationMarker/__tests__/LocationMarker.test.js
janjaap: That's how it's been done so far where Leaflet was included. I didn't think twice about it. I can change it, if you want me to.

janjaap:
Import renamed

PR #1002, opened 2020-07-01
Remove Amsterdam references

boris-arkenaar:
Removed some hard-coded Amsterdam specific references.

Updated the config. See the changes to the readme on which entries have been added.

closes Signalen/frontend#54
closes Signalen/frontend#55
closes Signalen/frontend#58
closes Signalen/frontend#59


environment.conf.json
```diff
@@ -25,10 +25,21 @@
...
-    "footer2": "op werkdagen van 08.00 tot 18.00 uur."
+    "footer2": "op werkdagen van 08.00 tot 18.00 uur.",
+    "incidentForm": {
+      "step5": {
+        "consent": "Ja, ik geef de gemeente Amsterdam toestemming om mijn contactgegevens te delen met andere organisaties, als dat nodig is om mijn melding goed op te lossen."
+      },
+      "step7": {
+        "extraInfo": "Voor spoedeisende zaken kunt u ook telefonisch contact opnemen met 14 020."
+      }
+    }
```
janjaap, line 39: I'm not sure about the explicit `step5` and `step7` prop names. What if there is steps added in between or the order of steps is changed? Then the numbering doesn't make sense any more. I think we're better off choosing proper prop names instead of trying to nest it in a clever way. I propose using `allowConsent` and `urgentContactInfo` in these cases.

That having said, these two Amsterdam specific texts aren't the only two that refer to the municipality of Amsterdam. There are texts in the definitions of `overlast-van-dieren`, `overlast-op-het-water`, `overlast-van-en-door-personen-of-groepen`, `wegen-verkeer-straatmeubilair` and `wizard-step-7-fout` that can only be applied to Amsterdam.

src/components/SiteHeader/index.js
```diff
@@ -247,11 +247,7 @@ const MenuItems = ({ onLogOut, showItems }) => {
       {showLogout && (
         <Fragment>
           <MenuItem>
-            <StyledMenuButton
-              forwardedAs="a"
-              href="https://tamtam.amsterdam.nl/do/office?id=1723860-6f6666696365"
-              target="_blank"
-            >
+            <StyledMenuButton forwardedAs="a" href={configuration.links?.help} target="_blank">
```
janjaap, line 250: The conditional chaining implies that the `help` prop might not be present. In that case, `StyledMenuButton` shouldn't be rendered at all.

environment.conf.json
boris-arkenaar: I agree that these explicit references to the steps aren't very pretty. I'll update those, but keep them nested in `incidentForm`. Or do you prefer not to have any nesting at all?

All the other references you mention are the additional questions and those are being moved to the backend anyway. The only one left is the one you mention in step-7, but I moved that one to the configuration as well already, as you can see.

src/components/SiteHeader/index.js
boris-arkenaar: That'll be more correct indeed.

environment.conf.json
boris-arkenaar: With the names you came up with, I'd say it doesn't need to be nested at all actually!

I'd use a different name for `allowConsent` though. Like this, it seems like a pleonasm. A better description would be `consentToContactSharing`.

boris-arkenaar:
@janjaap Made changes based on your feedback.

boris-arkenaar:
@aditudorache Do you have any experience by any change with the seemingly unrelated unit test that fails in this PR?

    FAIL src/components/FormFooter/__tests__/FormFooter.test.js
      ? src/components/FormFooter should render correctly

       No style rules found on passed Component

PR #1007, opened 2020-07-03
[SIG-2905] Footer text

janjaap:
This PR adds a section to the site's footer with a notification on privacy. In order to get that working, this PR:
- adds texts to the configuration file
- adds the general phone number to the configuration file
- replaces occurrences of the phone number with a reference to the configuration file
- export a function (`stringFormatter`) that can replace occurrences of multiple string in a source string
- refactors the `Footer` component
- makes the phone number in the footer clickable (and thus a callable link on mobile devices)
- moves some of the CSS from `global.scss` to the `App` container


boris-arkenaar:
We should consider some i18n package. I don't like the
`dangerouslySetInnerHTML` and the text replacements.

janjaap:
@boris-arkenaar I was thinking about that too. Do you have any suggestions that are low impact/footprint?

boris-arkenaar:
Not necessarily the lowest footprint, but a good one that is going to pay off in the long run: https://www.i18next.com/

janjaap:
@boris-arkenaar Why would you say it would pay off in the long run? By which criteria?

boris-arkenaar:
Because it seems to be well designed with good attention to all the
linguistical needs necessary. Especially when we would serve the application in
more than one language this will be a huge benefit. For developers and for
translators, and thus, ultimately the users as well!

janjaap:
I don't think there is any need to serve a multilingual application; we only need to be able to provide texts and perform interpolation on them. Maybe have plurals, but probably not even that.
I'm going to merge this PR and create a ticket on our board so that we can look into including a package that does that.

PR #1011, opened 2020-07-06
Remove unnecessary use of css template literal tag

boris-arkenaar:
## SIA

- [ ] double-check your branch is based on `develop` and targets `develop`
- [ ] Pull request has tests (we are going for 100% coverage!)
- [ ] Code is well-commented, linted and follows project conventions
- [ ] Internal code generators and templates are updated (if necessary)

### DIFF
--- a/src/components/FormAlert/index.js
+++ b/src/components/FormAlert/index.js
@@ -1,20 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled, { css } from 'styled-components';
-import {
-  themeColor,
-  Heading,
-  ascDefaultTheme,
-  themeSpacing,
-} from '@datapunt/asc-ui';
+import styled from 'styled-components';
+import { themeColor, Heading, ascDefaultTheme, themeSpacing } from '@datapunt/asc-ui';
 
 const Div = styled.div`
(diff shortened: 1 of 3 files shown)


src/components/FormAlert/index.js
```diff
@@ -27,12 +21,11 @@ const Title = styled(Heading).attrs({
   forwardedAs: 'h4',
 })`
   font-family: Avenir Next LT W01 Demi, arial, sans-serif;
-  color: ${({ isNotification }) =>
-    isNotification ? themeColor('support', 'valid') : themeColor('secondary')};
+  color: ${({ isNotification }) => (isNotification ? themeColor('support', 'valid') : themeColor('secondary'))};
 
   ${({ hasMessage }) =>
     !hasMessage &&
-    css`
```
janjaap, line 35: Why is it necessary to remove the `css` function call? The [documentation still mentions it is required](https://styled-components.com/docs/api#css).

boris-arkenaar: It seems to fix the failing test.
The `css` template tag is not necessary, unless you interpolate a function.

> If you're interpolating a string you do not need to use this, only if you're
> interpolating a function.

I've added the change to #1002

PR #1012, opened 2020-07-06
(Chore) update reactive form

aditudorache:
This PR 
- updates the react-reactive-form, few other devDependency packages.
- the asc packages are not upgraded due to a bug in the select component


src/signals/incident/components/IncidentNavigation/index.js
```diff
@@ -16,41 +16,44 @@ const IncidentNavigation = ({ controls, meta: { wizard, submitting, handleSubmit
...
+                    )}
+                  </NextButton>
+                )}
+              </div>
+            )
+          );
+        }}
+      />
+    )) ||
+    null
```
janjaap, line 56: Returning `undefined` also does the trick, right? Shouldn't be needed to explicitly return `null`

src/signals/incident/components/IncidentNavigation/index.js
```diff
@@ -16,41 +16,44 @@ const IncidentNavigation = ({ controls, meta: { wizard, submitting, handleSubmit
...
 
   return (
-    <WithWizard
-      render={({ next, previous, step }) => {
-        const currentStep = (step && step.id && step.id.split('/').pop()) || 0;
-        const wizardStep = currentStep && wizard[currentStep];
+    (wizard && (
+      <WithWizard
+        render={({ next, previous, step }) => {
+          const currentStep = (step && step.id && step.id.split('/').pop()) || 0;
```
janjaap, line 22: While you're at it, you could also change this line in
```jsx
const currentStep = (step?.id?.split('/').pop()) || 0;
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sia",
-  "version": "1.16.0",
+  "version": "1.20.0",
```
janjaap, line 3: We're currently at version `1.19.0`. Why do we need to update this number? We're not publishing this application as a private or public NPM package, so the version number has no meaning.

src/signals/incident-management/containers/DefaultTextsAdmin/components/SelectForm/index.test.js
```diff
@@ -30,14 +30,14 @@ describe('<SelectForm />', () => {
...
   it('should render form correctly', () => {
-    const { queryByTestId, queryByText, queryByDisplayValue } = render(
+    const { queryByTestId, queryByText, queryAllByText, queryByDisplayValue } = render(
       withAppContext(<SelectForm {...props} />)
     );
 
     expect(queryByTestId('selectFormForm')).not.toBeNull();
 
     expect(queryByText('Subcategorie')).not.toBeNull();
-    expect(queryByText('Asbest / accu')).not.toBeNull();
```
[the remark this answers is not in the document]
aditudorache: The last verision of select from asc has 2 times the selected version, one untrimmed which is hidden and one visible that can contain the ellipsis

src/signals/incident/components/IncidentNavigation/index.js
aditudorache: Yes!

src/signals/incident/components/IncidentNavigation/index.js
aditudorache: Not really. If I do not return null, some unittests are failing. I thought a component should render something or return null, right?

`src/signals/incident/components/IncidentForm/index.test.js:68:42`
```
<IncidentForm /> › rendering › expect to render correctly

    Field(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

src/shared/types/index.js
```diff
@@ -148,7 +148,7 @@ export const incidentType = PropTypes.shape({
     priority: PropTypes.string,
   }),
   signal_id: PropTypes.string,
-  source: PropTypes.string,
+  source: PropTypes.oneOfType([PropTypes.string, PropTypes.shape({})]),
```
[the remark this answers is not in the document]
aditudorache: It is actualy an object (of type File), I changed it to that

src/global.scss
```diff
@@ -184,10 +184,6 @@ fieldset {
   padding-top: 50px;
 }
 
-[aria-hidden="true"] {
-  display: none !important;
-}
-
```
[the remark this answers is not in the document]
janjaap: The `Heade` component from `asc-ui` renders elements that have the `aria-hidden` attribute set. That is why that rule is in place.

src/signals/incident/components/IncidentNavigation/index.js
janjaap: Ah, the illustrious incident form. Fair enough.

package.json
janjaap: We don't need the version number in `package.json` to be up to date; we're not publishing the package nor will we ever.

src/shared/types/index.js
```diff
@@ -148,7 +148,7 @@ export const incidentType = PropTypes.shape({
     priority: PropTypes.string,
   }),
   signal_id: PropTypes.string,
-  source: PropTypes.oneOfType([PropTypes.string, PropTypes.shape({})]),
+  source: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
janjaap, line 151: `PropTypes.object` Shouldn't be used; use either `objectOf` or `shape`.

aditudorache: My fault, is actually a shape({id, label})

aditudorache:
Closing this. Was addressed in #1129

PR #1020, opened 2020-07-08
[SIG-2886] New eslint rules part 2

jpoppe:
## Future steps (Not to self)

* Move ESlint stuff to it's own plugin/package (will do when mono repo is 'officially' in place), categorize rules 
* Remove Prettier and AirBNB plugins, those two are conflicting by default and most rules are already superseded, there are nice tools to extract the eventually missing rules
* Proposed rules should be picked up one day, really easy to check rule by rule with **eslint-nibble** (will add separate script)
* Change related React files to jsx (easy fix but needs to be planned), so we can fine tune rules even better
* Add and configure even more rules :sunglasses: 
* Add tooling which will search for new rules automatically


jpoppe:
GitHub issues..

PR #1021, opened 2020-07-09
Chore jenkinsfile prevent auto prod deploys

janjaap:
This PR contains changes to the `Jenkinsfile` that prevent automatic deploys to production on merges to `master`. Instead, only tags can be deployed. This change forces us to explicitly tag the `master` branch before each and every release. This also allows a rollback in case something went horribly wrong on deployment, for instance when we forget to include a required configuration prop in one of the `signals-amsterdam` or `signals-weesp` repositories.

Note that:
- the env var `GIT_COMMIT` has been replaced by `GIT_TAG`, which effectively will be a tag or `develop`
- both `version` and `description` props have been removed from `package.json`; neither are required and only make sense when the repository is published through NPM, which we never will. Also, having a `version` prop doesn't make sense when we're already tagging the `master` branch

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,7 +34,7 @@ node('BS16 || BS17') {
                     }
 
                     def buildParams = "--shm-size 1G " +
-                        "--build-arg BUILD_NUMBER=${env.BUILD_NUMBER} " +
+                        "--build-arg BUILD_NUMBER=${env.BUILD_NUMBER} "
 
                     buildParams += IS_TAG ? "--build-arg GIT_BRANCH=${BRANCH} " : ''
                     buildParams += '.'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,6 @@ Boolean IS_TAG = BRANCH ==~ /v(\d{1,3}\.){2}\d{1,3}/
 node('BS16 || BS17') {
     stage("Checkout") {
         def scmVars = checkout(scm)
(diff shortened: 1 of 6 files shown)


janjaap:
@aditudorache The `master` branch will remain the production branch, but it won't be deployed automatically. We will now only the tagged `master` branch.

janjaap:
`IS_TAG` Is a boolean value, as [per line 18](https://github.com/Amsterdam/signals-frontend/pull/1021/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R18). The env var `GIT_BRANCH` is [the name of the branch](https://github.com/Amsterdam/signals-frontend/pull/1021/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R18) that is built or, the reason this PR exists, a tag. A tag is not a branch, but can be referenced as such.

Jenkinsfile
```diff
@@ -13,15 +13,17 @@ def tryStep(String message, Closure block, Closure tearDown = null) {
         }
     }
 }
+
 String BRANCH = "${env.BRANCH_NAME}"
+Boolean IS_TAG = BRANCH ==~ /v(\d{1,3}\.){2}\d{1,3}/
```
jpoppe, line 18: Minor detail, maybe we can improve the `IS_TAG` constant name since we are actually checking for a valid `semantic version` here which could confuse.

PR #1023, opened 2020-07-09
[SIG-2903] Fix the error styling of the fileInput

aditudorache:
This PR addresses a minor fix in the fileInput error style


src/signals/incident/components/form/FileInput/index.js
```diff
@@ -8,6 +10,12 @@ import './style.scss';
 
 export const ERROR_TIMEOUT_INTERVAL = 8000;
 
+const FileInputError = styled.div`
+    color: ${themeColor('secondary')};
+    margin: 16px 0 0;
```
janjaap, line 15: Shouldn't we be using `themeSpacing` here?

aditudorache: Indeed! Thanks!

PR #1024, opened 2020-07-09
Add validation of environment config

boris-arkenaar:
- JSON Schema for `environment.conf.json` in `environment.conf.schema.json`
- Script for validating the config in `validate-config.js`
- Running the script from `start.sh`, exiting in case the config is not valid
- Script can be run with `npm run validate-config`
- Script added to the pre-commit hook
- Script added to the Github CI checks

Any feedback appreciated. Is this clear enough for an administrator in case the config is not valid? How could we improve?

closes Signalen/frontend#61


internals/scripts/validate.sh
```diff
@@ -0,0 +1,3 @@
+npm install --global @exodus/schemasafe
```
bartjkdp, line 1: I would move this step to the Dockerfile, so the Docker container already contains the dependency.

start.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+/validate.sh
+
+if [[ 0 -ne $? ]]; then
```
bartjkdp, line 5: I would use `set -e` at the top of the script instead of `if.... fi`

bartjkdp:
Looking great! Added two comments.

boris-arkenaar:
@bartjkdp Something like this? The `validate-config` step in Github should only be made 'Required' as well. Where do we do that?

bartjkdp:
@boris-arkenaar looking good 👍

jpoppe:
@boris-arkenaar great work :partying_face: , will review this on Monday.

internals/scripts/validate-config.js
```diff
@@ -0,0 +1,14 @@
...
+const schema = require('../../environment.conf.schema.json');
+const config = require('../../environment.conf.json');
+
+const validate = validator(schema, { includeErrors: true });
+const valid = validate(config);
+
+if (!valid) {
+  console.log('environment.conf.json is not valid according to environment.conf.schema.json');
+  console.log(validate.errors);
+  throw new Error('environment.conf.json is not valid according to environment.conf.schema.json');
```
janjaap, line 11: Throwing an error gives quite a messy output. Could we use `process.exit(1)` instead, maybe?

internals/scripts/validate-config.js
```diff
@@ -0,0 +1,14 @@
...
+} else {
+  console.log('environment.conf.json is valid');
```
janjaap, line 13: When the script is run during deployment, it wouldn't be necessary to output anything when the schema validates.

internals/scripts/validate-config.js
boris-arkenaar: Much better!

start.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+set -e
+
+node /internals/scripts/validate-config.js
+
```
janjaap, line 6: This would only be executed when the container is actually spinning up. I'm not sure if that is too late in our CI/CD setup.

bartjkdp: The validation step can be added to pipelines in infrastructure repositories as well, if you would like to do these checks earlier.

PR #1025, opened 2020-07-09
[SIG-2886] New eslint rules part2

jpoppe:
## Future steps (Note to self)

* Move ESlint stuff to it's own plugin/package (will do when mono repo is 'officially' in place)
* Categorize rules 
* Remove Prettier and AirBNB plugins, those two are conflicting by nature and most rules are already superseded, there are nice tools to extract the eventually missing rules
* Proposed rules should be picked up one day, really easy to check rule by rule with **eslint-nibble** (will add separate script)
* ~~Change related React files to jsx (easy fix (a shell one liner is sufficient :sunglasses:) but needs to be planned)~~ **It seems I was wrong about this, default for `typescript` is `tsx` for javascript `js`** (https://github.com/facebook/create-react-app/issues/87)
* Configure even more rules :partying_face:
* Add tooling which will search for new rules automatically


.eslintrc.js
```diff
@@ -22,78 +56,393 @@ module.exports = {
...
+      rules: {
+        'promise/always-return': 'off',
+        'promise/catch-or-return': 'off',
+      },
+    },
+  ],
   rules: {
-    camelcase: 'off',
-    'arrow-parens': ['error', 'as-needed', { requireForBlockBody: false }],
+    // proposed rules
```
janjaap, line 91: Once this is merged, the rules aren't part of a proposal any more. Remove this comment?

.eslintrc.js
```diff
@@ -22,78 +56,393 @@ module.exports = {
...
+    'no-implicit-coercion': 'off',
+    'react/display-name': 'off',
+    'unicorn/prevent-abbreviations': 'off',
+    'unicorn/catch-error-name': 'off',
+    'react/jsx-key': 'off',
+    'sonarjs/no-identical-functions': 'off',
+    'no-promise-executor-return': 'off',
+    'prefer-regex-literals': 'off',
+    'unicorn/no-zero-fractions': 'off', // Auto-fix available
+    'max-len': 'off', // 'max-len': ['warn', { code: 120, tabWidth: 2, ignoreUrls: true, comments: 140 }],
```
janjaap, line 124: `max-len` Should be set to 120 characters. What is the impact of enabling this rule?

.eslintrc.js
```diff
@@ -22,78 +56,393 @@ module.exports = {
...
+    'no-func-assign': 'warn',
+    'no-implied-eval': 'error',
+    'no-invalid-regexp': 'warn',
+    'no-iterator': 'error',
+    'no-label-var': 'error',
+    'no-labels': 'error',
+    'no-lone-blocks': 'error',
+    'no-loop-func': 'warn',
+    'no-mixed-operators': ['warn', {
+      groups: [['&', '|', '^', '~', '<<', '>>', '>>>'], ['==', '!=', '===', '!==', '>', '>=', '<', '<='], ['&&', '||'], ['in', 'instanceof']],
```
janjaap, line 216: This is the default value, isn't it? Is it necessary to explicitly mention it in this rule?

.eslintrc.js
```diff
@@ -22,78 +56,393 @@ module.exports = {
...
+    'unicorn/no-useless-undefined': 'off',
+    'unicorn/prefer-node-append': 'off',
+    'unicorn/prefer-node-remove': 'off',
+    'unicorn/prefer-query-selector': 'off',
+    'unicorn/prefer-string-slice': 'off',
+
+    // disabled rules - related code should be refactored to functional components
+    'no-invalid-this': 'off',
+    'class-methods-use-this': 'off',
+    'react/no-did-mount-set-state': 'off',
```
janjaap, line 446: This is considered to be an anti-pattern. What is the impact of turning this on?

.eslintrc.js
jpoppe: I would love to see most of those rules implemented, but they need manual intervention or to be discussed, I would prefer to keep them here, so I or we can pick up those rules gradually.

.eslintrc.js
jpoppe: Quite some (manual) changes needs to be done which I kept out of scope for this pull request because of time.

.eslintrc.js
jpoppe: I took this from the React Facebook plugin, you are right, will fix this in the next pull request.

.eslintrc.js
jpoppe: The reason is that we still have class components triggered by those rules which require manual intervention to fix.

PR #1029, opened 2020-07-13
[SIG-2910] Added some small scenario's

arjandata:


cypress/integration/changeSignalAfterSubmit.spec.js
```diff
@@ -132,6 +132,9 @@ describe('Change signal after submit', () => {
       cy.wait(1000);
       cy.get(CHANGE_LOCATION.buttonSubmit).click();
 
+      // Check if background colour of changed element is flashing orange
+      createSignal.checkFlashingOrange();
```
janjaap, line 136: It's actually yellow

cypress/support/commandsCreateSignal.js
```diff
@@ -24,6 +25,28 @@ export const checkDescriptionPage = () => {
...
   cy.contains('Voeg een foto toe om de situatie te verduidelijken').should('be.visible');
+  cy.get('body').then($body => {
+    if ($body.find(`${CREATE_SIGNAL.disclaimer}`).length > 0) {
+      checkHeaderFooter();
+    }
+  });
+};
+
+export const checkFlashingOrange = () => {
+  cy.get('.highlight--active')
```
janjaap, line 36: There's a pending change that will no longer have this class name available on the element. The element, however, will be targetable by the attribute `data-testid="highlight"` and, when active, with class name `animate`.

cypress/support/selectorsSignalDetails.js
```diff
@@ -17,7 +18,7 @@ export const CHANGE_STATUS = {
   buttonClose: '[class*="incident-detail__preview-close"]',
```
janjaap, line 18: Close buttons will be targetable by `data-testid="closeButton"`

cypress/integration/changeSignalAfterSubmit.spec.js
arjandata: Wow, difficult to see :-) But you are right.

cypress/support/selectorsSignalDetails.js
arjandata: There is no `data-testid="closeButton"` for the X button when changing status

janjaap: 'will be' Targetable. Not right now, but they will after #712 has been merged. It was more like a heads-up.

PR #1033, opened 2020-07-14
(Chore) Various JSON schema changes

jpoppe:
This pull request:
* Moves schema to `internals/schemas` directory so we could use more schemas in the future, and we could point for example your IDE to this directory
* Use 2 spaces like everywhere else
* Embed descriptions in `schema` so your IDE can guide you

Old screenshot (without Markdown tags removed) but you get the IDE idea:
</img>


docs/configuration/README.md
```diff
@@ -7,272 +7,4 @@ See [the `Dockerfile` in the `signals-amsterdam` repository](https://github.com/
...
-### TERMS_ENDPOINT
-
-URL for the private terms endpoint. Used by the incident management module to `GET` and `POST` default response texts for incident handling.<br>
-See https://api.data.amsterdam.nl/api/swagger/?url=/signals/swagger/openapi.yaml#/default/get_signals_v1_public_terms_categories for reference.
-
-### USERS_ENDPOINT
-
-URL for the private users endpoint. Used by the settings module to `GET`, `PATCH` and `POST` users.<br>
-See https://api.data.amsterdam.nl/api/swagger/?url=/signals/swagger/openapi.yaml#/default/get_signals_v1_private_users_ for reference.
+Properties and their descriptions are defined in [mounted into the container](../../internals/schemas/environment.conf.schema.json).
```
janjaap, line 10: Might as well remove the entire file and replace it with `docs/schemas/README.md`?

jpoppe: Sounds good to me! :+1:

internals/schemas/README.md
```diff
@@ -0,0 +1,10 @@
+# Configuration
```
janjaap, line 1: Why not place this file in the `docs` folder?

jpoppe: Sorry, I thought this is what you meant with: `Might as well remove the entire file and replace it with docs/schemas/README.md?'

jpoppe: So I just remove the file?

janjaap:
@boris-arkenaar Do you also approve the proposed changes in this PR?

Dockerfile
```diff
@@ -70,7 +70,7 @@ COPY start.sh /usr/local/bin/start.sh
 RUN chmod +x /usr/local/bin/start.sh
 
 COPY environment.conf.json /environment.conf.json
-COPY environment.conf.schema.json /environment.conf.schema.json
+COPY internals/schemas/environment.conf.schema.json /environment.conf.schema.json
```
boris-arkenaar, line 73: It needs to be placed in the same location in the container. i.e.

    COPY internals/schemas/environment.conf.schema.json /internals/schemas/environment.conf.schema.json

Otherwise the [validation script](https://github.com/Amsterdam/signals-frontend/pull/1033/files#diff-719e8ac6606473670bd5d07c12443218) won't work on the container

internals/scripts/validate-config.js
```diff
@@ -1,5 +1,5 @@
 const { validator } = require('@exodus/schemasafe');
-const schema = require('../../environment.conf.schema.json');
+const schema = require('../../internals/schemas/environment.conf.schema.json');
```
boris-arkenaar, line 2: This could be more simple:

    const schema = require('../schemas/environment.conf.schema.json');

Dockerfile
jpoppe: Thanks! It was on my list to test, but first wanted to await the reviews, will fix and double check!

internals/scripts/validate-config.js
jpoppe: @boris-arkenaar I propose to even simplify it more and use:
```
npx ajv-cli validate -s internals/schemas/environment.conf.schema.json -d environment.conf.json
``

Which eliminates the need of the custom script, do you agree?

boris-arkenaar: Might work. Would be nice if this removes the need for the script. Needs to be tested though, also from `start.sh` in the container.

jpoppe: Thanks for your thoughts, I am not worried about if it will work since this seems the unwritten `standard` for the `js` world to validate schemas and works fine in my tests.

This tool came up while I am doing investigation for another ticket and is (for me) out of scope of this pull request, thanks!

PR #1040, opened 2020-07-24
Show responsible department on incident detail page

bartjkdp:
It seems responsible department is missing from the incident detail page. It is shown on the split incident page.


bartjkdp:
Now see this was removed on purpose in https://github.com/Amsterdam/signals-frontend/pull/886

PR #1042, opened 2020-07-27
[SIG-2924] Municipality configuration bug

jpoppe:
Disclaimer: I removed quite some code since some of the logic was already moved to the backend. I am not 100% sure I missed something and maybe removed a bit to much, or even more could be removed related to **stadsdeel**


src/components/MapInput/__tests__/MapInput.test.js
```diff
@@ -188,7 +177,10 @@ describe('components/MapInput', () => {
         docs: [],
       },
     };
-    fetch.mockResponseOnce(JSON.stringify(noneFoundResponse)).mockResponseOnce(JSON.stringify(bagResponse));
+    fetch
+      .mockResponseOnce(JSON.stringify(noneFoundResponse))
+      .mockResponseOnce(JSON.stringify(geocoderResponse))
+      .mockResponseOnce(JSON.stringify(bagResponse));
```
janjaap, line 183: This line can go as well since there is no request and response to and from the BAG API

src/components/MapInput/index.js
```diff
@@ -102,17 +90,15 @@ const MapInput = ({ className, hasGPSControl, value, onChange, mapOptions, event
...
         setValuesAction({ location: option.data.location, address: option.data.address, addressText: option.value })
       );
 
-      const stadsdeel = configuration?.map?.options?.stadsdeel || await getStadsdeel(option.data.location);
-
       onChange({
         geometrie: locationTofeature(option.data.location),
         address: option.data.address,
-        stadsdeel,
+        stadsdeel: configuration.map?.municipality,
```
janjaap, line 101: It shouldn't be required any more to provide the API with the `stadsdeel` prop. Can you confirm?

janjaap: As a result `municipality` can be removed from `configuration.map`

jpoppe: Thank you this what I was wondering about, will fix

src/components/MapInput/__tests__/MapInput.test.js
jpoppe: :+1:

janjaap:
@jpoppe The `municipality` props has not been removed from the configuration. Also, tests are failing.

jpoppe:
> @jpoppe The `municipality` props has not been removed from the configuration.
That's because it's still being used

> Also, tests are failing.
No idea why, merged develop and :crossed_fingers:

PR #1043, opened 2020-07-27
[SIG-2994] Hides the send email checkbox for several statuses in the detail form

aditudorache:
This PR hides the send email checkbox in the detail form for the next statuses:
- `Ingepland`, `Heropend`, `Geannuleerd` en `Afgehandeld`

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
@@ -100,10 +100,14 @@ const StyledColumn = styled(Column)`
   position: relative;
 `;
 
+// Don't show the send email checkbox for statuses ingepland, hreopend, afgehandeld and geannuleerd
+const canSendMail = status => !['ingepland', 'reopened', 'o', 'a'].includes(status);
+
 const StatusForm = ({ defaultTexts }) => {
   const { incident, update, close } = useContext(IncidentDetailContext);
   const currentStatus = statusList.find(({ key }) => key === incident.status.state);
   const [warning, setWarning] = useState('');
+  const [showSendMail, setShowSendMail] = useState(canSendMail(currentStatus?.key));
   const isDeelmelding = !!incident?._links?.['sia:parent'];
 
   const form = useMemo(
@@ -120,6 +124,7 @@ const StatusForm = ({ defaultTexts }) => {
     form.controls.state.valueChanges.subscribe(status => {
       const found = statusList.find(s => s.key === status);
 
+      setShowSendMail(canSendMail(found?.key));
       setWarning(found?.warning);
 
       const hasDefaultTexts = Boolean(defaultTextsOptionList.find(s => s.key === status));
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
```diff
@@ -100,10 +100,15 @@ const StyledColumn = styled(Column)`
   position: relative;
 `;
 
+// Don't show the send email checkbox for statuses ingepland, hreopend, afgehandeld and geannuleerd
+export const EMAIL_CHECKBOX_EXCLUDED_STATES = ['ingepland', 'reopened', 'o', 'a'];
+const canSendMail = status => !EMAIL_CHECKBOX_EXCLUDED_STATES.includes(status);
```
janjaap, line 105: Looking at the statuses and the ticket https://datapunt.atlassian.net/browse/SIG-2929, there seems to be a discrepancy. The ticket mentions that the checkbox should only be shown when the current status equals Gemeld, In afwachting van behandeling, In behandeling, On hold, Extern verzoek tot afhandelen. Considering the list of statuses, this means that the checkbox should be shown when the status code equals either 'm', 'i', 'b' or 'closure requested' and should not be shown when the status code equals either 'o', 's', 'ingepland', 'a', 'reopen requested', 'reopened', 'ready to send', 'sent', 'send failed' or 'done external'.

The logic in this PR, whoever, differs from that. Can you explain why it is different?

src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.test.js
```diff
@@ -197,4 +197,26 @@ describe('signals/incident-management/containers/IncidentDetail/components/Statu
     expect(queryByTestId('statusFormSendEmailField')).not.toBeInTheDocument();
     expect(getByText(DEELMELDING_EXPLANATION)).toBeInTheDocument();
   });
+
+  it('hides the  send checkbox when the excludes state is selected', async () => {
+    const { queryByTestId } = render(renderWithContext());
+
+    const defaultRadioButton = queryByTestId('status-m');
+
+    await EMAIL_CHECKBOX_EXCLUDED_STATES.forEach(state => {
```
janjaap, line 206: Why is `await` necessary here? `forEach` Is not an `async` function.

src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
aditudorache: I used the last message from our PO: `belangrijkste is het NIET tonen van het vinkje bij afgehandeld, ingepland en heropend`. I added though the status `a` to it because for this status the message says that no mail will be sent.

src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.test.js
aditudorache: Is not needed. I have removed it!

PR #1045, opened 2020-07-27
[SIG-2993] Updates the styling and texts in the status form

aditudorache:
This PR fixes the alignment and the text in the status form

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
@@ -2,7 +2,7 @@ import React, { useCallback, useState, useEffect, useMemo, useContext } from 're
 import { FormBuilder, FieldGroup, Validators } from 'react-reactive-form';
 import styled, { css } from 'styled-components';
 
-import { Heading, themeSpacing, Row, Column } from '@datapunt/asc-ui';
+import { Heading, themeSpacing, Row, Column, themeColor } from '@datapunt/asc-ui';
 import { defaultTextsType } from 'shared/types';
 import statusList, {
   defaultTextsOptionList,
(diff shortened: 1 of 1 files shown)


src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
```diff
@@ -36,7 +36,7 @@ const Form = styled.form`
     'options'
     'texts'
     'form';
-  grid-column-gap: 20px;
+  grid-column-gap: ${themeSpacing(5)};
```
janjaap, line 39: Not entirely; the value for `grid-column-gap` should come from the theme configuration, since it has been configured there explicitly. It's not 5 times the grid unit value, but a fixed 20 pixels.

src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
```diff
@@ -81,23 +81,25 @@ const Notification = styled.div`
...
+      padding-left: ${themeSpacing(5)};
       border-color: #ec0000;
     `}
 
   line-height: 22px;
 `;
 
 const Wrapper = styled(Row)`
-  padding-top: 20px;
+  background-color: ${themeColor('tint', 'level1')};
```
janjaap, line 93: White remains white, no matter which way you turn it. Replacing white with white seems a bit pointless.

src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
aditudorache: I understand the reason but we need to find an easy/logic way to determine this. My choice would be to describe all with theme spacing for consistency. If we need to decide each time weather are `px` or `themeSpacing` will always get discussions in some edge cases. A solution would be for instance that when coosing for `px`, then explicit to provide a comment why so is clear that should not be changed.

src/signals/incident-management/containers/IncidentDetail/components/StatusForm/index.js
aditudorache: Agree but for the consistency...

PR #1050, opened 2020-07-29
(Chore) Track custom dimension

janjaap:
This PR contains a change for the way page views are tracked in Matomo.

Because of [a known bug in Matomo](https://forum.matomo.org/t/filter-page-urls-by-segments/34859/17), we cannot differentiate between visitors that visit the maintenance part of the website and visitors that merely submit incidents (and aren't authenticated). 

We can, however, track those visits by setting up [custom dimensions](https://matomo.org/docs/custom-dimensions/). This requires a plugin to be installed which is still a pending request.


aditudorache:
Closing this for now. Will be reopened when a new story will be added on the backlog.

PR #1026, opened 2020-07-13
Feature/62 configurable city districts

boris-arkenaar:
- Feature flag `useAreasInsteadOfStadsdeel` to enable fetching 'districts' from the `areas` endpoint, instead of the legacy way of using the hardcoded `stadsdelen`.
- The configuration `areaTypeCodeForDistrict` will determine which `areas` are the 'districts' we need, based on `area_type_code`.
- Fetch city districts from the `areas` endpoint and save them in the store.
- Use them to show city districts in the filters form, for filtering incidents in the back office.
- Filter on city district (when calling the incidents endpoint) by using the unique combination of the general `area_code` and the `area_type_code` of the city district(s) selected.
- The `signals` endpoint needs to be adapted on the backend.

Todo:

- [x] Make the filter param, passed to the incidents endpoint, be the more general `area` instead of `stadsdeel`.
- [x] Use the districts from the store in the split incident form.
- [x] Use the districts from the store in the incident overview.
- [x] Use the districts from the store in the incident detail.

depends on Amsterdam/signals#544

closes Signalen/frontend#62

PR #1065, opened 2020-08-12
Feature/46 configurable sources

boris-arkenaar:
based on #1026

- [x] Allow #1026 to be merged first
- [x] Change the base back to `develop`

The possible sources of a signal will now be fetched from the backend. Loaded on start of the app. There being used in the incident form (when authenticated) and in the back office for filtering incidents.

All is hidden behind a feature flag.

closes Signalen/frontend#46


janjaap:
@boris-arkenaar Can you fix the conflicts in `FilterTagList.test.js` and `IncidentWizard/index.js`?

environment.conf.json
```diff
@@ -112,6 +113,7 @@
   "QUESTIONS_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/public/questions/",
   "ROLES_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/private/roles/",
   "SEARCH_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/private/search",
+  "SOURCES_ENDPOINT": "https://acc.api.data.amsterdam.nl/signals/v1/private/sources/",
```
janjaap, line 116: This prop should be conditionally required; when `fetchSourcesFromBackend` is true, this prop should be present. If `fetchSourcesFromBackend` is false, there is no need to provide a string of uri format.

src/containers/App/index.js
```diff
@@ -69,42 +75,58 @@ export const AppContainer = ({ resetIncidentAction }) => {
     };
   }, [history]);
 
+  useEffect(() => {
+    // prevent continuing (and performing unncessary API calls)
+    // when the current session has not been authenticated
+    if (!isAuthenticated()) return;
+
+    if (configuration.fetchSourcesFromBackend) {
+      getSourcesAction();
```
janjaap, line 84: You can use `useDispatch` here instead of having to use `bindActionCreators` which, in turn, requires you to use `enzyme` in tests to ensure that the `getSourcesAction` function is present in the component's props. 

It doesn't make much sense to assert that a component has a specific (magically injected) prop. We do, however, need to assert that the action is called and that is what should be tested for.

src/containers/App/selectors.js
```diff
@@ -101,24 +101,23 @@ const makeSelectUserCanAccess = createSelector(
...
+export const makeSelectSources = createSelector(selectGlobal, globalState =>
+  globalState.get('sources').size
+    ? globalState
+      .get('sources')
+      .toJS()
+      .map(source => ({
+        key: String(source.id),
+        value: source.name,
+      }))
+    : null
```
janjaap, line 122: Are components that rely on the return value of this selector able to deal with the value `null` instead of getting an array of values?

src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -43,6 +44,7 @@ import reducer, { init } from './reducer';
  * Component that renders the incident filter form
  */
 const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, onUpdateFilter }) => {
+  const { sources } = useContext(AppContext);
```
janjaap, line 47: What if you provide the value for `sources` to the `IncidentManagementContext.Provider` component? That way you'd only have to test a single provider for the `FilterForm` component.

janjaap: True. I was thinking of setting the `sources` prop in the `IncidentManagementContext.Provider`, so that the `incident-management` module only uses the `IncidentManagementContext` instead of using both `IncidentManagementContext` and `AppContext`, but there's no real benefit in doing that, I guess.

src/containers/App/selectors.js
boris-arkenaar: Yes, and just added/updated some unit tests to reflect that.

src/containers/App/index.js
boris-arkenaar: Awesome!

janjaap (approved):

janjaap (approved):

PR #1070, opened 2020-08-17
[SIG-2160] Fixes the navigation back when an error occurs during save

aditudorache:
This PR fixes the back navigation when an error occurs during save of the incident.

Note that this is build on top of the SIG-1553 where the saving process is improved.


aditudorache:
Closing this because the solution was based on changes that were not accepted [SIG-1553].

PR #1077, opened 2020-08-20
[SIG-2160] Adds an extra check to prevent resetting the incident

aditudorache:
This PR prevents the resetting of the incident data by introducing an extra check so that when the current page is `inicdent/fout`, the incident data is not resset


janjaap:
Can we keep this change in conjunction with #1078 or should we close this PR?

aditudorache:
Closing this another solution was chosen to solve the issue

PR #1078, opened 2020-08-20
[SIG-2160] Adds a loading state to the wizzard

aditudorache:
This PR 
- Prevents the resetting of the incident data on error by introducing an `opslaan` step to the wizard.
- The navigation to the `incident/bedankt` occurs only when the saving of the incident is successfull
- This is the preferred solution because on slow internet the address will show `incident/bedankt` during save and will switch to `incident/fout`. 
- With this implementation 'incident/opslaan' will be shown in the address bar during the save operation.

### DIFF
--- a/src/signals/incident/containers/IncidentContainer/saga.js
+++ b/src/signals/incident/containers/IncidentContainer/saga.js
@@ -77,6 +77,7 @@ export function* createIncident(action) {
     }
 
     yield put(createIncidentSuccess(incident));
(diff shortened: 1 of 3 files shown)


janjaap (approved):

janjaap:
Nice solution given the limitations that we're facing by using `react-reactive-form`. I'm still seeing a 27 second timeout locally when throttling the network. Might be worth looking into that.

PR #1081, opened 2020-08-20
[SIG-2566] Adds the incident link in the 'bedankt' page

aditudorache:
This PR 
- Extends the PlainText control with html capablility 
- Disables all html anchors when the user is not authenticated.


src/signals/incident/components/form/PlainText/index.js
```diff
@@ -1,19 +1,47 @@
...
+    ? css`
+            color: ${themeColor('primary')};
+            font-weight: bold;
+          `
+    : css`
+            pointer-events: none;
+            cursor: default;
+            text-decoration: none;
+            color: inherit;
+          `
```
janjaap, line 25: Whitespace use seems off here. Looks like four spaces instead of two.

src/signals/incident/components/form/PlainText/index.js
```diff
@@ -1,19 +1,47 @@
...
+  return (
+    <Span
+      authenticated={isAuthenticated()}
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{
+        __html: text,
+      }}
+    />
+  );
+};
```
janjaap, line 44: Can you add a test case that tests the difference between being authenticated and not authenticated? Now, the snapshot only contains not being authenticated.

janjaap (changes requested):

janjaap (approved):

PR #1039, opened 2020-07-22
[SIG-2909] Translation handling

janjaap:
This PR contains changes that would provide the application with a structure for serving and maintaining (translation) strings.

Consider [ADR 0006](https://github.com/Amsterdam/signals-frontend/blob/05b52f4c91e6a07338ff3d13e53157a18d8b8c54/docs/adr/0006-handling-string-translations.md) for reference.

One (potentially big) hurdle is the dynamic loading of `date-fns/locale` modules based on the value of the `lang` prop in the configuration; since the application is built prior to application of the configuration settings, a construct like

```javascript
import nl from 'date-fns/locale/nl';

registerLocale('nl', nl);
```

cannot be used.

PR #1094, opened 2020-08-27
Add base configuration / simplify configuration of API and Maps base URL

bartjkdp:
This PR simplifies configuration of the Signals frontend.

- It introduces `app.base.json` that contains the default configuration. The environment-specific settings are extended based on the default configuration, in an optional `app.json`.
- OIDC settings are grouped under `oidc.*`.
- Configuring the Signals API can now be done with the setting `apiBaseURL` instead of a setting per endpoint.
- All endpoints have been moved to `src/shared/services/configuration/endpoints.json`.
- Reading and merging the configuration, as well as injecting it, is done in one location: `internals/scripts/helpers/config.js`.

## Todo

- [x] Remove environment.conf.json from the repository and handle case when file does not exist.

## Breaking changes

Update the environment configuration as follows:

- Locally: rename your `environment.conf.development.json` to `app.json`.
- Production: make sure you inject your configuration to `app.json` instead of `environment.conf.json`.
- Add `apiBaseUrl`, `oidc.*`, `map.staticUrl` and `map.layers.*` settings to `app.json` (as long as they need to be different from the default values in `app.base.json`).
- Remove `OIDC_*`, `*_ENDPOINT`, `STATIC_MAP_SERVER_URL`, `OVL_KLOKKEN_LAYER` and `OVL_VERLICHTING_LAYER` settings from `app.json`.
- Now remove anything that's left in `app.json` which defines the same values as defined in `app.base.json`.

Closes Signalen#71


jpoppe:
I have been looking into the pull request and I think this is a nice step in the right direction, thanks!

That said, I am worried about the amount of places and scripts being used to be able to validate a schema and then we also have the `inject` magic, this is hard to follow.

I would prefer too see 1 script (or maybe 2 when shell is still needed) which could perform the required validation/merging tasks, so we keep this logic in one place and then we would also be able to use this from the  `Amsterdam/signalen` repository.

Maybe we can discuss this tomorrow (Thursday).

bartjkdp:
Thanks for your feedback. I think centralizing the logic for config parsing and validation is a good idea. @boris-arkenaar will take this pull request further as I am on holiday next week.

jpoppe:
> Thanks for your feedback @jpoppe. I think centralizing the logic for config parsing and validation is a good idea. @boris-arkenaar will take this pull request further as I am on holiday next week.

Thanks for the reply and enjoy (your probably well deserved) holiday! 🌴

README.md
```diff
@@ -54,14 +52,14 @@ Running all tests and generate the coverage report:
 When deploying the docker container, there are some files you can overwrite for
 branding purposes.
 
-- `environment.conf.json` with configuration for theming and endpoint URLs.
-- The logo, as defined by `logo.url` in `environment.conf.json`.
+- `environment.conf.json` with specific configuration for theming and endpoint URLs.
+- The logo, as defined by `logo.url` in `environment.base.conf.json`.
```
boris-arkenaar, line 56: Maybe we can rename these files to `configuration.json` and `configuration.base.json` respectively. I find the current name a bit long and confusing.

boris-arkenaar: Ended up renaming it to `app.json` and `app.base.json` respectively. `configuration` is still a bit to vague (configuration for what). `app` makes this clear and the fact that it's `json` and in the root makes it clear enough to me that it's configuration.

boris-arkenaar:
I've updated @bartjkdp s first message with all the changes made and the breaking changes.

src/shared/services/configuration/endpoints.js
```diff
@@ -0,0 +1,10 @@
+import isNil from 'lodash.isnil';
```
janjaap, line 1: Do we really need an external package to check if something is `null` or `undefined`?

src/shared/services/configuration/endpoints.json
```diff
@@ -0,0 +1,22 @@
...
+  "INCIDENT_PRIVATE_ENDPOINT": "/signals/v1/private/signals/",
+  "INCIDENT_PUBLIC_ENDPOINT": "/signals/v1/public/signals/",
+  "PERMISSIONS_ENDPOINT": "/signals/v1/private/permissions/",
+  "PREDICTION_ENDPOINT": "/signals/category/prediction",
+  "QUESTIONS_ENDPOINT": "/signals/v1/public/questions/",
+  "ROLES_ENDPOINT": "/signals/v1/private/roles/",
+  "SEARCH_ENDPOINT": "/signals/v1/private/search",
+  "SOURCES_ENDPOINT": "/signals/v1/private/sources/",
+  "TERMS_ENDPOINT": "/signals/v1/private/terms/categories/",
+  "USERS_ENDPOINT": "/signals/v1/private/users/"
```
janjaap, line 21: All of the original descriptions in the schema are lost. Can they be added here as comments?

internals/scripts/helpers/config.js
```diff
@@ -0,0 +1,37 @@
...
+  $SIGNALS_CONFIG: JSON.stringify(config),
+  $SIGNALS_FAVICON: config.head.favicon,
+  $SIGNALS_IOS_ICON: config.head.iosIcon,
+  $SIGNALS_MATOMO_SITE_ID: config.matomo.siteId,
+  $SIGNALS_MATOMO_URL_BASE: config.matomo.urlBase,
+  $SIGNALS_PWA_SHORT_TITLE: config.language.shortTitle,
+  $SIGNALS_PWA_TITLE: config.language.title,
+  $SIGNALS_SITE_TITLE: config.language.siteTitle,
+  $SIGNALS_STATUS_BAR_STYLE: config.head.statusBarStyle,
+  $SIGNALS_THEME_COLOR: config.head.themeColor,
```
janjaap, line 22: Since these props are no longer env vars, we don't have to prefix them with `$`, do we?

internals/scripts/validate-config.js
```diff
@@ -1,12 +1,14 @@
 const { validator } = require('@exodus/schemasafe');
-const schema = require('../schemas/environment.conf.schema.json');
-const config = require('../../environment.conf.json');
+
```
janjaap, line 2: For the translation handling (#1039), I've rewritten this file so that not just the first validation error will be shown, but all the errors that were found. Helps with solving all mismatches in the schema at once. I'd propose changing this file's contents to:

```javascript
const { validator } = require('@exodus/schemasafe');

/* eslint-disable no-console, global-require */

const validationSections = {
  config: {
    schema: require('../schemas/app.schema.json'),
    json: require('../../app.json'),
  },
};

let hasValidationErrors = false;

Object.entries(validationSections).forEach(([section, { schema, json }]) => {
  const validate = validator(schema, { includeErrors: true, allErrors: true });

  if (!validate(json)) {
    console.error(`${section} is not valid`);
    console.table(validate.errors);

    hasValidationErrors = true;
  }
});

if (hasValidationErrors) {
  process.exit(1);
}
```

That way we can easily add other schemas at a later point, when needed.

janjaap (changes requested):

internals/scripts/helpers/config.js
boris-arkenaar: They've never been env vars. The `$` is a nice way to avoid any conflicts since it searches and replaces all of these strings in `index.html`.

janjaap: They were, as per the contents of `start.sh`.

internals/scripts/validate-config.js
boris-arkenaar: Including all errors is a good idea!
It nice that we could easily add more schema's with this script. However, at the moment we have only one schema to check and I am not a fan of adding logic and complexity that might be used in the future, while it is not necessary now.

src/shared/services/configuration/endpoints.json
boris-arkenaar: Good one! How about this?

src/shared/services/configuration/endpoints.json
```diff
@@ -1,22 +1,42 @@
 {
+  "_AREAS_ENDPOINT": "Retrieves geographical areas\n\nSee https://api.data.amsterdam.nl/api/swagger/?url=/signals/swagger/openapi.yaml#/default/get_signals_v1_private_areas_geography_ for reference",
```
janjaap, line 2: Why not turn this file into a `.js` file? That way you can return an object in which you can actually place comments instead of having to resort to polluting the JSON output.

boris-arkenaar: That's better indeed!

internals/scripts/helpers/config.js
boris-arkenaar: You are right indeed!

aditudorache (dismissed):
Approved, will make the necessary changes in our configuration repository before merging this. Let's make a note to include the JJ error proposal in the next iteration

aditudorache:
@boris-arkenaar, there is still an error in the ci Test build regarding matomo, could you take a look at that?

boris-arkenaar:
@aditudorache All errors will be logged on config validation now.
Removing the `$` from the config placeholders in `index.html` caused the issue with the build. So. I put them back.

aditudorache (approved):

aditudorache:
The changes in the SIA configuration repository are done, this PR can be merged.

janjaap (approved):

PR #1101, opened 2020-09-01
[SIG-1848] User history

janjaap:
This PR contains changes that render a user's edit history in the user detail page. For that, this PR:
- Adds an extra request to the `UserDetail` component to retrieve the history
- Applies layout changes to the `UserForm` component so that its children are properly aligned
- Rewrites all the tests for the `UserDetail` component
- Applies a fix to the `useFetchResponseNotification` hook that prevents unnecessary rerenders for every component that uses that hook.


jpoppe (approved):
Like the fixture names, and tests are looking really clean 👍🏻 , nice!

PR #1110, opened 2020-09-02
[SIG-3088] Refactor KTO form

janjaap:
This PR rewrites both the `KtoContainer component as well as its inner form component. All redux/saga boilerplate has been removed, since it's not necessary for the form's functionality. 

Judging by the amount of changed files, this PR might seem a tad big, but it's core only consists of four changed files, a lot of moved files and a couple of minor changes to supporting components as well as some unrelated snapshots (because of the changes to the supporting components).


jpoppe (approved):
Looks much better, thanks! :)

PR #1116, opened 2020-09-04
[SIG-3103] Fix incident date filter

jpoppe:
IMHO the backend should accept dates without the timestamps so we don't need this prone to error logic on the front-end.

When the dates from the filter form are stored the reducer will append time strings to the dates.

When a user re-opens the filter form when dates are stored, the reducer assumes dates without time string,s this pull request fixes this by stripping the time string before the date gets further processed by the reducer.

### DIFF
--- a/src/signals/shared/filter/__tests__/parse.test.js
+++ b/src/signals/shared/filter/__tests__/parse.test.js
@@ -4,12 +4,37 @@ import categories from 'utils/__tests__/fixtures/categories_private.json';
 import { filterForSub, filterForMain } from 'models/categories/selectors';
 import dataLists from 'signals/incident-management/definitions';
 
-import { parseOutputFormData, parseInputFormData, parseToAPIData } from '../parse';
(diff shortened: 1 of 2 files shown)


janjaap:
@jpoppe This fix doesn't quite seem to fix the issue. Consider the following:
- Open the filter form
- Set the 'Vanaf' date and click 'Filteren'
- Open the filter form again
- Click 'Filteren'
Note that after clicking on the 'Filteren' button for the second time, the value in any date field is gone.

jpoppe:
> @jpoppe This fix doesn't quite seem to fix the issue. Consider the following:
> 
> * Open the filter form
> * Set the 'Vanaf' date and click 'Filteren'
> * Open the filter form again
> * Click 'Filteren'
>   Note that after clicking on the 'Filteren' button for the second time, the value in any date field is gone.

I have tested it again and works like expected here, will ping you to double check..

janjaap (approved):

PR #1121, opened 2020-09-08
[SIG-2969] Split container

janjaap:
This PR adds the `IncidentSplitContainer` component that:
- fetches incident data on mount
- renders a form (for now, uses a stub component which needs to be replaced)
- posts data on form submission
- shows a notification on error or success and redirects to the incident page


jpoppe (approved):
Thank you ❤️

PR #1123, opened 2020-09-09
[SIG-3040] Styling "Thank you" page

janjaap:
This PR styles the submission 'Thank you' page differently from the rest of the form.

It:
- doesn't render the navigation bar
- renders the page container as `display: block` instead of `display: grid` to have the contents span the full width of the page instead of falling in the grid's defined columns
- turns a regular text block into a heading
- cleans up unnecessary code from the `IncidentNavigation` component and replaces all of its tests


jpoppe (approved):

src/signals/incident/components/IncidentNavigation/index.test.js
```diff
@@ -12,204 +15,124 @@ jest.mock('shared/services/auth/auth', () => ({
...
-  let withWizard;
-
-  function getComponent() {
-    wrapper = mount(
-      <Wizard history={historySpy}>
-        <IncidentNavigation {...props} />
-      </Wizard>
-    );
+const steps = Object.keys(wizardDefinition)
+  .filter(key => !['opslaan'].includes(key))
```
boris-arkenaar, line 19: `  .filter(key => key !== 'opslaan')`

src/signals/incident/components/IncidentNavigation/index.test.js
```diff
@@ -12,204 +15,124 @@ jest.mock('shared/services/auth/auth', () => ({
...
+            </Step>
+          </Steps>
+        </Wizard>
+      )
+    );
 
-      expect(wrapper).toMatchSnapshot();
-      expect(withWizardWrapper).toMatchSnapshot();
-    });
+    expect(getByTestId('nextButton')).toBeInTheDocument();
```
boris-arkenaar, line 49: You could test here that the previous button does not exist.

src/signals/incident/components/IncidentNavigation/index.test.js
```diff
@@ -12,204 +15,124 @@ jest.mock('shared/services/auth/auth', () => ({
...
+              <IncidentNavigation {...props} />
+            </Step>
+          </Steps>
+        </Wizard>
+      )
+    );
 
-      expect(props.meta.handleSubmit).toHaveBeenCalledWith(event, context.wizard.next, undefined);
-    });
+    expect(getByTestId('previousButton')).toBeInTheDocument();
```
boris-arkenaar, line 84: Test here the next button does not exist.

src/signals/incident/components/IncidentNavigation/index.test.js
```diff
@@ -12,204 +15,124 @@ jest.mock('shared/services/auth/auth', () => ({
...
+        </Wizard>
+      )
+    );
 
-      withWizardWrapper.find('[data-testid="nextButton"]').simulate('click', event);
+    expect(handleSubmit).not.toHaveBeenCalled();
 
-      expect(props.meta.handleSubmit).toHaveBeenCalledWith(event, context.wizard.next, 'CREATE_INCIDENT');
+    act(() => {
+      fireEvent.click(getByTestId('nextButton'));
```
boris-arkenaar, line 135: You're missing a check that `handleSubmit` has been called after this statement.

boris-arkenaar (commented):
Wouldn't it be better to make this thank you page a separate page instead of a step in the wizzard? Since now you have some very explicit logic in the code for making exceptions for that step.

boris-arkenaar (approved):
> Wouldn't it be better to make this thank you page a separate page instead of a step in the wizzard? Since now you have some very explicit logic in the code for making exceptions for that step.

What about this? ☝️

aditudorache (approved):

janjaap:
> > Wouldn't it be better to make this thank you page a separate page instead of a step in the wizzard? Since now you have some very explicit logic in the code for making exceptions for that step.
> 
> What about this? ☝️

Agreed. But out of scope for this ticket.

PR #1128, opened 2020-09-10
(Chore) Fix async test issues

janjaap:
This PR resolves issues in test cases where the underlying components relied on the response of (async) fetch requests. Not all requests were properly awaited and caused error later on in unrelated test cases/suits.

### DIFF
--- a/src/components/Footer/index.test.js
+++ b/src/components/Footer/index.test.js
@@ -5,19 +5,19 @@ import configuration from 'shared/services/configuration/configuration';
 
 import Footer from '.';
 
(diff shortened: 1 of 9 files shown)


boris-arkenaar (approved):
Looks a lot cleaner!

boris-arkenaar:
In some tests the configuration would be altered, but not reset at the end. Not sure if this could potentially be the cause of any problems in subsequent tests, but better to clean it up anyway. All tests that alter the configuration now perform a reset of the configuration in the `afterEach` function.

boris-arkenaar (approved):

PR #1129, opened 2020-09-10
[SIG-2762] ASC-UI upgrade

janjaap:
This PR upgrades `@datapunt/asc-ui` to its latest stable version and
- Makes sure that everywhere in the application the `components/Checkbox` and `component/RadioButton` components are used. These contain subtle changes to the styling of the components coming from `asc-ui` to correct alignment
- Uses the `asc-ui` `Label` component as much as possible where it is used as a wrapper for any checkbox or radio button element

In addition to the above, a few `useEffect` hook dependencies are fixed, because of a change made to the `useFetch` hook in an earlier PR (#1111).


jpoppe (approved):

aditudorache:
Threre are still some unittests failing, it looks that the extra events introduced in the (asc)select component are the problem

boris-arkenaar (dismissed):

internals/webpack/webpack.base.babel.js
```diff
@@ -142,6 +142,12 @@ module.exports = options => ({
     .concat(options.plugins)
     .filter(Boolean),
   resolve: {
+    // Setting aliases to make sure that, whenever dependent packages are linked, shared dependencies won't conflict
```
boris-arkenaar, line 145: You get conflicts when you link another package?
I don't like explicitly naming and aliasing packages here.

package.json
```diff
@@ -51,8 +51,8 @@
   ],
   "dependencies": {
     "@datapunt/amsterdam-react-maps": "^0.1.2",
-    "@datapunt/asc-assets": "^0.24.0",
-    "@datapunt/asc-ui": "^0.24.0",
+    "@datapunt/asc-assets": "^0.24.1",
+    "@datapunt/asc-ui": "^0.23.1-alpha.54",
```
boris-arkenaar, line 55: Why do you go back to a previous, alpha version?

internals/webpack/webpack.base.babel.js
janjaap: Getting conflicts when linking `@datapunt/asc-ui`. It conflicts with, at least, Styled components. Setting explicit aliases fixes that problem.

boris-arkenaar (approved):

aditudorache (approved):

PR #1130, opened 2020-09-11
(Fix) Prop warnings and hooks usage

janjaap:
This PR fixes a prop warning that was shown in the incident overview page. The cause was the use of an internal variable name that was the same as a prop in that same component.

In addition, the `FormAlert` component is removed; wasn't used any more. 

For the `List` and `Pagination` components, internal variables have been wrapped in a `useMemo` call to prevent memory leaks.


boris-arkenaar (approved):
Nice refactoring!

boris-arkenaar:
Tests indicate somethings not yet right though.

aditudorache (dismissed):

jpoppe (approved):

boris-arkenaar (approved):

PR #1139, opened 2020-09-16
Put map filtering under feature flag

boris-arkenaar:
In the back office, the map was fixed to showing only the incidents of the last 24 hours. This has now been put under a feature flag `mapFilter24Hours`, which is true by default.
When turned off, the map will show all incidents (still with a limit of 4000) and the start and end date filter now have an effect on the map as well.

- [x] Update unit tests


internals/schemas/app.schema.json
```diff
@@ -73,6 +77,7 @@
         "links",
         "logo",
         "map",
+        "mapFilter24Hours",
```
janjaap, line 80: Shouldn't be required, right?

src/signals/incident-management/containers/IncidentOverviewPage/components/Map/index.js
```diff
@@ -103,14 +103,21 @@ const OverviewMap = ({ showPanelOnInit, ...rest }) => {
...
+        ? format(subDays(new Date(), -1), "yyyy-MM-dd'T'HH:mm:ss")
+        : filterParams.created_after,
+      created_before: configuration.mapFilter24Hours
+        ? format(new Date(), "yyyy-MM-dd'T'HH:mm:ss")
+        : filterParams.created_before,
+      // fixed page size (default is 50; 4000 is 2.5 times the highest daily average)
+      page_size: 4000,
+    }),
+    [filterParams]
+  );
```
janjaap, line 120: Make sure to test the conditions in this block.

janjaap (changes requested):

internals/schemas/app.schema.json
boris-arkenaar: Debatable. All boolean parameters could be seen as not-required. If they're not specified, they'll get a value of false implicitly. However, the code base requires them to have a value; they are either true or false, undefined doesn't mean anything. So in that sense, they are required. And I would rather have their values explicitly defined in the configuration than implicitly set.

janjaap: Not entire true that `undefined` doesn't mean anything. Looking at the condition

```jsx
{showsMap && configuration.mapFilter24Hours && <MapHeading>Afgelopen 24 uur</MapHeading>}
```

`mapFilter24Hours` is expected to be truthy and in javascript that can be a lot of things. Where at the same time `undefined` is not truthy. Unless you specifically check for `configuration.mapFilter24Hours === true`, it doesn't matter if the config parameter is present or not.

boris-arkenaar: That is totally true!

I was more thinking on the level of what the parameter means, represents. It's a feature flag, either on or off, true or false (truthy or falsy).
And it's different from a parameter like `matomo`, whose value should be an object which can contain different values to configure matomo. It is not required, though, and when it is not defined, matomo is not configured, and thus turned off in the application.
This makes me realize that matomo is also turned off implicitly and thus undermining my argument not allowing feature flags to be turned off implicitly.

So, we could make `mapFilter24Hours` not required, but then I would consistently make all feature flags not required (`fetchDistrictsFromBackend`, `fetchQuestionsFromBackend`, `fetchSourcesFromBackend`, `showVulaanControls`, `useStaticMapServer`).

jpoppe: I also prefer to be `explicit` with the configuration options and make it required so every 'domain developer' will be aware of this option by just looking to the specific domain configuration, and if we will ever switch to strong typing it will also help.

boris-arkenaar: Okay. All required for now.
Good topic of discussion, but out of scope for this PR.

Good to think about this. What are the base settings (Amsterdam specific vs neutral). Not defined vs explicitly defined as false or null.
I just realized that sentry and matomo are enabled for other municipalities now with the change in configuration. Since these settings are not defined, because not needed. Worked fine in the past, but now it keeps the default setting of Amsterdam. So, should be explicitly overridden with null for every other deployment besides Amsterdam. Or should be undefined in the base config and Amsterdam should explicitly set these values for their deployments because Amsterdam needs sentry and matomo.

janjaap:
@boris-arkenaar Is renaming files from `index.js` to `<ComponentName>.js` a deliberate decision in consensus? It deviates from the approach in the rest of the project.

boris-arkenaar:
@janjaap Yes this is deliberate. In some places this is already the case in the code base, but most of the time the components are called `index.js` which is very inconvenient in your IDE have tons of `index.js` files open. The `index.js` file is still there as a barrel file though.
I mentioned this briefly in a dev meeting once and decided to make this change for files I touch.

aditudorache:
@boris-arkenaar, what would be your suggestion for the naming strategy for actions.js, selectors.js and constants.js from the containers? There are also a lot of them.

boris-arkenaar:
@aditudorache Keep those file names as they are.

janjaap (approved):

PR #1140, opened 2020-09-16
ADR on forms

boris-arkenaar:


aditudorache (dismissed):
Nice research! Totally agree!

docs/adr/0006-forms.md
```diff
@@ -0,0 +1,57 @@
...
+- *Must have* integration with React hooks.
+- *Must have* validation.
+
+
+| Package                                                               | No run-time deps. | < 10 kB | Hooks | Validation | Customizable | No re-render | Stargazers | Updated        |
+| --------------------------------------------------------------------- | ---------------- | -------- | ----- | ---------- | ------------ | ------------ | ---------- | -------------- |
+| [formik](https://github.com/jaredpalmer/formik)                       | ❌                | ❌        | ❌     | ✅*         | ✅            | ❌            | 24.5k      | 2 months ago   |
+| [react-hook-form](https://github.com/react-hook-form/react-hook-form) | ✅                | ✅        | ✅     | ✅**        | ✅            | ✅            | 14.1k      | recently       |
+| plain HTML                                                            | ✅                | ✅        | ❌     | ✅          | ✅            | ✅            | –          | –              |
+| [react-reactive-form](https://github.com/bietkul/react-reactive-form) | ✅                | ✅        | ❌     | ✅          | ✅            | ❌            | 250        | 25 months ago  |
```
janjaap, line 44: What I'm missing here is the option to use our own (controlled) components. Since we're dealing with components coming from `@datapunt/asc-ui`, this is a hard requirement.

docs/adr/0006-forms.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Decision
+
+Only [react-hook-form](https://github.com/react-hook-form/react-hook-form) has integration with React hooks and doesn't rerender on input changes to form fields. Furthermore, it has a small footprint, built in validation, requires very little code to use and follows HTML standards closely.
+
+Generating a form based on a definition coming from the back end can easily be done with this library using native JavaScript/React logic.
+
+## Consequences
+
+These libraries can exist side by side in the code base. New forms can be built using the new [react-hook-form](https://github.com/react-hook-form/react-hook-form) while existing forms still use [react-reactive-form](https://github.com/bietkul/react-reactive-form). Existing forms can be re-written when decide it would be worth the effort.
```
janjaap, line 57: Can you add an overview of which forms are in the applicatie, which components they use and how those components should be rewritten so that they can be used with the proposed lib and in which way?

janjaap (changes requested):

docs/adr/0006-forms.md
boris-arkenaar: The column `customizable` refers to the bullet "*Should be* easy to customize, or use custom components, validation, localization." So, using custom components is covered in that column.

docs/adr/0006-forms.md
```diff
@@ -0,0 +1,112 @@
...
+| [formik](https://github.com/jaredpalmer/formik)                       | ❌                | ❌        | ❌     | ✅*         | ✅            | ❌            | 24.5k      | 2 months ago   |
+| [react-hook-form](https://github.com/react-hook-form/react-hook-form) | ✅                | ✅        | ✅     | ✅**        | ✅            | ✅            | 14.1k      | recently       |
+| plain HTML                                                            | ✅                | ✅        | ❌     | ✅          | ✅            | ✅            | –          | –              |
+| [react-reactive-form](https://github.com/bietkul/react-reactive-form) | ✅                | ✅        | ❌     | ✅          | ✅            | ❌            | 250        | 25 months ago  |
+
+\* Only support for 3rd party validation, integrated support for [Yup](https://github.com/jquense/yup).
+** Integrated validation follows the [HTML standard for form validation](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation). Support for 3rd party validation, integrated support for [Yup](https://github.com/jquense/yup) [Joi](https://github.com/hapijs/joi) and [Superstruct](https://github.com/ianstormtaylor/superstruct).
```
aditudorache, line 47: Could you add an `Enter` here for better readablity in the markdown preview
```
\* Only...
<enter>
\** Integrated validation follows the ...
```

docs/adr/0006-forms.md
```diff
@@ -59,33 +59,33 @@ These libraries can exist side by side in the code base. New forms can be built
...
 ### Incident form
 
-- src/signals/incident/components/IncidentForm/index.js
+- [IncidentForm/index.js](../../src/signals/incident/components/IncidentForm/index.js)
 
 Two form components specifically using the dependency. These can easily be updated when using the new lib.
 
-- src/signals/incident/components/form/EmphasisCheckboxInput/index.js
-- src/signals/incident/components/form/Header/index.js
+- [form/Header/index.js](../../src/signals/incident/components/form/Header/index.js)
```
aditudorache, line 66: This link doesn't work

boris-arkenaar: There were two of theme next to each other. Removed one. The link works for me however. Does it work for you now?

jpoppe:
Nice ADR, I ❤️ your steps section, this is also how I would probably approach it. Thanks!

docs/adr/0006-forms.md
aditudorache: In VSCode doesn't link correct

boris-arkenaar: Ah, it does on Github.

aditudorache (dismissed):

docs/adr/0006-forms.md
janjaap: Formik definitely doesn't support custom components, so I would split that column up in its appropriate parts, because now we're losing overview because of grouping of criteria.

janjaap (approved):

PR #1141, opened 2020-09-16
[SIG-3092] Set subcategory on url

aditudorache:
This PR implements the prefill of the category/subcategory from the url and disable the prediction service when the category/subcategory is prefilled

### DIFF
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -103,7 +103,7 @@ export const AppContainer = () => {
               <Route path="/instellingen" component={SettingsModule} />
               <Route path="/incident" component={IncidentContainer} />
               <Route path="/kto/:satisfactionIndication/:uuid" component={KtoContainer} />
-              <Route path="/category/:category/:subcategory" component={IncidentClassification} />
+              <Route path="/categorie/:category/:subcategory" component={IncidentClassification} />
               <Route component={NotFoundPage} />
             </Switch>
           </ContentContainer>
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -22,6 +22,7 @@ import KtoContainer from 'signals/incident/containers/KtoContainer';
 import useLocationReferrer from 'hooks/useLocationReferrer';
 import useIsFrontOffice from 'hooks/useIsFrontOffice';
 
+import IncidentClassification from 'signals/incident/components/IncidentClassification';
 import { getSources } from './actions';
 import AppContext from './context';
 import reducer from './reducer';
@@ -102,6 +103,7 @@ export const AppContainer = () => {
               <Route path="/instellingen" component={SettingsModule} />
               <Route path="/incident" component={IncidentContainer} />
               <Route path="/kto/:satisfactionIndication/:uuid" component={KtoContainer} />
+              <Route path="/category/:category/:subcategory" component={IncidentClassification} />
               <Route component={NotFoundPage} />
             </Switch>
           </ContentContainer>
--- a/src/signals/incident/components/IncidentClassification/index.js
+++ b/src/signals/incident/components/IncidentClassification/index.js
@@ -5,6 +5,7 @@ import configuration from 'shared/services/configuration/configuration';
 import { useHistory, useParams } from 'react-router-dom';
 import { useDispatch } from 'react-redux';
 import { setClassification } from 'signals/incident/containers/IncidentContainer/actions';
+import { isAuthenticated } from 'shared/services/auth/auth';
 import LoadingIndicator from 'components/LoadingIndicator';
 
 const IncidentClassification = () => {
@@ -14,11 +15,16 @@ const IncidentClassification = () => {
   const dispatch = useDispatch();
 
   useEffect(() => {
-    get(`${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`);
-  }, [category, subcategory, get]);
+    if (isAuthenticated()) {
+      history.push('/');
+    }
+    else {
+      get(`${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`);
+    }
+  }, [category, subcategory, get, history]);
 
   useEffect(() => {
-    if (data) dispatch(setClassification({ category, subcategory }));
+    if (data && data.is_active) dispatch(setClassification({ category, subcategory }));
     if (data || error) history.push('/');
   }, [data, error, history, dispatch, category, subcategory]);
 
--- a/src/signals/incident/components/IncidentClassification/index.js
+++ b/src/signals/incident/components/IncidentClassification/index.js
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import React, { useEffect } from 'react';
 import { useFetch } from 'hooks';
 import configuration from 'shared/services/configuration/configuration';
 
@@ -14,7 +14,7 @@ const IncidentClassification = () => {
 
   useEffect(() => {
     get(`${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`);
-  }, [category, subcategory, get, history]);
+  }, [category, subcategory, get]);
 
   useEffect(() => {
     if (!data && !error) return;
@@ -23,10 +23,10 @@ const IncidentClassification = () => {
     history.push('/');
   }, [data, error, history, dispatch, category, subcategory]);
 
-  // This component doesn't render anything, it is used to build the logic
+  // This component is empty, it is used to build the logic
   // for setting the category and subcategory from the url
   // before redirecting to the incident page
-  return null;
+  return <div data-testid="incidentClassification"></div>;
 };
 
 export default IncidentClassification;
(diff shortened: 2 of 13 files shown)


src/containers/App/index.js
```diff
@@ -102,6 +103,7 @@ export const AppContainer = () => {
               <Route path="/instellingen" component={SettingsModule} />
               <Route path="/incident" component={IncidentContainer} />
               <Route path="/kto/:satisfactionIndication/:uuid" component={KtoContainer} />
+              <Route path="/category/:category/:subcategory" component={IncidentClassification} />
```
janjaap, line 106: Wel netjes om het URL Nederlands te houden

src/signals/incident/components/IncidentClassification/index.js
```diff
@@ -0,0 +1,32 @@
...
+    if (!data && !error) return;
+
+    if (data) dispatch(setClassification({ category, subcategory }));
+    history.push('/');
+  }, [data, error, history, dispatch, category, subcategory]);
+
+  // This component is empty, it is used to build the logic
+  // for setting the category and subcategory from the url
+  // before redirecting to the incident page
+  return <div data-testid="incidentClassification"></div>;
```
janjaap, line 29: You don't need to render anything here, right? You can test for changes in `history.location.pathname`.

src/signals/incident/components/IncidentClassification/index.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+    act(() => history.push('/initial-location'));
+
+    const { findByTestId } = render(withAppContext(<IncidentClassification />));
+
+    expect(fetch).toHaveBeenCalledWith(`${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`, expect.objectContaining({
+      method: 'GET',
+    }));
+
+    await findByTestId('incidentClassification');
```
janjaap, line 44: Instead of having to render an empty HTML element, you can also do this:
```jsx
await waitFor(() => expect(dispatch).toHaveBeenCalledWith(setClassification({ category, subcategory })));
```

src/signals/incident/components/IncidentClassification/index.js
```diff
@@ -0,0 +1,32 @@
...
+  const { category, subcategory } = useParams();
+  const { get, data, error } = useFetch();
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    get(`${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`);
+  }, [category, subcategory, get]);
+
+  useEffect(() => {
+    if (!data && !error) return;
```
janjaap, line 20: What if there is a `500` or `404` response from the API? Can you add a test for that?

janjaap (changes requested):

src/containers/App/index.js
aditudorache: incident is not Nederlands :-). Should I change that too?

aditudorache: done

src/signals/incident/components/IncidentClassification/index.test.js
aditudorache: wouldn't work for the other test where `expect(dispatch).not.toHaveBeenCalled();`

src/signals/incident/components/IncidentClassification/index.js
aditudorache: Not rendering anything would create testing problems when for instance some function should NOT be called in between renders because there would not be posible to await the render.

src/containers/App/index.js
aditudorache: done

src/signals/incident/components/IncidentClassification/index.js
aditudorache: tests added

boris-arkenaar: Why not explicitly write:
```
    if (data) dispatch(setClassification({ category, subcategory }));
    if (data || error) history.push('/');
```

src/signals/incident/components/IncidentClassification/index.js
boris-arkenaar: Render a loading indicator?

src/signals/incident/components/IncidentClassification/index.test.js
```diff
@@ -0,0 +1,111 @@
...
+  const dispatch = jest.fn();
+
+  beforeEach(() => {
+    jest.spyOn(reactRedux, 'useDispatch').mockImplementation(() => dispatch);
+
+    fetch.resetMocks();
+    dispatch.mockReset();
+  });
+
+  afterEach(() => {
```
boris-arkenaar, line 26: Empty. Can be removed.

src/signals/incident/components/IncidentClassification/index.js
aditudorache: Good suggestions, thanks!

src/signals/incident/components/IncidentClassification/index.js
aditudorache: Done!

src/signals/incident/components/IncidentClassification/index.test.js
aditudorache: Sure!

src/signals/incident/components/form/DescriptionWithClassificationInput/index.test.js
```diff
@@ -122,5 +123,25 @@ describe('Form component <DescriptionWithClassificationInput />', () => {
...
+          meta: {
+            ...parent.meta,
+            incidentContainer: { usePredictions: false },
+          },
+        },
+      });
+
+      wrapper.find(TextArea).simulate('blur', { target: { value: 'diabolo' } });
+      expect(parent.meta.getClassification).not.toHaveBeenCalled();
+      expect(parent.meta.updateIncident).not.toHaveBeenCalledWith();
```
boris-arkenaar, line 144: I expect you mean to use `toHaveBeenCalled` here as well, like the line above?

boris-arkenaar (dismissed):

src/signals/incident/components/form/DescriptionWithClassificationInput/index.test.js
aditudorache: Indeed.

src/signals/incident/components/IncidentClassification/index.js
```diff
@@ -13,20 +15,23 @@ const IncidentClassification = () => {
   const dispatch = useDispatch();
 
   useEffect(() => {
-    get(`${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`);
-  }, [category, subcategory, get]);
+    if (isAuthenticated()) {
+      history.push('/');
```
janjaap, line 19: Why override setting the category and subcategory for authenticated users? The AC describe that the ML service should still be queried, but not that the category should not be set.

src/signals/incident/components/IncidentClassification/index.js
```diff
@@ -13,20 +15,23 @@ const IncidentClassification = () => {
...
+    }
+  }, [category, subcategory, get, history]);
 
   useEffect(() => {
-    if (!data && !error) return;
-
-    if (data) dispatch(setClassification({ category, subcategory }));
-    history.push('/');
+    if (data && data.is_active) dispatch(setClassification({ category, subcategory }));
+    if (data || error) history.push('/');
```
janjaap, line 28: I would think that, since the combination of `category` and `subcategory` is not a valid one at this point in the code, it would be best to replace the history entry instead of pushing another one onto the stack. There is no point in going back one page anyway.

src/signals/incident/components/IncidentClassification/index.test.js
```diff
@@ -6,62 +6,113 @@ import configuration from 'shared/services/configuration/configuration';
...
   ...jest.requireActual('react-router-dom'),
 }));
 
-describe('signals/incident/components/IncidentClassification', () => {
-  const dispatch = jest.fn();
+const dispatch = jest.fn();
+const testUnhappyWorkflow = async setFetchResult => {
+  const category = 'foo';
+  const subcategory = 'bar';
+  jest.spyOn(reactRouterDom, 'useParams').mockImplementation(() => ({ category: 'foo', subcategory: 'bar' }));
```
janjaap, line 22: Why not 
```jsx
jest.spyOn(reactRouterDom, 'useParams').mockImplementation(() => ({ category, subcategory }));
```

janjaap (changes requested):

src/signals/incident/components/IncidentClassification/index.test.js
aditudorache: of course!

src/signals/incident/components/IncidentClassification/index.js
aditudorache: agree, I used replace on history here now.

src/signals/incident/components/IncidentClassification/index.js
aditudorache: This is how the PO has requested, will update the AC

janjaap (approved):

PR #1147, opened 2020-09-22
[SIG-3092] Bugfix preset the incident classification 

aditudorache:
This PR fixes an issue that was found in acceptance. The problem was that the category was not preset and the prediction tool was not disabled when the correct presset link was used. In production mode, initialization of the incident container reducer was done after the InidentClassification events were triggered. The solution was to move the IncidentClassification component to the IncidentContainer in order to ensure the correct handling of the redux events.

### DIFF
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -22,7 +22,6 @@ import KtoContainer from 'signals/incident/containers/KtoContainer';
 import useLocationReferrer from 'hooks/useLocationReferrer';
 import useIsFrontOffice from 'hooks/useIsFrontOffice';
 
-import IncidentClassification from 'signals/incident/components/IncidentClassification';
 import { getSources } from './actions';
 import AppContext from './context';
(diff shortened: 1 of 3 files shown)


src/containers/App/index.js
```diff
@@ -103,7 +102,7 @@ export const AppContainer = () => {
               <Route path="/instellingen" component={SettingsModule} />
               <Route path="/incident" component={IncidentContainer} />
               <Route path="/kto/:satisfactionIndication/:uuid" component={KtoContainer} />
-              <Route path="/categorie/:category/:subcategory" component={IncidentClassification} />
+              <Route path="/categorie/:category/:subcategory" component={IncidentContainer} />
```
janjaap, line 105: Might also be better to make this route an exact match
```jsx
<Route exact path="/categorie/:category/:subcategory" component={IncidentContainer} />
```

janjaap (approved):

PR #1150, opened 2020-09-22
[SIG-3156] Fetch subcategories at App level

janjaap:
This PR moves the fetch of subcategories to the `App` container level. Previously, subcategories were retrieved in the `incident-management` and `settings` modules. Since subcategories are also required in the `incident` modules, it makes sense to move the responsibility of retrieving the data to the highest level in the application.

### DIFF
--- a/src/containers/App/index.js
+++ b/src/containers/App/index.js
@@ -1,17 +1,17 @@
 import React, { Fragment, useEffect } from 'react';
 import styled from 'styled-components';
 import { Switch, Route, Redirect, useHistory } from 'react-router-dom';
-import { compose } from 'redux';
 import { useDispatch, useSelector } from 'react-redux';
 
 import configuration from 'shared/services/configuration/configuration';
 import { authenticate, isAuthenticated } from 'shared/services/auth/auth';
-import ThemeProvider from 'components/ThemeProvider';
-import injectSaga from 'utils/injectSaga';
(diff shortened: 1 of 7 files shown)


src/containers/App/index.js
```diff
@@ -55,6 +55,9 @@ export const AppContainer = () => {
   const isFrontOffice = useIsFrontOffice();
   const headerIsTall = isFrontOffice && !isAuthenticated();
 
+  useInjectSaga({ key: 'global', saga });
+  useInjectReducer({ key: 'global', reducer });
```
aditudorache, line 59: Sinds this is a global model definition now, woudn't be beter to place this in the `src/models/index.js` where the rest of global models are injected?

aditudorache (changes requested):

src/containers/App/index.js
janjaap: I'm not sure what you mean. The `categories` model is already instantiated in `app.js`. The `App` container has its own reducer and saga that need to be injected.

boris-arkenaar (approved):

src/containers/App/index.js
aditudorache: What I mean is that the `loadModules` function that is defined in `src/models/index.js` is responsible for loading the redux models and it will be good to have the `global` model at this place loaded for consistency.

aditudorache (approved):

PR #1153, opened 2020-09-24
[SIG-3159] Fixes the autorization check in the backoffice

aditudorache:
This PR, fixes an autorization check in the backoffice.

### DIFF
--- a/src/containers/App/services.test.js
+++ b/src/containers/App/services.test.js
@@ -1,0 +1,40 @@
+import injectReducerModel from 'utils/injectReducerModel';
+import injectSagaModel from 'utils/injectSagaModel';
+
+import reducer from './reducer';
+import saga from './saga';
+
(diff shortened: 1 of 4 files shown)


src/containers/App/services.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+jest.mock('utils/injectReducerModel');
+jest.mock('utils/injectSagaModel');
+
+jest.mock('./reducer');
+jest.mock('./saga');
+
+describe('containers/App/services', () => {
+  const store = { foo: 'bar' };
+  let spy;
```
jpoppe, line 17: Could you make `spy`  a bit more descriptive?

jpoppe (changes requested):

jpoppe (approved):

PR #1158, opened 2020-09-28
[SIG-3115] Runtime chunk optimization

janjaap:
# Runtime chunk optimization

The current Webpack config, albeit a lot more optimized than a year ago, transpiles chunks that are too big for streamlined consumption by older devices or on narrower bandwidths.

Prior to the changes mentioned below, the total production bundle (javascript) size was ~ 3.2 MB, the main chunk was ~2.8 MB and each page loaded approximately 3 MB in javascript.

In order to further improve performance, this PR:
1. References `date-fns` locales specifically instead of using named exports
2. Removes `cacheGroups` from the Webpack config so that Webpack can figure out for itself which chunks it needs to create
3. Uses `lazy`/`Suspense` to lazy-load main parts of the application; forces Webpack to create a lot more chunks

### Referencing `date-fns` locales specifically

Instead of using
```javascript
import { nl } from 'date-fns/locale';
```

the PR imports the correct locale, like so:
```javascript
import nl from 'date-fns/locale/nl';
```

This reduces the production bundle size to ~2.4 MB, a reduction of roughly 800 KB. All in unused locale data.

### Removing `cacheGroups`

In itself, removing the `cacheGroups` prop from the Webpack configuration doesn't have that much effect. In fact, it slightly increases the bundle size, depending on the order in which Webpack transpiles the separate modules it comes across. The true value of letting Webpack decide which chunks it needs to generate comes when we apply module lazy loading for the main parts of the application.

### Using `lazy`/`Suspense`

The main parts of the application have been extracted into lazy-loaded modules. This allows Webpack to create separate chunks from those modules and to extract code reused across modules in a more efficient manner. This doesn't drastically reduce the total chunk size, but it splits the code in such a way that it can be loaded a lot faster.

Note that lazy-loading modules isn't something that can be tested properly. That is why the import statements have been excluded from coverage generation.

In comparison, the (gzipped) amount of javascript that is loaded, per page:
| Page                   	| Before   	| After     	|
|--------------------------	|--------	|--------	|
| /incident/beschrijf      	| 981 KB 	| 530 KB 	|
| /manage/incidents        	| 981 KB 	| 532 KB 	|
| /instellingen/gebruikers 	| 981 KB 	| 332 KB 	|
| /kto/.../...             	| 1 MB   	| 336 KB 	|


aditudorache (dismissed):

jpoppe (approved):
Great work 👍🏻

boris-arkenaar (approved):
Looks good. Impressive results!

PR #1161, opened 2020-10-05
[SIG-3178] fix show extra questions

aditudorache:
This PR fixes a problem that prevented the extra questions to be shown. For this, the category_slug is added to the subcategory list and the classification is saved as an object in the incident


src/signals/incident/containers/IncidentContainer/selectors.js
```diff
@@ -1,6 +1,17 @@
 import { createSelector } from 'reselect';
 import { initialState } from './reducer';
 
+export const getClassificationData = (category, subcategory, { _links, name, slug, handling_message }) => ({
```
janjaap, line 4: With more than two parameters, it would be better to switch to an options object as a single parameter where the keys would then be `category`, `subcategory` and `subcategoryData` (or something similar). Also, add a test.

src/signals/incident/containers/IncidentContainer/saga.js
```diff
@@ -39,7 +33,7 @@ export function* getClassification(action) {
       `${configuration.CATEGORIES_ENDPOINT}${category}/sub_categories/${subcategory}`
     );
 
-    yield put(getClassificationSuccess(getClassificationData(classification)));
+    yield put(getClassificationSuccess(getClassificationData(category, subcategory, classification)));
```
janjaap, line 36: Why is it necessary to call `getClassificationData` here when the payload of the action already contains the return value of `getClassificationData`? As far as I can see, only the value of the `classification` prop from the   `getClassificationData` result is used by the reducer.

aditudorache: All the fields returned are used in the reducer. the results of the `getIncidentClassification` will update all the fields from `getClassificationData` in the `incident`.

src/signals/incident/containers/IncidentContainer/selectors.js
aditudorache: Sure, good point!

src/signals/incident/containers/IncidentContainer/saga.js
janjaap: Not true, only `classification` is used by `action.type` `GET_CLASSIFICATION_SUCCESS`

janjaap (approved):

PR #1163, opened 2020-10-06
[SIG-2972] Add directing department to the splitsing

aditudorache:
This PR updates the directing department in the parent incident after the child incidents are created. Part of the implementations is moving the fetching of the departments list to the main App component as now is used in all the backoffice containers.


src/containers/App/index.js
```diff
@@ -77,6 +78,7 @@ export const AppContainer = () => {
     if (!isAuthenticated()) return;
 
     dispatch(fetchCategoriesAction());
+    dispatch(fetchDepartmentsAction());
```
janjaap, line 81: Add a test?

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -26,32 +27,78 @@ const getParentIncident = incident => ({
...
     () => subcategories?.map(category => ({ ...category, value: category.extendedName })),
     [subcategories]
   );
 
+  const updateDepartment = useCallback(
+    name => {
+      if (!name || !departments) return;
+
+      const department = departments.list.find(d => d.code === name);
+      setDirectingDepartment(department ? [{ id: department.id }] : []);
```
janjaap, line 57: The `find` filter function could potentially return `undefined` which would result in an error on this line. Shouldn't that case be handled?

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -26,32 +27,78 @@ const getParentIncident = incident => ({
...
+
   useEffect(() => {
-    get(`${configuration.INCIDENT_PRIVATE_ENDPOINT}${id}`);
-  }, [get, id]);
+    getParent(`${configuration.INCIDENT_PRIVATE_ENDPOINT}${id}`);
+  }, [getParent, id]);
 
   useEffect(() => {
-    if (isSuccess === undefined || error === undefined) return;
+    if (isLoadingParent || isSuccessParent || errorParent === undefined) return;
```
janjaap, line 67: This evaluates `isSuccessParent` as a truthy value whereas the previous line checked for a falsy value. Is this condition formulated correctly?

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -26,32 +27,78 @@ const getParentIncident = incident => ({
...
+        showGlobalNotification({
+          title: 'De melding kon niet gesplitst worden',
+          variant: VARIANT_ERROR,
+          type: TYPE_LOCAL,
+        })
+      );
+      history.push(`${INCIDENT_URL}/${id}`);
+    }
+
+    // Disabling linter; patch and history dependencies are generating infinite loops
```
janjaap, line 89: This shouldn't be the case, certainly not for `patch`, because that is a memoized function in the `useFetch` hook. Which dependency causes the loop?

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -26,32 +27,78 @@ const getParentIncident = incident => ({
...
+  useEffect(() => {
+    if (isSuccessUpdate === undefined || errorUpdate === undefined) return;
+    const notificationProps = isSuccessUpdate
       ? {
         title: 'De melding is succesvol gesplitst',
         variant: VARIANT_SUCCESS,
       }
       : {
-        title: 'De melding kon niet gesplitst worden',
+        title: 'De melding is gesplits maar het bijwerken van de hoofdmelding regie is niet gelukt',
```
janjaap, line 101: This is a bit of a shaky sentence. Apart from the typos, wouldn't it be better to just mention the fact that setting the directing department wasn't successful? The user is directed back to the incident page and it will be visible that the split was successful by the number of child incidents shown there. The error notification should maybe also mention a way to fix the problem.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -26,32 +27,78 @@ const getParentIncident = incident => ({
...
+      patch(`${configuration.INCIDENT_PRIVATE_ENDPOINT}${id}`, { directing_department: directingDepartment });
+    } else {
+      dispatch(
+        showGlobalNotification({
+          title: 'De melding kon niet gesplitst worden',
+          variant: VARIANT_ERROR,
+          type: TYPE_LOCAL,
+        })
+      );
+      history.push(`${INCIDENT_URL}/${id}`);
```
janjaap, line 86: This part doesn't seem to be tested

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
aditudorache: both history and patch

src/containers/App/index.js
aditudorache: Done!

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
aditudorache: Yes.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
aditudorache: when the `department` is not found (undefined), the `department.id` won't be evaluated but the `[]` from the `else` branch

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
aditudorache: After the discussion with the PO this logic is eliminated for the time being.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
aditudorache: Now it is!

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
janjaap: Might be necessary to change the `useFetch` hook so that the returns functions are memoized:
```javascript
const post = useMemo(() => modify('POST'), [modify]);
const patch = useMemo(() => modify('PATCH'), [modify]);
const put = useMemo(() => modify('PUT'), [modify]);
```

aditudorache: I'll look into that

aditudorache: It helps, but I still have a problem with the history. It's generating also the infinite loop. I could memoize that but it doesn't fill right...

janjaap: This might be related: https://github.com/ReactTraining/react-router/issues/7000

aditudorache: It looks like, indeed. I'll eliminate the `history` the dependency for now.

jpoppe (approved):

PR #1167, opened 2020-10-07
[SIG-3091] Category select width

aditudorache:
This PR styles the Category select width to fit the content of the options list


jpoppe:
@aditudorache I think this can be closed, I took a simpler approach which is now on `develop`.

aditudorache:
Closing this, another solution is implemented

PR #1177, opened 2020-10-12
Fix cat and subcat sent to questions endpoint

boris-arkenaar:
Due to a previous change, the parameters `category` and `subcategory` weren't sent to the `questions` endpoint anymore. This change fixes that bug.

Unit test has been updated to catch this in the future.


aditudorache (approved):
Good catch, I missed that!

PR #1181, opened 2020-10-16
fixes the condition for setting the parent incident

aditudorache:
This PR correct the conditions for setting the parentIncident in the IncidentSplitContainer

### DIFF
--- a/src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
+++ b/src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
@@ -62,13 +62,13 @@ const IncidentSplitContainer = ({ FormComponent }) => {
   }, [getParent, id]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -62,13 +62,13 @@ const IncidentSplitContainer = ({ FormComponent }) => {
   }, [getParent, id]);
 
   useEffect(() => {
-    if (isLoadingParent || isSuccessParent || errorParent === undefined) return;
+    if (errorParent === undefined && dataParent === undefined) return;
```
boris-arkenaar, line 65: I'm not a fan of all this `undefined` stuff and distinguishing between different types (`undefined`, `false`). We should do something about the `useFetch` implementation.

boris-arkenaar (approved):

PR #1187, opened 2020-10-20
Replace fixed column widths with nowrap

boris-arkenaar:
Removed fixed-width columns for the incident list in the back office. The accompaning comment indicated it was done to make sure icon and text would not be wrapped.

Replaced it with the much simpler `white-space: nowrap` which also achieves this goal.
Setting explicit widths is very fragile and decreases maintainability and refactorability significantly!

Also made it possible to explicitly set a cell to nowrap. This is useful for the date column in this case.

A unit test has been provided.

### DIFF
--- a/src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.js
+++ b/src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.js
@@ -61,31 +61,19 @@ const Table = styled.table`
 const Th = styled.th`
   cursor: pointer;
   font-weight: normal;
+  white-space: nowrap;
 
   &:hover {
     text-decoration: underline;
   }
+`;
 
-  ${props => {
-    // making sure that both text and icon are placed on one line by setting explicit min-width values
-    switch (props['data-testid']) {
-      case 'sortId':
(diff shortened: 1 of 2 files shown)


jpoppe (dismissed):
Nice!

aditudorache:
This is approved, just waiting for the design.

boris-arkenaar:
As discussed with the designer; only the `Stadsdeel` column will have a fixed `min-width` of `120px`. This to make sure that `Nieuw-West` doesn't wrap, while there is the freedom for `Het Amsterdamse Bos` to be wrapped.
Without a minimum width `Nieuw-West` would sometimes wrap, which isn't desirable, while having a `white-space: nowrap` wouldn't allow `Het Amsterdamse Bos` to wrap, which isn't desirable either.

While not the most elegant solution, this still removes most of the fixed minimum widths that were there. Furthermore, this will not hinder any other municipality, having other values. This is because all other municipalities are using `districts` instead of `stadsdelen` which is a different column in the table, and thus doesn't have the applied minimum width.

aditudorache (approved):

PR #1188, opened 2020-10-20
[SIG-3211] Enable typescript

aditudorache:
This PR 
- enables typescript development along with the js code
- converts the App component to typescript
- fixes the errors in the unittests
- changes the threshold so the unittests are accepted as is.


src/app.tsx
```diff
@@ -42,7 +42,7 @@ if (dsn) {
 // Create redux store with history
 const initialState = Immutable.Map();
 const store = configureStore(initialState, history);
-const MOUNT_NODE = document.getElementById('app');
+const MOUNT_NODE = document.getElementById('app')!;
```
boris-arkenaar, line 45: What does that exclamation mark do?

src/containers/App/saga.test.ts
```diff
@@ -167,7 +171,8 @@ describe('containers/App/saga', () => {
     });
 
     afterEach(() => {
-      global.location.reload.mockRestore();
+      // eslint-disable-next-line no-extra-parens
```
boris-arkenaar, line 174: These parens don't seem to be 'extra'. Could we tweak the linting rules, so we don't need to disable this rule in such cases?

src/signals/incident-management/containers/FilterTagList/__tests__/FilterTagList.test.js
```diff
@@ -164,6 +164,8 @@ describe('signals/incident-management/containers/FilterTagList', () => {
     it('works with feature flag fetchDistrictsFromBackend enabled', () => {
       configuration.fetchDistrictsFromBackend = true;
 
+      // stadsdeel is intentionaly removed from the tags list here
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
boris-arkenaar, line 168: It can be written as `const { , ...otherTags } = tags;` then, can't it?

boris-arkenaar (commented):
Nice work!

src/app.tsx
aditudorache: ensures the result is treated as not `null` or `undifined`

src/containers/App/saga.test.ts
aditudorache: Agree, is just for now. I will ask @jpoppe to help with this.

src/signals/incident-management/containers/FilterTagList/__tests__/FilterTagList.test.js
aditudorache: I tried that but the linter doesn't like it. Let's solve this later.

boris-arkenaar (approved):

PR #1191, opened 2020-10-21
[SIG-2880] Parent icon in the incidents overview

aditudorache:
This PR extends the Incident overview list with a colomn that marks the parent incidents with an icon.
- To properly test, the incidents fixture list is extended with an extra element.


src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
```diff
@@ -153,7 +154,19 @@ describe('List', () => {
...
 
-      expect(elementsWithTextContent).toHaveLength(2);
+      expect(elementsWithTextContent).toHaveLength(3);
+    });
+
+    it('should render an icon for each parent incident', () => {
+      const { container, rerender } = render(withContext(<List {...props} />));
+      const incidentList = [...props.incidents];
+
+      const parentsCount = incidentList.filter(incident => incident.has_children).length;
```
boris-arkenaar, line 164: I find this way too much code/logic for a unit test, making the test very error prone. I'd prefer a hard coded value for the count.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
```diff
@@ -153,7 +154,19 @@ describe('List', () => {
...
+      expect(container.querySelectorAll('span > svg').length).toEqual(parentsCount);
```
boris-arkenaar, line 165: Can't we use a 'testing-library' selector? Something by role maybe.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
```diff
@@ -153,7 +154,19 @@ describe('List', () => {
...
+
+      const incidentParents = incidentList.map(parent => ({ ...parent, has_children: true }));
```
boris-arkenaar, line 167: Again quite some logic. And, is this really necessary to test this case?

src/utils/__tests__/fixtures/incidents.json
```diff
@@ -135,5 +135,102 @@
...
+    "notes": [],
+    "directing_departments": [
+      {
+        "id": 6,
+        "code": "ASC",
+        "name": "Actie Service Centrum",
+        "is_intern": true
+      }
+    ],
+    "has_children": true
```
boris-arkenaar, line 234: The `has_children` is the only addition, right? I'd just add that property to some of the already existing incident fixtuers.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.js
```diff
@@ -88,6 +75,37 @@ const Th = styled.th`
   }}
 `;
 
+const TdStyle = styled.td`
```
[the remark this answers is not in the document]
aditudorache: The convention as used in ASC as wel is: when creating a new style we use `<HtmlElement>Style`, when we override a Styled component we use `Styled<Component>`

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
aditudorache: I prefer no fixed values as then the tests are to be changed when there are some changing in the stucture/values of the test data.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
aditudorache: removed

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
boris-arkenaar: Because of the code changes you just made above, now you can see directly how many parents there are; exactly one, right? Then this is overkill.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
boris-arkenaar: Nice one!

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.js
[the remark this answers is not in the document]
jpoppe: @boris-arkenaar I agree, we don't have an official convention yet, lets's work on that and refactor after.

jpoppe (approved):

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.test.js
aditudorache: This construction allows me not to be concerned about the contents of the data but if the application does what it should. In this case will test that we render one image for each parent no matter how many parents there are.

jpoppe (approved):

PR #1201, opened 2020-10-26
Validate base config on pre commit hook

boris-arkenaar:
The pre-commit hook validated the configuration in `app.base.json`, including any extended configuration in `app.json`. It only needs to validate `app.base.json` however, since `app.json` won't be committed as it's ignored.

This makes it easier to keep some extended configuration in `app.json` in development, while switching branches.

`validate-base-config.js` in this PR is a copy of `validate-config.js` with the only difference that it does `const { baseConfig } = require('./helpers/config');` instead of `const { config } = require('./helpers/config');`.

### DIFF
--- a/internals/scripts/validate-base-config.js
+++ b/internals/scripts/validate-base-config.js
@@ -1,0 +1,14 @@
+const { validator } = require('@exodus/schemasafe');
+
+const { baseConfig } = require('./helpers/config');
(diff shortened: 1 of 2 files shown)


jpoppe:
Nice, approved ;)

jpoppe (approved):

PR #1208, opened 2020-10-27
(Chore) ADR for testing library conventions

boris-arkenaar:


jpoppe:
🚀 Nice ❤️, should we also put a conventions for consistent fixture import naming in this ADR?

boris-arkenaar:
@jpoppe Not sure what you mean exactly. You could work that out in a new ADR.

aditudorache (approved):

jpoppe:
> @jpoppe Not sure what you mean exactly. You could work that out in a new ADR.



> @jpoppe Not sure what you mean exactly. You could work that out in a new ADR.

It could also be addressed in a new ADR, so I will come back to this later.

jpoppe (approved):

PR #1209, opened 2020-10-27
(Chore) unit test conventions

boris-arkenaar:
According to the new testing library conventions.

- Remove `cleanup` statements.
- Remove unnecessary `waitFor` statements.
- Add the `@testing-library/user-event` dependency.
- Change `fireEvent` statements in a few files to `userEvent` statements.
- Remove unnecessary `act` statements in a few files.


src/signals/settings/categories/Detail/__tests__/Detail.test.js
```diff
@@ -59,11 +59,11 @@ describe('signals/settings/categories/Detail', () => {
   });
 
   it('should render a backlink', async () => {
-    const { container } = render(withAppContext(<CategoryDetailContainer />));
+    const { findByTestId } = render(withAppContext(<CategoryDetailContainer />));
 
-    await waitFor(() => container.querySelector('a'));
+    const backLink = await findByTestId('backlink');
```
MauritsR, line 64: Given the conventions 'Using the wrong query' wouldn't the following be preferred?
`const backLink = screen.getByRole('link', {name: 'Terug naar overzicht'});`

jpoppe: Great comment, I leave it to @boris-arkenaar to implement this now, or later ;)

jpoppe (approved):

src/signals/settings/categories/Detail/__tests__/Detail.test.js
boris-arkenaar: As far as I know, there is no `link` role. You are right in that there should be a better alternative than `findByTestId`. However, I find that to be out of scope of this PR since it would require more in-depth changes.

MauritsR: Agree that it's out of scope 👍 

> As far as I know, there is no link role
`link` is the [implicit role](https://www.w3.org/TR/html-aria/#docconformance) of the <a href=""/> tag

boris-arkenaar: Thanks! I was looking at an incomplete list here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

PR #1212, opened 2020-10-27
(Chore) Improve directing department selection

aditudorache:
This PR
- Moves the selector for the directing departments to a separate hook
- Uses the departments can_direct prop to determine if a department should be included in the directing departments list.


MauritsR (approved):
Looks good 😄

PR #1216, opened 2020-10-28
Fix PDOK suggest for multiple municipalities

boris-arkenaar:
closes Signalen/backend#73

With multiple municipalities configured in `map.municipality` in `app.json`, the results of the PDOK suggest service was unpredictable. Some cities within a municipality configured would not show up in the suggest results. While other municipalities not configured would show up in the suggest results.

It turns out that calling the PDOK suggest service with the following parameter is not giving the expected results:

```
fq=gemeentenaam:"boxtel" "altena"
```

When we write this parameter as follows, we do get the expected results:

```
fq=gemeentenaam:"boxtel"gemeentenaam:"altena"
```

This should be updated accordingly in the backend as well.

### Unexpected results

The following request will give you the expected results of the `Mezenlaan` in `Boxtel`.

```
➜ curl https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest\?fq\=gemeentenaam:%22boxtel%22%20%22altena%22\&fq\=bron:BAG\&fq\=type:adres\&fl\=id,weergavenaam,straatnaam,postcode,woonplaatsnaam,gemeentenaam\&q\=mezen
```

The following request, however, will give you no results, while you would expect results of the `Jettebosstraat` in `Woudrichem` which falls under the municipality of `Altena`.

```
➜ curl https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest\?fq\=gemeentenaam:%22boxtel%22%20%22altena%22\&fq\=bron:BAG\&fq\=type:adres\&fl\=id,weergavenaam,straatnaam,postcode,woonplaatsnaam,gemeentenaam\&q\=jett
```

### Expected results

The following request will still give you the expected results of the `Mezenlaan` in `Boxtel`.

```
➜ curl https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest\?fq\=gemeentenaam:%22boxtel%22gemeentenaam:%22altena%22\&fq\=bron:BAG\&fq\=type:adres\&fl\=id,weergavenaam,straatnaam,postcode,woonplaatsnaam,gemeentenaam\&q\=mezen
```

The following request will now give you the expected results of the `Jeetebosstraat` in `Woudrichem` which falls under the municipality of `Altena`.
```
➜ curl https://geodata.nationaalgeoregister.nl/locatieserver/v3/suggest\?fq\=gemeentenaam:%22boxtel%22gemeentenaam:%22altena%22\&fq\=bron:BAG\&fq\=type:adres\&fl\=id,weergavenaam,straatnaam,postcode,woonplaatsnaam,gemeentenaam\&q\=jett
```

### DIFF
--- a/src/components/PDOKAutoSuggest/__tests__/PDOKAutoSuggest.test.js
+++ b/src/components/PDOKAutoSuggest/__tests__/PDOKAutoSuggest.test.js
@@ -4,15 +4,15 @@ import { withAppContext } from 'test/utils';
 
 import JSONResponse from 'utils/__tests__/fixtures/PDOKResponseData.json';
 import { INPUT_DELAY } from 'components/AutoSuggest';
-import { formatPDOKResponse } from 'shared/services/map-location';
+import { formatPDOKResponse, pdokResponseFieldList } from 'shared/services/map-location';
 import PDOKAutoSuggest, { formatResponseFunc } from '..';
 
 const mockResponse = JSON.stringify(JSONResponse);
 
 const onSelect = jest.fn();
-const municipalityQs = 'fq=gemeentenaam:';
+const municipalityName = 'gemeentenaam:';
+const municipalityQs = 'fq=';
 const fieldListQs = 'fl=';
-const defaultFieldsQs = 'id,weergavenaam';
 
 const renderAndSearch = async (value = 'Dam', props = {}) => {
   const result = render(withAppContext(<PDOKAutoSuggest onSelect={onSelect} {...props} />));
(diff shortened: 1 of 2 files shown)


aditudorache:
We don't have this problem with multiple municipalities. We are using in one of the configurations the array construction:

```
  "map": {
    "municipality": [
      "amsterdam",
      "amstelveen"
    ],
    "options": {
      "maxBounds": [
...
```

Did you tried to configure it in this way?

boris-arkenaar:
@aditudorache That's exactly how we have it configured as well! The problem is not the configuration, but how the request is made to PDOK. The fact that the results are a bit unpredictable doesn't mean that it'll go wrong in every situation. So it might work for those municipalities. The new logic however, gives a more stable result.

boris-arkenaar:
I think the problem arises when you have one `gemeentenaam` with different `woonplaatsnamen`. This is not the case for `Amsterdam` en `Amstelveen`.

aditudorache:
Does this bugfix affect our existing configurations?

boris-arkenaar:
It changes how the request to the PDOK suggest service is done. Functionally nothing changes. Your configuration is still correct and does not need to be changed.

aditudorache (approved):

PR #1221, opened 2020-10-30
[SIG-3283] Hides split button

MauritsR:
This PR prevents the 'delen' button to be shown. 

Background info:
The incident split functionality was developed and accepted, but the customer is not yet ready to make use of it in the coming release. This commit will be reverted when the customer is ready. 

Concerning the e2e tests, @arjandata is aware of this change, and the periodic e2e tests will succeed because the deelmeldingen tests are not (yet) enabled.


aditudorache (approved):
Can be merged as soon as the release preparations start!

PR #1222, opened 2020-10-30
[SIG-3233] Sharable ESLint configuration

jpoppe:
Todo before merge: run fix on a few comma dangles, check GitHub integration

* Lint rules have been separated and moved to it's own plugin
* Typescript ready :)
* Cleaned and optimized configuration
* Reimplemented and fixed 'broken' lint-staged plugin
* Added Jest plugin
* Added Node plugin
* Added find rules script
* Linting now also checks additonal scripts
* Removed clashing prettier/airbnb plugins (packages will be removed later) 
* Made clear separation and proposed rules really clear, needs to be picked up as chores? Out of scope for now
* Plugin could be shared with Amstedam organization but needs to be discussed
* Much more :)


package.json
```diff
@@ -6,7 +6,7 @@
   },
   "engines": {
     "npm": ">=6",
-    "node": "12"
+    "node": "15"
```
aditudorache, line 9: Is this version of node necessary for eslint? I think it might conflict with gihub actions workflow where the used version is 12.8

.eslintrc.yaml
```diff
@@ -0,0 +1,131 @@
+root: true
```
aditudorache, line 1: I noticed that this file is `yaml` while some others configuration files are `js`. Shouldn't we keep one file type for consistency?

package.json
aditudorache, line 9: Is this version of node necessary? It doesn't match the one used in the github actions (12.8) can we sync that?

aditudorache (commented):
LGTM; I have a few questions and I saw there are few `comma-dangle` errors when running the linter.

aditudorache:
Can we target this to the `post-release` branch before merging?

boris-arkenaar (commented):
So many rules 😱 
Can't we just stick to an industry-standard as much as possible? Like `airbnb`.

package.json
jpoppe: No this should not be there, nice catch! thanks!

jpoppe:
> So many rules 
> Can't we just stick to an industry-standard as much as possible? Like `airbnb`.

No, you will miss about 70% of the rules (many of those rules prevent discussions here), `airbnb` was where the project started.

.eslintrc.yaml
jpoppe: The YAML files is the application configuration and should be as good as empty in the ideal world, the js files are part of the `eslint-plugin-amsterdam` which could eventually live somewhere out of this repository.

aditudorache: Not sure I understand the explanation, why don't we keep the same format for eslint configuration? It doesn't really matter which format as long is consistent overall. I would choose js because most of the projects at `datapunt` are using this format. As a matter of fact, this is the original format before this PR.

eslint-plugin-amsterdam/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "eslint-plugin-amsterdam",
+  "description": "eslint plugin shared by Amsterdam",
+  "version": "0.0.1",
+  "author": "Jasper Poppe",
```
aditudorache, line 5: Can we use `Datapunt Amsterdam` here? maintainers names can be put in a MAINTAINERS.md file if needed

boris-arkenaar:
> > So many rules
> > Can't we just stick to an industry-standard as much as possible? Like `airbnb`.
> 
> No, you will miss about 70% of the rules (many of those rules prevent discussions here), the project started with `airbnb` rules. which were not sufficient and left to much room for discussion on GitHub reviews.

Got it!

eslint-plugin-amsterdam/package.json
jpoppe: I removed the author field for now, we could add add Datapunt or something else if this plugin will ever be published.

MauritsR (approved):
Looks good :)

.eslintrc.yaml
boris-arkenaar: I'm also still interested to hear your reasoning for choosing the yaml format. :)

aditudorache: @boris-arkenaar, mee too. @jpoppe, we agreed to talk about this PR next week. I think this might be merged by mistake. Could you please revert this?

PR #1223, opened 2020-10-30
[FE#87] Use correct logo height on mobile

boris-arkenaar:
closes Signalen/frontend#87

On smaller screens, non-Amsterdam logo's wouldn't be shown in their smaller version and thus not fit nicely.

image

This PR fixes that.

@jpoppe I removed the `max-height` on the logo. I see you've added this in the past. I'm not sure for what reason. It messes with the fixed height from the configuration. Let me know if this is okay.

### DIFF
--- a/internals/schemas/app.schema.json
+++ b/internals/schemas/app.schema.json
@@ -273,7 +273,7 @@
           ]
         },
         "height": {
-          "description": "Length value that sets the logo image's height",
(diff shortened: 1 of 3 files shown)


jpoppe:
@boris-arkenaar I am sure that max-height needs to be there for our setup, we need it for our `weesp logo`, I'll ping you next week to discuss 'offline'

jpoppe:
@boris-arkenaar Herjen will pick this up with you.

PR #1224, opened 2020-11-03
[SIG-3127] Read sources from the backend

aditudorache:
This PR
- Enables using the `source` data from the API
- Removes the  `fetchSourcesFromBackend` as is no longer necessary
- Fixes a bug by using the sources name instead of if in the filter requests
- Removes the static definition of the sources


src/signals/incident-management/__tests__/selectors.test.js
```diff
@@ -314,17 +290,17 @@ describe('signals/incident-management/selectors', () => {
...
 
       expect(actual.id).toEqual(sourceFilter.id);
       expect(actual.options).toEqual({
-        source: [],
+        source: [sources[0]],
       });
     });
 
-    it('should work without sources with fetchSourcesFromBackend disabled', () => {
+    it('should not work without sources', () => {
```
MauritsR, line 303: This test is a duplicate of the one on line 313 (can be removed)

MauritsR (changes requested):
Looks good (only a minor change necessary)

MauritsR (approved):

jpoppe (approved):

PR #1226, opened 2020-11-03
[SIG-3269] Refresh the incident children after each incident change

aditudorache:
This PR fixes a problem that occured after the parent incident was closed or canceled; The open child incidents are automatically canceled, but this change was not direct visible in the detail page.


MauritsR (approved):
👍

PR #1229, opened 2020-11-03
(Chore) Renames the SelectInput base component to Select

aditudorache:
This PR:
- Renames the base component `SelectInput` to `Select` in 'components/Select'
- This base component is the implementation of the `Select` component from `amsterdam-styled-components` package
- This component name is now consistent used throughout the codebase.
- This change prepares the implementation of the category groups in the select component (SIG-3029)
- The existing functionality is not altered


boris-arkenaar (changes requested):
This `SelectInput` `src/signals/incident/components/form/SelectInput/index.js` is used through `expandQuestions` https://github.com/Amsterdam/signals-frontend/blob/develop/src/signals/incident/definitions/wizard-step-2-vulaan.js#L23-L40 and should thus be updated here https://github.com/Amsterdam/signals-frontend/blob/develop/src/signals/incident/containers/IncidentContainer/services.js#L16

boris-arkenaar:
Apart from that, nice work on making this more consistent. Has been bugging me in the past as well 👍

aditudorache (commented):
Not sure is correct. That `SelectInput` referes the `src/signals/incident/components/form/SelectInput` which is an Select that supports the reactive-form interface.

boris-arkenaar:
> Not sure that's correct. That `SelectInput` referes the `src/signals/incident/components/form/SelectInput` which is an Select that supports the reactive-form interface. The `Select` that is renamed in this PR is the Styled component implementation of the select.

You are right. So, maybe change that `SelectInput` to `Select` as well? That could be out of scope here, though.

boris-arkenaar (approved):

PR #1232, opened 2020-11-05
[SIG-3029] Group the subcategories by category in the select

aditudorache:
This PR
- Adds `optgroup` functionality to the `Select`
- Adds a new selector that retrieves the categories and subcategories to prepare the subcategory select control
- Uses the `Select` component as base for all 4 occurences of the Subcategory select input
- Builds a consistent list of options and groups that is passed the the Subcategory select input


src/components/Select/Select.js
```diff
@@ -1,30 +1,69 @@
...
+  ({ label, onChange, name, value, options, optionKey = 'key', optionValue = 'name', groups, emptyOption }, ref) => (
+    <AscSelect value={value} onChange={onChange} data-testid={name} label={label} name={name} ref={ref}>
+      {emptyOption && (
+        <option key={`${name}-${emptyOption[optionKey]}`} value={emptyOption[optionKey]}>
+          {emptyOption[optionValue]}
+        </option>
+      )}
+
+      {groups?.length > 1 ?
+        groups?.map(group => (
```
boris-arkenaar, line 23: The `?.` is not necessary here.

src/components/Select/Select.js
```diff
@@ -1,30 +1,69 @@
...
+        : (
+          <SelectOptions name={name} options={options} optionKey={optionKey} optionValue={optionValue} />
+        )}
+    </AscSelect>
+  )
 );
 
+Select.defaultProps = {
+  groups: null,
+  emptyOption: null,
```
boris-arkenaar, line 42: Setting default values to `null` isn't really necessary I imagine.

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.test.js
```diff
@@ -203,17 +194,18 @@ describe('MetaList', () => {
 
     it('should be visible', () => {
       render(renderWithContext());
-
       expect(screen.getByText(subcategoryLabel)).toBeInTheDocument();
-      expect(screen.getByText(parentIncident.category.sub)).toBeInTheDocument();
+      const re = new RegExp(`${parentIncident.category.sub}`, 'g');
```
boris-arkenaar, line 198: Why does this need to be a regular expression?

src/components/Select/Select.js
aditudorache: removed!

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.test.js
aditudorache: The generation of the display values for the categories contains at the end a list of the departments. I have chosen not to test that that was build and keep the unittest changes minimal. I could add that extra test if you think is necessary.

jpoppe (approved):

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.test.js
boris-arkenaar: I don't understand. You can still test it with a string instead of a regular expression, right?

src/components/Select/Select.js
boris-arkenaar: The same goes for `groups` of course.

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.test.js
aditudorache: Of course, I'll use a static string then, thanks!

boris-arkenaar (approved):

PR #1236, opened 2020-11-09
[SIG-3275] Add note field to split form

MauritsR:
This PR adds the note field to the split form. 

Furthermore, this PR fixes two small bugs I noticed while working on this feature:

- [[SIG-3300](https://datapunt.atlassian.net/browse/SIG-3300)] Regie van hoofdmelding moet aangepast kunnen worden
- [[SIG-3299](https://datapunt.atlassian.net/browse/SIG-3299)] Regie voerende afdeling niet wijzigen wanneer deze niet gewijzigd is


src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -50,6 +67,37 @@ const IncidentSplitContainer = ({ FormComponent }) => {
     [departments, setDirectingDepartment]
   );
 
+  const handleCompletedFormSubmit = useCallback(
```
aditudorache, line 70: Long names are difficult to read especially for component internal members. How about `submitCompleted`?

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -30,11 +30,28 @@ const IncidentSplitContainer = ({ FormComponent }) => {
   const dispatch = useDispatch();
   const [parentIncident, setParentIncident] = useState();
   const [directingDepartment, setDirectingDepartment] = useState([]);
+  const [note, setNote] = useState();
   const departments = useSelector(makeSelectDepartments);
   const directingDepartments = useSelector(makeSelectDirectingDepartments);
 
   const [subcategoryGroups, subcategoryOptions] = useSelector(makeSelectSubcategoriesGroupedByCategories);
 
+  const parentIncidentPatchData = useMemo(() => {
```
aditudorache, line 39: I would use a useCallback call and not an useMemo here. What do you think @boris-arkenaar ?
Long names are difficult to read especially for component internal members. In this case `patchData` or `getPatchData` would be enough to understand what's going on.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -30,11 +30,28 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
+    if (!parentIncident) return {};
+
+    const shouldPatchDirectingDepartment =
```
aditudorache, line 42: Long name for a boolean. This construction is difficult to understand. Can we think about something more readable? Maybe destructure the `parentIncident`?

aditudorache (changes requested):
Nice work! Just a few improvements suggested. I have noticed that on the page there is not enough space above the note, could you check with the design?

src/signals/incident-management/containers/IncidentSplitContainer/components/IncidentSplitRadioInput/index.js
```diff
@@ -12,7 +12,6 @@ const IncidentSplitRadioInput = ({ id, name, display, options, initialValue, reg
 
   const onChange = useCallback(
     event => {
-      event.preventDefault();
```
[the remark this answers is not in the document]
MauritsR: This small change fixes https://datapunt.atlassian.net/browse/SIG-3300. Because the event was unnecessarily prevented, react-hook-form would not pick up the change (unless other form elements were changed afterwards).

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
MauritsR: Will rename `parentIncidentPatchData ` to `patchData`, but I'm curious about the benefit of using useCallback instead of useMemo here.

boris-arkenaar: As it is, it is calculating data that is being used elsewhere. That means `useMemo` is the correct way to go.

You could argue that this logic can be executed below, directly where the data is actually being used. Since that's the only place it's being used, I believe. You might then be inclined to extract that logic from there to a separate function, to keep the code clean, and that function would be inside a `useCallback`. So, following that trail of thought, this can go inside a `useCallback` indeed, instead of a `useMemo`.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -30,11 +30,28 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
 
+  const patchData = useMemo(() => {
+    if (!parentIncident) return {};
+
+    const shouldPatchDirectingDepartment =
+      !parentIncident.directing_departments ||
+      parentIncident.directing_departments.length !== directingDepartment.length ||
+      (!parentIncident.directing_departments.includes(department => department?.id === directingDepartment[0]?.id) &&
+        typeof directingDepartment[0]?.id !== 'undefined');
+    const shouldAddNote = !!note?.trim();
```
boris-arkenaar, line 47: Prefer to use the more explicit `Boolean(...)`.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -30,11 +30,28 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
   const [subcategoryGroups, subcategoryOptions] = useSelector(makeSelectSubcategoriesGroupedByCategories);
 
+  const patchData = useMemo(() => {
+    if (!parentIncident) return {};
+
+    const shouldPatchDirectingDepartment =
+      !parentIncident.directing_departments ||
+      parentIncident.directing_departments.length !== directingDepartment.length ||
+      (!parentIncident.directing_departments.includes(department => department?.id === directingDepartment[0]?.id) &&
+        typeof directingDepartment[0]?.id !== 'undefined');
```
boris-arkenaar, line 46: Is this necessary, to check for `typeof` `undefined`?

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -30,11 +30,28 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
 
   const [subcategoryGroups, subcategoryOptions] = useSelector(makeSelectSubcategoriesGroupedByCategories);
 
+  const patchData = useMemo(() => {
+    if (!parentIncident) return {};
+
+    const shouldPatchDirectingDepartment =
+      !parentIncident.directing_departments ||
+      parentIncident.directing_departments.length !== directingDepartment.length ||
+      (!parentIncident.directing_departments.includes(department => department?.id === directingDepartment[0]?.id) &&
```
boris-arkenaar, line 45: `includes` compares the given value. Now it checks if the function you pass as the first argument is included in the array. I'd use `some`.

boris-arkenaar: Can't we assume all departments in `parentIncident.directing_departments` are at least objects? Which would make the `?.` in `department?.id` unnecessary.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
boris-arkenaar: Aside from my comments on the lines below, I find this okay to read.
You cold make it a bit more readable by using some more consts along the way, like this:

```
const directingDepartments = parentIncident.directing_departments;
const differentLength = directingDepartments.length !== directingDepartment.length;
const differentValues = differentLength || !directingDepartments.some(department => department?.id === directingDepartment[0]?.id);
const shouldPatch = !directingDepartments || differentValues;
```

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -65,40 +113,25 @@ const IncidentSplitContainer = ({ FormComponent }) => {
 
   useEffect(() => {
     if (isSuccessSplit === undefined || errorSplit === undefined) return;
+
     if (isSuccessSplit) {
-      patch(`${configuration.INCIDENT_PRIVATE_ENDPOINT}${id}`, { directing_departments: directingDepartment });
+      if (Object.keys(patchData).length > 0) {
```
boris-arkenaar, line 118: You could do this (less `if`s and `else`s):

```
const hasPatchData = Object.keys(patchData).length > 0;
if (isSuccessSplit && hasPatchData) {
  patch(`${configuration.INCIDENT_PRIVATE_ENDPOINT}${id}`, patchData);
} else {
  submitCompleted({ success: isSuccessSplit });
}
```

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -55,6 +55,37 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
+  const handleCompletedFormSubmit = useCallback(
+    /**
+     * @param {Object} params
+     * @param {boolean} params.success
+     */
+    ({ success }) => {
+      if (success) {
+        dispatch(
+          showGlobalNotification({
+            title: 'De melding is succesvol gedeeld',
```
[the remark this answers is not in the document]
MauritsR: The message was already there (I moved it to this useCallback), but I share your opinion - will verify the text

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -30,11 +30,28 @@ const IncidentSplitContainer = ({ FormComponent }) => {
   const dispatch = useDispatch();
   const [parentIncident, setParentIncident] = useState();
   const [directingDepartment, setDirectingDepartment] = useState([]);
+  const [note, setNote] = useState();
   const departments = useSelector(makeSelectDepartments);
   const directingDepartments = useSelector(makeSelectDirectingDepartments);
 
   const [subcategoryGroups, subcategoryOptions] = useSelector(makeSelectSubcategoriesGroupedByCategories);
 
+  const patchData = useMemo(() => {
```
aditudorache, line 39: I believe my comment on this is lost in `translation`. The suggestion was to use here a useCallback because the patchData is only needed once and only if the splitsing has succeded. Returning a null value when nothing changes would make unnecessary the use of `Object.keys().length` construction.

boris-arkenaar: It's here: https://github.com/Amsterdam/signals-frontend/pull/1236#discussion_r519885335

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.test.js
```diff
@@ -261,7 +319,7 @@ describe('signals/incident-management/containers/IncidentSplitContainer', () =>
 
     expect(dispatch).toHaveBeenCalledWith(
       showGlobalNotification({
-        title: 'De melding is succesvol gedeeld',
+        title: 'Deelmelding gemaakt',
```
aditudorache, line 322: Did you discuss this change with the designer? I believe here could be more that one Incident and the existing message was covering that.

src/signals/incident-management/containers/IncidentSplitContainer/components/IncidentSplitForm/index.js
```diff
@@ -47,23 +44,32 @@ const IncidentSplitForm = ({ parentIncident, subcategories, directingDepartments
...
 
-          <IncidentSplitRadioInput
-            display="Regie"
-            register={register}
-            initialValue={parentIncident.directingDepartment}
-            name="department"
-            id="department"
-            data-testid="incidentSplitFormRadioInputDepartment"
-            options={directingDepartments}
+          <StyledBottomWrapper>
```
aditudorache, line 47: I would add this bottom margin as a style on the `IncidentSplitRadioInput`.

aditudorache (changes requested):

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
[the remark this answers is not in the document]
MauritsR: Discussed with designer and changed the text

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
MauritsR: It was not - removed it

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.test.js
MauritsR: Good catch, it does not account for multiple messages. Checked with designer and changed it to _Deelmelding(en) gemaakt_

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -36,19 +36,23 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
     if (!parentIncident) return {};
 
-    const shouldPatchDirectingDepartment =
-      !parentIncident.directing_departments ||
-      parentIncident.directing_departments.length !== directingDepartment.length ||
-      (!parentIncident.directing_departments.includes(department => department?.id === directingDepartment[0]?.id) &&
-        typeof directingDepartment[0]?.id !== 'undefined');
-    const shouldAddNote = !!note?.trim();
+    // Initially, directing_departments can be undefined
+    const directing_departments = parentIncident.directing_departments || [];
```
boris-arkenaar, line 43: Identifiers are camel cased, not snake cased.

boris-arkenaar (approved):

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -50,6 +50,8 @@ const IncidentSplitContainer = ({ FormComponent }) => {
     const shouldPatchDirectingDepartment = differentLength || differentValue;
     const shouldPatchNote = Boolean(note?.trim());
 
+    if (!shouldPatchDirectingDepartment && !shouldPatchNote) return null;
```
boris-arkenaar, line 53: I would rather see you combine this with the `return` statement using a ternary expression.

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
```diff
@@ -80,15 +82,15 @@ const IncidentSplitContainer = ({ FormComponent }) => {
...
+            title: 'Deelmelding(en) gemaakt',
             variant: VARIANT_SUCCESS,
             type: TYPE_LOCAL,
           })
         );
       } else {
         dispatch(
           showGlobalNotification({
-            title: 'De deelmelding kon niet gemaakt worden',
+            title: 'De deelmelding(en) kon niet gemaakt worden',
```
boris-arkenaar, line 93: Hmm, then it should be `kon(den)` as well...

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.test.js
```diff
@@ -345,7 +345,7 @@ describe('signals/incident-management/containers/IncidentSplitContainer', () =>
 
     expect(dispatch).toHaveBeenCalledWith(
       showGlobalNotification({
-        title: 'De deelmelding kon niet gemaakt worden',
+        title: 'De deelmelding(en) kon niet gemaakt worden',
```
boris-arkenaar, line 348: Here too

boris-arkenaar (changes requested):

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.test.js
MauritsR: True - although this occurrence will always reflect the text in the main file

src/signals/incident-management/containers/IncidentSplitContainer/IncidentSplitContainer.js
MauritsR: Yes.. which makes it less readable. I've discussed it with design and will change the text to the singular version (i.e. Deelmelding gemaakt). Even though in practice it can be multiple deelmeldingen, the message is clear and concise.

src/signals/incident-management/containers/IncidentSplitContainer/components/IncidentSplitForm/index.js
MauritsR: I've refactored the implementation for styling this element, please take another look :)

aditudorache (approved):

PR #1238, opened 2020-11-10
[SIG-3029] Uses the extendedName for the subcategory in the category list

aditudorache:
This PR fixes a bug where in the display name of the subcategory the departments list were missing


jpoppe (changes requested):
Is it correct that I miss the test which could prevent the bug next time?

jpoppe (approved):
Thanks :)

MauritsR (approved):

PR #1244, opened 2020-11-13
[SIG-3294] Show split issues processing time

jpoppe:


src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -214,6 +238,8 @@ const IncidentDetail = () => {
     dispatch({ type: CLOSE_ALL });
   }, []);
 
+  if (!Object.keys(handlingTimesBySlug).length) return null;
```
aditudorache, line 241: Is this check necessary?

src/utils/__tests__/fixtures/handlingTimesBySlug.json
```diff
@@ -0,0 +1,122 @@
+{
```
aditudorache, line 1: This information is already in the category fixture and can be by adding a selector in `src/utils/__tests__/fixtures/index.js`. That way is no extra json file needed

aditudorache (changes requested):
Impressive work, I made some suggestions though. I think we can avoid adding the 2 new json fixture files while the info is already present in another fixture and can be easily extracted. This will  save us json file maintenance later ;-). Could you take a look at that please? Thanks!

src/signals/incident-management/containers/IncidentDetail/index.js
jpoppe: With this in place the page will load after handling times are loaded (like it happens on a page refresh)

jpoppe (commented):
.

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
```diff
@@ -140,16 +129,10 @@ const MetaList = () => {
...
-    if (!incident?.category) return undefined;
-
-    const category = subcategoryOptions?.find(option => option.slug === incident.category.sub_slug);
-
-    if (!category) return undefined;
-
-    return getHandlingTime(category.sla.n_days, category.sla.use_calendar_days);
-  }, [incident, subcategoryOptions]);
+  const handlingTime = useMemo(
+    () => !incident?.category ? undefined : handlingTimesBySlug[incident.category.sub_slug],
```
boris-arkenaar, line 133: I would prefer a positive condition (i.e. without the exclamation mark).

src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -62,6 +64,24 @@ const Preview = styled.div`
   z-index: 1;
 `;
 
+export const formatWeekOrWorkdays = (days, isCalendarDays) => {
```
boris-arkenaar, line 67: Shouldn't it be better to put these helper functions in their own service/helper/something file/folder?

src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -62,6 +64,24 @@ const Preview = styled.div`
...
+  const dayString = days === 1 ? 'dag' : 'dagen';
+  return isCalendarDays ? dayString : `werk${dayString}`;
+};
+
+const getDaysString = (days, isCalendarDays) => `${days} ${formatWeekOrWorkdays(days, isCalendarDays)}`;
+
+export const getHandlingTimes = subcategories => {
+  const handlingTimes = {};
+
+  subcategories?.forEach(subcategory => {
```
boris-arkenaar, line 77: You can use reduce here.

```JavaScript
export const getHandlingTimes = subcategories =>
  (subcategories || []).reduce((acc, {slug, sla}) => ({
    ...acc,
    [slug]: getDaysString(sla.n_days, sla.use_calendar_days),
  }), {});
```

src/signals/incident-management/containers/IncidentDetail/index.js
boris-arkenaar:
```JavaScript
return Object.keys(handlingTimesBySlug).length && state.incident ? (
  <IncidentDetailContext.Provider>
    ...
  </IncidentDetailContext.Provider>
) : null;
```

boris-arkenaar: Don't you want to show a loading indicator instead?

src/components/ChildIncidents/index.js
```diff
@@ -102,15 +103,25 @@ const Li = styled(ListItem)`
...
-        <DisplayValue key={key} title={value}>
-          {value}
-        </DisplayValue>
-      ));
+    {incidents.map(({ href, status, values: childIncident, changed }) => {
+      const valueEntries = (
+        <Fragment>
+          <DisplayValue key="id" title={childIncident.id}>{childIncident.id}</DisplayValue>
+
+          <DisplayValue key="status-and-category" title={childIncident.status}>
```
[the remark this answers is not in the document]
MauritsR: Discussed earlier - will leave the current _span_ solution

src/components/ChildIncidents/index.js
```diff
@@ -102,15 +103,25 @@ const Li = styled(ListItem)`
...
+            {childIncident.status}
+            <br />
```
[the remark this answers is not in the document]
MauritsR: Discussed earlier - will leave the current _span_ solution

src/signals/incident-management/containers/IncidentDetail/index.js
aditudorache: With the designer agreed that this slows down showing the page on slower connections so it shoulsd be removed. A new solution that is proposed in another story will solve that issue.

boris-arkenaar (approved):

src/signals/incident-management/containers/IncidentDetail/index.js
```diff
@@ -73,6 +93,10 @@ const IncidentDetail = () => {
   const { get: getDefaultTexts, data: defaultTexts } = useFetch();
   const { get: getChildren, data: children } = useFetch();
 
+  const subcategories = useSelector(makeSelectSubCategories);
+
+  const handlingTimesBySlug = useMemo(() => getHandlingTimes(subcategories), [subcategories]);
```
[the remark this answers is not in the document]
jpoppe: We could address this in another issue, the business really wan't everything to be loaded when the page is rendered, this will require some refactoring in how we handle subcategories on application.

src/signals/incident-management/containers/IncidentDetail/index.js
jpoppe: It should be better if we would get this data from the backend so all this logic does not have to be implemented at all.

I agree and it crossed my mind while putting it here, I  have moved it to a new transform service for now.

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
jpoppe: Sonar auto fixed my initial if statement through ESLint, I agree, there is an ESLint rule for this, and it is in the proposed rules I think, feel free to enable some ;)

src/shared/services/transform/index.js
```diff
@@ -0,0 +1,15 @@
+export const formatWeekOrWorkdays = (days, isCalendarDays) => {
```
boris-arkenaar, line 1: `transform` is a very generic name.
Why not move it to `src/shared/services/date-utils/index.js`?

src/shared/services/transform/index.test.js
```diff
@@ -0,0 +1,17 @@
+import { formatWeekOrWorkdays } from '.';
+
+describe('transform service', () => {
```
boris-arkenaar, line 3: Could you add tests for `getHandlingTimesBySlugFromSubcategories`?

boris-arkenaar (changes requested):

src/components/ChildIncidents/index.js
```diff
@@ -102,15 +103,25 @@ const Li = styled(ListItem)`
 
 const ChildIncidents = ({ className, incidents }) => (
   <List className={className} data-testid="childIncidents">
-    {incidents.map(({ href, status, values, changed }) => {
-      const valueEntries = Object.entries(values).map(([key, value]) => (
-        <DisplayValue key={key} title={value}>
-          {value}
-        </DisplayValue>
-      ));
+    {incidents.map(({ href, status, values: childIncident, changed }) => {
```
MauritsR, line 106: Instead of reassigning _values_ to _childIncident_, it would be cleaner to assign the 'correct' key in the ChildIncidents component directly. Besides that, I think that the 'values' data could be moved to the root.

MauritsR, line 106: Instead of reassigning `values` to `childIncident`, it would be cleaner to assign the 'correct' key in the ChildIncidents component directly. Besides that, I think that the 'values' data could be moved to the root.

src/components/ChildIncidents/index.js
```diff
@@ -102,15 +103,25 @@ const Li = styled(ListItem)`
...
+
+          <DisplayValue key="handling-time" title={childIncident.handlingTime}>
+            {childIncident.handlingTime}
+          </DisplayValue>
+        </Fragment>
+      );
 
       return (
-        <Li key={JSON.stringify(values)} status={status} numValues={values.length} changed={changed}>
+        <Li key={JSON.stringify(childIncident)} status={status} numValues={childIncident.length} changed={changed}>
```
MauritsR, line 124: numValues is not used in Li so it can be removed

MauritsR, line 124: Status is not defined - should be `childIncident.status`

src/components/ChildIncidents/index.js
aditudorache: Good point Maurits! Gets more readable in this way.

src/components/ChildIncidents/index.js
jpoppe: Nope, like discussed there is also `incident.status`

jpoppe: This was already there, maybe could we use `props validation` on `styled components`  to prevent this in the future.

Removed it anyway ;)

src/components/ChildIncidents/index.js
jpoppe: To make it less confusing and 'more readable' I un-destructured everything except props.

src/shared/services/transform/index.test.js
jpoppe: Done, method was covered by tests from the components and this test is not required if we would follow the `testing-library best practices`. That said it crossed my mind to explicitly add this test since we are inconsistent by applying those best practices anyway, test has been added ;)

src/shared/services/transform/index.js
jpoppe: Did not know we had that one, done ;)

MauritsR (approved):
Nice work 👍

aditudorache (approved):

PR #1246, opened 2020-11-16
[SIG-3324] Reset the edit mode when the incident is changed

aditudorache:
This PR initializes the ChangeValue state each time the incident changes. This fixes a bug where bij navigation to a different incident, the edit state of a change value remained open.


MauritsR (approved):
👍

PR #1262, opened 2020-11-18
[SIG-3271] Change in the municipality autorization text

aditudorache:
This PR makes a minor change in the municipality autorization text


MauritsR (approved):
👍

PR #1264, opened 2020-11-18
[SIG-3270] Add a application variant style to the Button component and use it consistent

aditudorache:
This PR
- Adds a standard application variant style to the Button
- Uses this consistently throughout the code
- Ensures the button style remains consistent when this is renders as anchor `<a/>`


src/components/Button/index.js
```diff
@@ -10,4 +10,11 @@ const Button = styled(AscButton)`
   }
 `;
 
+export const ApplicationButton = styled(Button).attrs({ variant: 'application', type: 'button' })`
```
MauritsR, line 13: Is it possible to do this in the Button component directly?

  ${({ variant, type }) => (type === 'button' && variant === 'application') && `
    padding: ${themeSpacing(1, 4)};

    // Required for buttons that are rendered as 'Link'
    color: ${themeColor('tint', 'level7')};
  `}

MauritsR (changes requested):

src/components/Button/index.js
aditudorache: `ApplicationButton` is the SIA **implementation** of the 'application' type Button variant of `asc`.
I wouldn't place specific style that only serves this implementation. By doing that would keep one Button component would be difficult to understand later what it does and when it can be used..

boris-arkenaar: I agree with @MauritsR 
Keep consistent with how `asc` works, or it'll get confusing.
So, `asc` works with a `Button` component with different `type`s and `variant`s. Keep that same structure here. If you change something for the `application` `variant`, then do that and don't create a separate `Application` component.

aditudorache: All right, will implement that!

boris-arkenaar (approved):

MauritsR (approved):

PR #1271, opened 2020-11-23
[SIG-3385] Fixes conditional rendering of vakantieverhuur questions

MauritsR:
This PR fixes the following bugs:

- When entering 'vakantieverhuur' step-2 questions, all first-level questions should be visible
- When answering 'Weet ik niet' on the first question, the same questions are shown as when answering 'Ja' or 'Nee'
- When entering 'wonen overig' step-2 questions, only the first question should be visible


jpoppe (approved):
Nice!

aditudorache (approved):

PR #1272, opened 2020-11-23
[SIG-3382] Fixes saving of the directing department in the filter

aditudorache:
This PR 
- Fixes the bug of the directing deparment save in the filter
- Improves the unit test coverage
- Updates the unittest tresholds


MauritsR (approved):
👍

PR #1276, opened 2020-11-24
[SIG-3387] Disable save and cancel buttons during splitsing

aditudorache:
This PR disables save and cancel buttons during splitsing


src/signals/incident-management/containers/IncidentSplitContainer/components/IncidentSplitForm/index.js
```diff
@@ -103,6 +103,7 @@ IncidentSplitForm.propTypes = {
   }).isRequired,
   subcategories: PropTypes.array.isRequired,
   directingDepartments: directingDepartmentsType.isRequired,
+  isSaving: PropTypes.bool.isRequired,
```
MauritsR, line 106: Minor point: I'd use `isSubmitting` instead of `isSaving` to keep the relation with the naming of the form consistent (`onSubmit`).

MauritsR (changes requested):

MauritsR (approved):

PR #1277, opened 2020-11-24
[SIG-3392] Hides the  'delen' button when the the split status is s

aditudorache:
This PR hides the delen button also when the status is s

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
@@ -90,7 +90,7 @@ const DetailHeader = () => {
   const location = useLocation();
 
   const showSplitButton = useMemo(() => {
-    if (incident.status.state === 'o' || incident.status.state === 'a') return false;
+    if (['o', 'a', 's'].some(value => value === incident.status.state)) return false;
 
     if (incident?._links?.['sia:parent']) return false;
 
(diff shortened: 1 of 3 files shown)


src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
```diff
@@ -90,7 +90,7 @@ const DetailHeader = () => {
   const location = useLocation();
 
   const showSplitButton = useMemo(() => {
-    if (incident.status.state === 'o' || incident.status.state === 'a') return false;
+    if (['o', 'a', 's'].some(value => value === incident.status.state)) return false;
```
jpoppe, line 93: I would use `includes`  instead of `some` here.

jpoppe (commented):
I left a note, looks good!

src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
aditudorache: Thanks, I did that first, but further in this file is this construction with `some` used so I prefer to keep it consistent with the rest of the code. Performance wise makes no difference.

MauritsR: I'd use `includes` as well - if that's the initial thought for all 3 of us, it makes sense to use `includes` here and change the other occurance of `some` to `includes` as well

MauritsR (approved):

PR #1280, opened 2020-11-25
Adds category_url to value of extra questions

MauritsR:
This PR adds the category_url of the incident to the answer of every extra question. 

Functionally, this means that the backend is able to relate questions/answers on an incident to a specific (sub)category. This allows for filtering where only users with rights to specific categories can see the answers to those questions.


aditudorache (approved):
Nice!

PR #1281, opened 2020-11-25
[SIG-3301] Add the aria and role attributes required by WCAG to the Logo

aditudorache:
This PR provides the required aria and role attributes to the Logo that are required for the WCAG

### DIFF
--- a/internals/schemas/app.schema.json
+++ b/internals/schemas/app.schema.json
@@ -199,6 +199,10 @@
         "urgentContactInfo": {
           "description": "Extra information to show in case of an error",
           "type": "string"
+        },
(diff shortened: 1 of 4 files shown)


internals/schemas/app.schema.json
```diff
@@ -199,6 +199,10 @@
         "urgentContactInfo": {
           "description": "Extra information to show in case of an error",
           "type": "string"
+        },
+        "logoDescription": {
+          "description": "ARIA description fof the site logo",
```
MauritsR, line 204: `fof` -> `for`

MauritsR (changes requested):

PR #1285, opened 2020-11-27
[SIG-3207] Integrate keycloak

jpoppe:
**TO THE REVIEWERS**: The `keycloak` configuration option is TEMPORARY, we need to run Grip next too Keycloak for a couple of days after this the configuration will live in `oidc` again, and code could be adjusted accordingly and tests can be modified.


aditudorache (changes requested):
Looks good so far. I see that we don't use the POC local keycloak configuration at all so let's remove that and the references for now. I couldn't test the work because of some authorization issues with my account. I'll try again and will finish the review.

MauritsR:
Removed config for running Keycloak in Docker: this will be implemented in a separate PR

Todo:
- [ ] Implement token refresh flow

src/app.js
```diff
@@ -82,26 +82,28 @@ if (module.hot) {
...
-}
+authenticate()
+  .then(credentials => store.dispatch(authenticateUser(credentials)))
+  .finally(() => {
+    render();
 
-if ('serviceWorker' in navigator && process.env.PROXY) {
-  navigator.serviceWorker.register('/sw-proxy.js');
-}
+    // Install ServiceWorker and AppCache in the end since
```
aditudorache, line 91: Can we extract this service worker instantiation to a separate function?

src/app.js
```diff
@@ -82,26 +82,28 @@ if (module.hot) {
...
-  // eslint-disable-next-line global-require
-  const runtime = require('offline-plugin/runtime');
-
-  runtime.install({
-    onUpdateReady: () => {
-      runtime.applyUpdate();
-    },
-  });
-}
+authenticate()
```
aditudorache, line 86: Nice solution!

src/shared/services/auth/auth.js
```diff
@@ -71,30 +73,18 @@ function catchError() {
...
-  if (AUTH_PARAMS.some(param => params[param] === undefined)) {
-    return;
-  }
-
+function saveToken(state, accessToken) {
   // The state param must be exactly the same as the state token we
   // have saved in local storage (to prevent CSRF)
   const localStateToken = storage.getItem(STATE_TOKEN_KEY);
-  const paramStateToken = decodeURIComponent(params.state);
+  const paramStateToken = decodeURIComponent(state, accessToken);
```
aditudorache, line 80: This function only accepts one parameter.

src/shared/services/auth/auth.js
```diff
@@ -113,6 +103,50 @@ function handleCallback() {
...
+/**
+ * Handle login callback.
+ */
+async function handleCallback() {
+  // Parse query string into object
+  const params = queryStringParser(global.location.hash);
+
+  if (!params || !params.state) return;
+
+  if (params.code) { // handling keycloak callback
```
aditudorache, line 139: Maybe we can here in the comments make distinction between code and token flow and not specific keycloak an other callbacks.

src/shared/services/auth/auth.js
```diff
@@ -113,6 +103,50 @@ function handleCallback() {
...
+  });
+
+  return response.json();
+}
+
+/**
+ * Handle login callback.
+ */
+async function handleCallback() {
+  // Parse query string into object
```
aditudorache, line 134: This comment is not needed

src/shared/services/auth/auth.js
```diff
@@ -113,6 +103,50 @@ function handleCallback() {
...
+  if (!params || !params.state) return;
+
+  if (params.code) { // handling keycloak callback
+    const token = await fetchTokenByCode(params.code);
+    saveToken(params.state, token.access_token);
+    return;
+  }
+
+  // handling other callbacks
+  if (AUTH_PARAMS.some(param => params[param] === undefined)) return;
```
aditudorache, line 146: What does this check? Is this a common way to check the authorization was is not successfull?

aditudorache (changes requested):
Nice job! I am missing the refresh_token scenario here. Is that correct?

src/shared/services/auth/auth.js
MauritsR: Solved in [1308](https://github.com/Amsterdam/signals-frontend/pull/1308)

src/app.js
MauritsR: Solved in PR 1308

aditudorache (changes requested):
In the last test, it looks like that back action on login keycloak page doesn't return to the login choice page (manage/incidents) but to the previous page (incident/beschrijf)

MauritsR:
> In the last test, it looks like that back action on login keycloak page doesn't return to the login choice page (manage/incidents) but to the previous page (incident/beschrijf)

Logging in with Keycloak now returns to the correct

aditudorache (approved):

PR #1289, opened 2020-12-01
(Chore) Improve config

aditudorache:
This PR keeps the `amsterdam` configuration after generalizing the base config. This is needed for backwards compatibility with the existing deployment pipelines and the amsterdam netlify configuration.

Importatnt: this PR  [https://github.com/Amsterdam/signalen/pull/41](https://github.com/Amsterdam/signalen/pull/41) must be merged first to avoid crashing the deployment pipeline!


Makefile
```diff
@@ -12,7 +13,7 @@ build:                              ## Build docker image
 	BUILD_ENV=$(BUILD_ENV) $(dc) build --no-cache
 
 start:                              ## Run frontend
-	$(dc) run --service-ports frontend
+	CONFIG=$(CONFIG) npm start
```
MauritsR, line 16: In my opinion simple npm commands don't belong in this file - especially if the `start` and `stop` commands use different tooling (docker vs npm)

aditudorache: Fair enough, will revert this

app.amsterdam.json
```diff
@@ -0,0 +1,129 @@
...
+    "title": "Signalen Informatievoorziening Amsterdam",
+    "urgentContactInfo": "Voor spoedeisende zaken kunt u ook telefonisch contact opnemen met 14 020.",
+    "logoDescription": "Logo gemeente Amsterdam, naar amsterdam.nl"
+  },
+  "links": {
+    "help": "https://tamtam.amsterdam.nl/do/office?id=1723860-6f6666696365",
+    "home": "https://www.amsterdam.nl",
+    "privacy": "https://www.amsterdam.nl/privacy/specifieke/privacyverklaringen-wonen/meldingen-overlast-privacy/"
+  },
+  "logo": {
```
boris-arkenaar, line 42: You can leave the whole `logo` object. Only define it if you need a logo (other than the Amsterdam logo).

app.amsterdam.json
```diff
@@ -0,0 +1,129 @@
...
+  "oidc": {
+    "authEndpoint": "https://acc.api.data.amsterdam.nl/oauth2/authorize",
+    "clientId": "sia",
+    "responseType": "token",
+    "scope": "SIG/ALL"
+  },
+  "sentry": {
+    "dsn": "https://3de59e3a93034a348089131aa565bdf4@sentry.data.amsterdam.nl/27"
+  },
+  "theme": {
```
boris-arkenaar, line 109: You can leave the whole `theme` object as well. It contains the default values in this case, from https://github.com/Amsterdam/amsterdam-styled-components/blob/master/packages/asc-ui/src/theme/default/colors.ts

app.amsterdam.json
aditudorache: We need to keep this so the impact on our build street minimal is.

MauritsR (dismissed):

boris-arkenaar (approved):

PR #1291, opened 2020-12-02
Naming conventions ADR

boris-arkenaar:
Proposed naming conventions ADR for:

- folder structure
- file and folder names
- component names
- styling names and structure

### DIFF
--- a/docs/adr/0009-naming-conventions.md
+++ b/docs/adr/0009-naming-conventions.md
@@ -14,7 +14,7 @@ Date: 2020-09-24
 - The aim is to create consitency at all levels in the application
 - A `src/pages` folder will hold 'root' components and reflect the navigation structure (menu, routes) of the app
 - Generic/shared components, hooks, interfaces, services, in their respective folders in `src`.
-- Tests will be placed in a __tests__ folder at te same level as the the 'root' component
+- Tests will be placed in a __tests__ folder at the same level as the file under test.
 
 - Components needed only by one 'root' component go into that root component's folder in a separate folder with the name of the component, no extra nesting.
 - Components needed by more than one 'root' component go into `src/components` 
--- a/docs/adr/0009-naming-conventions.md
+++ b/docs/adr/0009-naming-conventions.md
@@ -91,7 +91,7 @@ src
         __tests__
         Form.tsx
       Kto.tsx
-      index.tsx
+      index.ts
     settings
       users
       ...
--- a/docs/adr/0009-naming-conventions.md
+++ b/docs/adr/0009-naming-conventions.md
@@ -21,7 +21,7 @@ Date: 2020-09-24
 - Component folders will be written in `PascalCase`, non component folders will be written in `kebab-case`. This includes 'pages', 'services', etc.
 - The nested components will prevent as much as possible prefixing with the parent component name. 
   A child component of `IncidentSplit` will be `IncidentSplit/ChangeValue` and **not** `IncidentSplitChangeValue`
-- Component that have more than one file (containers) will keep the convention that the additional files (like context, reducer, actions, constants, styles ...) will **NOT be prefixed** with the component name:
+- Components that have more than one file (containers) will keep the convention that the additional files (like context, reducer, actions, constants, styles, ...) will **not be prefixed** with the component name:
 
 ### Example of a `container` component
 ```
--- a/docs/adr/0009-naming-conventions.md
+++ b/docs/adr/0009-naming-conventions.md
@@ -20,7 +20,7 @@ Date: 2020-09-24
 - Components needed by more than one 'root' component go into `src/components` 
 - Component folders will be written in `PascalCase`, non component folders will be written in `kebab-case`. This includes 'pages', 'services', etc.
 - The nested components will prevent as much as possible prefixing with the parent component name. 
-  A child component of `IncidentSplit` will be `ChangeValue` and **Not** `IncidentSplitChangeValue`
+  A child component of `IncidentSplit` will be `IncidentSplit/ChangeValue` and **not** `IncidentSplitChangeValue`
 - Component that have more than one file (containers) will keep the convention that the additional files (like context, reducer, actions, constants, styles ...) will **NOT be prefixed** with the component name:
 
 ### Example of a `container` component
--- a/docs/adr/0009-naming-conventions.md
+++ b/docs/adr/0009-naming-conventions.md
@@ -11,143 +11,126 @@ Date: 2020-09-24
 ### Principles
 
 - Keep it simple, reduce nesting
+- The aim is to create consitency at all levels in the application
 - A `src/pages` folder will hold 'root' components and reflect the navigation structure (menu, routes) of the app
-- Components needed only by one 'root' component go into that root component's folder, no extra nesting
-- Components needed by more than one 'root' component go into `src/components`
-- Component folders, files, and auxiliary files, camel cased, starting with upper case
-- Component names and file names will be unique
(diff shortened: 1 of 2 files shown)


docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,126 @@
...
+### Principles
+
+- Keep it simple, reduce nesting
+- 'container'-type components in their own folder directly under `src`.
+- Components needed only by one 'root' component go into that root component's folder, no extra nesting
+- Components needed by more than one 'root' component go into `src/components`
+- Component folders, files, and auxiliary files, camel cased, starting with upper case
+- Component names and file names will be unique
+- Unique `data-testid` names by prefixing them with component names
+- 'root' level components, hooks, interfaces, services, in their respective floders.
```
MauritsR, line 20: -> folders

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Example tree
+
+```
+src
+  App
+    App.tsx
+    App.styles.ts
+    App.test.ts
+  DefaultTexts
```
MauritsR, line 30: I feel like having all container-type components at the root level will pollute the `src` directory. Why not keep a `containers` directory at this level?

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,126 @@
...
+`;
+```
+
+## Action plan
+
+- Merge `src/shared` to `src`
+- Merge or move components from `src/containers` to either `src` or `src/components`
+- Merge `src/signals/incident/containers` with `src`
+- Merge `src/signals/incident/components` into the (now) root incident component (keep the form components grouped together, for now).
+- Continue this process for `incident-management`
```
MauritsR, line 126: To take this as an example, are you proposing to add an IncidentManagement folder in `src`, which is a module that imports containers from other folders in the `src` directory? 

I'm not against it per se, but it would make it harder to understand the hierarchy between modules/containers. For example that the `DefaultTextsAdmin` container is used by `IncidentManagement` module.

MauritsR (commented):
Looks like this would improve consistency and developer experience (especially finding files). I added some comments.

docs/adr/0009-naming-conventions.md
boris-arkenaar: Yes, that might be a good idea.

docs/adr/0009-naming-conventions.md
boris-arkenaar: No. Same as for the current `incident` folder. Move the containers over to `src`(`/containers`) and components into the respective container they are used in.
So, no `incident` and `incident-management` 'module'-like folders anymore. Their function now is unclear and inconsistent and there is now clear separation between the two.

docs/adr/0009-naming-conventions.md
boris-arkenaar: I don't like marking them as `containers` though. It's the structure of the application.
How about `src/app`? Or even `src/App` and put the `App` component files directly in there.

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Component files
+
+### Principles
+
+- Keep it simple, reduce nesting
+- A `src/pages` folder will hold 'root' components and reflect the navigation structure (menu, routes) of the app
+- Components needed only by one 'root' component go into that root component's folder, no extra nesting
+- Components needed by more than one 'root' component go into `src/components`
+- Component folders, files, and auxiliary files, camel cased, starting with upper case
```
aditudorache, line 17: The folder names are starting with lower case as I can see in the example below. what should it be? `kebab-case` or `PascalCase`? Is not usual to use `camelCase` when naming `folders`.

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,158 @@
...
+- Component names and file names will be unique
```
aditudorache, line 18: I don't agree here: this implies that the small child component names will be needing extra prefixing. It should possible naming files like in the example below: `report/Incident.tsx` and `incidents/incident/Incident.tsx`

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,158 @@
...
+
+2020-12-02 Proposed
+
+## Component files
+
+### Principles
+
+- Keep it simple, reduce nesting
+- A `src/pages` folder will hold 'root' components and reflect the navigation structure (menu, routes) of the app
+- Components needed only by one 'root' component go into that root component's folder, no extra nesting
```
aditudorache, line 15: Agreed if each component has its own folder together with it's tests (accepting the fact that almost each component will have its own test).
```
kto
      Form
          Form.tsx
          Form.test.tsx
          index.ts
      Kto.tsx
      Kto.test.tsx
      index.ts
```

Otherwise the size of a main component folder will be to large and difficult to understand.

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,158 @@
...
+      incidents
+        incident
+          split
+            IncidentSplit.tsx
+            SplitForm.tsx
+            SplitFormIncident.tsx
+          Incident.tsx
+        Incidents.tsx
+        IncidentsList.tsx
+        IncidentsMap.tsx
```
aditudorache, line 42: This one for instance should be called `Map.tsx` and placed in the `Map` folder

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,158 @@
...
+  hooks
+    useFetch.ts
+    useFetch.test.ts
+  interfaces
+  services
+```
+
+This gives a clear hierarchy without much nesting.
+
+## Styling
```
aditudorache, line 99: I thought about this and I don't agree with imposing this construction for many reasons, few of these would be: 
- loosing the capability (or making difficult) implementing small styles inside a component. I have further searched the internet and I don't see a trend in this direction.
- Is almost impossible to convert the whole codebase to this format in a acceptable period of time
- Requires for almost each component an extra file.

docs/adr/0009-naming-conventions.md
```diff
@@ -0,0 +1,158 @@
...
+2020-12-02 Proposed
+
+## Component files
+
+### Principles
+
+- Keep it simple, reduce nesting
+- A `src/pages` folder will hold 'root' components and reflect the navigation structure (menu, routes) of the app
+- Components needed only by one 'root' component go into that root component's folder, no extra nesting
+- Components needed by more than one 'root' component go into `src/components`
```
aditudorache, line 16: 👍🏻

aditudorache (changes requested):

aditudorache (approved):

docs/adr/0009-naming-conventions.md
```diff
@@ -11,143 +11,126 @@ Date: 2020-09-24
...
-      IncidentPreview.tsx
-      IncidentWizard.tsx
+      Incident
+        __tests__
+        IncidentForm
+        IncidentNavigation
+        IncidentPreview
+        IncidentWizard
+        Incident.tsx
+        index.tsx
```
boris-arkenaar, line 87:
```suggestion
      __tests__
      Form
      Navigation
      Preview
      Wizard
      Report.tsx
      index.ts
```

MauritsR (approved):

aditudorache:
Closes in favor or #1396

PR #1296, opened 2020-12-03
[SIG-3207] Adds functionality to refresh access tokens

MauritsR:
This PR adds functionality to enable refreshing access tokens

### Background info

The existing authentication mechanism implements `Implicit flow`, where a long-lived _access_token_ is provided (e.g. 10hrs lifespan) and the user is presented with a login screen after its expiry. 

Keycloak uses the `Authorization Code` flow, which results in the client storing a _access token_ and a _refresh token_. The lifespan of the _access token_ is 5 minutes and the lifespan of the _refresh token_ is 30 minutes. When an access token is refreshed, a fresh refresh token is also obtained. 

**WIP**
Although the implementation in this PR is unfinished, I've created this PR to start a discussion on the implementation of this feature and to discuss possible risks.


src/shared/services/auth/auth.js
```diff
@@ -70,6 +71,67 @@ function catchError() {
...
+      method: 'POST',
+      headers: { 'Content-type': 'application/x-www-form-urlencoded' },
+    });
+
+    if (response.status === 200) {
+      const data = await response.json();
+      storage.setItem(ACCESS_TOKEN_KEY, data.access_token);
+      storage.setItem(REFRESH_TOKEN_KEY, data.refresh_token);
+    } else {
+      logout();
```
MauritsR, line 106: This code means that if a user loses internet connection at the wrong time, he'll be logged out. There's a couple of other options: (1) ignore non-successful responses and wait for the next refresh interval to be triggered, (2) retry mechanism

Thoughts?

src/shared/services/auth/auth.js
```diff
@@ -244,9 +310,14 @@ export function logout() {
...
  */
 export async function initAuth() {
-  restoreAccessToken(); // Restore acces token from local storage
+  restoreAccessToken(); // Restore access token from local storage
   catchError(); // Catch any error from the OAuth2 authorization service
   await handleCallback(); // Handle a callback from the OAuth2 authorization service
+
+  if (getOauthDomain() === 'keycloak') {
+    await handleRefreshAccessToken(); // Refresh access token if expired
+    setRefreshAccessTokenInterval(); // Start periodic check for validity of access token, and refresh if expired
```
MauritsR, line 319: Ideally, we would refresh access tokens silently on demand (e.g. refresh tokens when a user requests data but the access token is expired). 

However, this is impractical in this application because (1) UI authentication state is directly dependent on the access_token expiry time and (2) we use rotating refresh tokens (i.e. with a 30 min lifespan) so if we don't refresh often, the user will be logged out after 30 minutes of not using the app.

src/shared/services/auth/services/access-token-parser/access-token-parser.js
```diff
@@ -13,7 +13,7 @@ function decodeToken(token) {
   }
 }
 
-export default function parseAccessToken(token) {
+export default function accessTokenParser(token) {
```
aditudorache, line 16: :+1:

aditudorache (changes requested):

src/shared/services/auth/services/access-token-parser/access-token-parser.js
boris-arkenaar: I don't agree here.
Function names should indicate an action, not a 'thing.'
`parseAccessToken` conveys an action, 'accessTokenParser` conveys a 'thing'.
Using the same name for file/folder is a good thing, though 👍

src/shared/services/auth/auth.js
```diff
@@ -26,6 +26,9 @@ const AUTH_REDIRECT_URI = `${global.location.protocol}//${global.location.host}/
 
 const domainList = ['datapunt', 'grip'];
 
+const ACCESS_TOKEN_REFRESH_TIME = 60000; // If token expiry is less than 60000 ms in the future, it should be refreshed
```
bartjkdp, line 29:
```suggestion
const ACCESS_TOKEN_REFRESH_TIME = 60000; // If token expiry is less than 60s in the future, it should be refreshed
```

src/shared/services/auth/auth.js
```diff
@@ -26,6 +26,9 @@ const AUTH_REDIRECT_URI = `${global.location.protocol}//${global.location.host}/
...
+const ACCESS_TOKEN_REFRESH_INTERVAL = 20000; // Check expiry of access token every 20000 ms
```
bartjkdp, line 30:
```suggestion
const ACCESS_TOKEN_REFRESH_INTERVAL = 20000; // Check expiry of access token every 20s
```

src/shared/services/auth/auth.js
```diff
@@ -80,41 +83,37 @@ function saveRefreshToken(refreshToken) {
  */
 async function refreshTokens() {
   const { authEndpoint, realm, clientId } = configuration.keycloak;
-  const token = storage.getItem(REFRESH_TOKEN_KEY);
+  const refreshToken = storage.getItem(REFRESH_TOKEN_KEY);
+  const decodedRefreshToken = accessTokenParser(refreshToken);
+
+  if (decodedRefreshToken.expiresAt * 1000 < Date.now()) return;
```
bartjkdp, line 89: You are depending on local system time here. When the local system time is wrong, the application will not work. Why not just try the access token and if the token is not working, try to refresh the token?

src/shared/services/auth/auth.js
```diff
@@ -80,41 +83,37 @@ function saveRefreshToken(refreshToken) {
...
+    storage.setItem(REFRESH_TOKEN_KEY, refresh_token);
   }
 }
 
 const handleRefreshAccessToken = async () => {
   const decoded = accessTokenParser(storage.getItem(ACCESS_TOKEN_KEY));
 
-  // Should refresh if access token expires within the next minute
-  const shouldRefresh = decoded.expiresAt * 1000 - 60000 < Date.now();
+  const shouldRefresh = decoded.expiresAt * 1000 - ACCESS_TOKEN_REFRESH_TIME < Date.now();
```
bartjkdp, line 116: Same remark as above ^^

src/shared/services/auth/auth.js
```diff
@@ -123,12 +122,13 @@ const handleRefreshAccessToken = async () => {
 
 const setRefreshAccessTokenInterval = () => {
   const refreshToken = storage.getItem(REFRESH_TOKEN_KEY);
-  const decoded = accessTokenParser(refreshToken);
+  const decodedRefreshToken = accessTokenParser(refreshToken);
 
-  if (decoded.expiresAt < Date.now()) {
+  // Only start interval if refresh token is not expired
+  if (decodedRefreshToken.expiresAt * 1000 > Date.now()) {
```
bartjkdp, line 128: Same remark as above ^^

src/shared/services/auth/auth.js
```diff
@@ -63,13 +67,71 @@ function handleError(code, description) {
...
+  }
+};
+
+const setRefreshAccessTokenInterval = () => {
+  const refreshToken = storage.getItem(REFRESH_TOKEN_KEY);
+  const decodedRefreshToken = accessTokenParser(refreshToken);
+
+  // Only start interval if refresh token is not expired
+  if (decodedRefreshToken.expiresAt * 1000 > Date.now()) {
+    setInterval(() => {
```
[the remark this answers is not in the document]
MauritsR: Refactored this code and it's now cleared appropriately (and not set multiple times)

src/shared/services/auth/auth.js
MauritsR: Your suggestion makes sense, but the current behavior is that users are authenticated as long as they have an access token that is not expired. 

The Keycloak adapter also uses Date.now() under the hood to check for token expiry. There's an option to provide a time skew (effectively an estimate of the time difference between client and server), but I can't think of a way to effectively estimate it. 

I can set a default time skew of e.g. 2 seconds

src/components/LoginPage/index.test.js
```diff
@@ -38,6 +42,18 @@ describe('components/LoginPage', () => {
...
     fireEvent.click(button);
 
-    expect(loginSpy).toHaveBeenCalledWith('datapunt');
+    expect(loginSpy).toHaveBeenCalledWith('grip');
+  });
+
+  it('should login on keycloak when Inloggen Keycloak button is clicked', () => {
+    const loginSpy = jest.spyOn(auth, 'login');
+    const { getByText } = render(withAppContext(<LoginPage />));
+    const button = getByText('Inloggen Keycloak').parentNode;
```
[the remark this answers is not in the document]
MauritsR: `screen.getByText` is indeed more appropriate here, fixed it

src/shared/services/auth/auth.js
```diff
@@ -1,214 +1,28 @@
...
-  if (!params || !params.state) return;
-
-  if (params.code) {
-    // Handle keycloak callback
-    const tokens = await fetchTokensByCode(params.code);
-    saveToken(params.state, tokens.access_token);
-    saveRefreshToken(tokens.refresh_token);
-    return;
+export function getAuth() {
+  if (configuration.keycloak && getOauthDomain() === 'keycloak') {
```
boris-arkenaar, line 21: Use ternary expression here.

boris-arkenaar: Why no function expressions?

src/shared/services/auth/auth.js
```diff
@@ -217,146 +31,53 @@ async function handleAuthorizationCallback() {
...
-  if (accessToken) {
-    return { userName: getName(), userScopes: getScopes(), accessToken };
-  }
-
-  return null;
-};
+/**
+ * Perform user authentication on app init.
+ */
+// eslint-disable-next-line require-await
```
boris-arkenaar, line 82: Could you elaborate?

src/shared/services/auth/auth.js
MauritsR: What do you mean with function expressions in this case?

src/shared/services/auth/auth.js
MauritsR: For the authenticate function, the `authz` implementation is synchronous while the `keycloak` implementation is asynchronous (due to how keycloak-js implements the init function). I've added async to this line to have Babel this function promisified. 

Adding `await` which is what the eslint rule suggests, triggers the `no-return-await` linting rule

src/shared/services/auth/services/keycloak.test.js
```diff
@@ -0,0 +1,205 @@
...
+        expect(updateTokenSpy).toHaveBeenCalled();
+
+        jest.clearAllTimers();
+      });
+    });
+
+    describe('stopRefreshInterval', () => {
+      it('removes refresh interval stored in keycloak instance', () => {
+        // jest.spyOn(keycloak.keycloak, 'updateToken').mockImplementation(() => {});
+        // clearInterval = jest.fn();
```
aditudorache, line 196: forgot to remove?

src/signals/incident/components/form/RadioInputGroup/index.test.js
```diff
@@ -60,6 +54,32 @@ describe('Form component <RadioInput />', () => {
...
+            incident: {
+              [metaFields.name]: {
+                id: 'foo',
+              },
+            },
+          },
+        },
+      });
+
+      expect(wrapper).toMatchSnapshot();
```
aditudorache, line 80: We don't use snapshots anymore. Could you use the testing-library here?

src/signals/incident/components/form/RadioInputGroup/index.test.js
```diff
@@ -60,6 +54,32 @@ describe('Form component <RadioInput />', () => {
       expect(wrapper).toMatchSnapshot();
     });
 
+    it('should render info from current selection correctly', () => {
```
aditudorache, line 57: Is this not out of scope here? Maybe an separate chore PR?

aditudorache (changes requested):
Looks good! Works good with the other authentication types. We are almost there!! 🎉. I have added couple of suggestions for the finishing touch.

aditudorache:
@bartjkdp, @boris-arkenaar, Could you please remove your assignment when you are ready with reviewing? I would really help us to keep track on whose turn it is 😄!

src/signals/incident/components/form/RadioInputGroup/index.test.js
MauritsR: I added it because of a coverage error was triggered

MauritsR: Removed this test

aditudorache (approved):

src/shared/services/auth/auth.js
boris-arkenaar: `export const getAuth = () => configuration.keycloak && getOauthDomain() === 'keycloak' ? keycloak : authz;`

MauritsR: You're right - I've used function here because it was line with code already in place. Will refactor to arrow functions. 

(I was confused because your comment says 'why no function expressions' while function expressions are being used)

aditudorache (approved):

PR #1297, opened 2020-12-03
[SIG-3399] Cypress tests category history

arjandata:


e2e-tests/cypress/integration/manageSettings/manageCategories.spec.js
```diff
@@ -131,7 +180,7 @@ describe('Manage categories', () => {
       cy.checkHeaderText('Categorieën');
 
       // Open category Afwatering brug
-      cy.contains('Afwatering brug').click();
+      cy.contains('Afgewaterde brug').click();
```
aditudorache, line 183: Shouldn't this be a string const? Is used 2 times here.

aditudorache (approved):

MauritsR (approved):

PR #1299, opened 2020-12-07
[SIG-3045] Prevent default submit in the map container

aditudorache:
This PR fixes the issues that a submit event was triggered in the map container for AutoSuggest and GPSButton

### DIFF
--- a/src/components/AutoSuggest/__tests__/AutoSuggest.test.js
+++ b/src/components/AutoSuggest/__tests__/AutoSuggest.test.js
@@ -1,5 +1,5 @@
 import React, { Fragment } from 'react';
-import { render, fireEvent, act } from '@testing-library/react';
+import { render, fireEvent, act, screen } from '@testing-library/react';
 import { withAppContext } from 'test/utils';
 
 import AutoSuggest, { INPUT_DELAY } from '..';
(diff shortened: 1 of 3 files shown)


src/components/AutoSuggest/__tests__/AutoSuggest.test.js
```diff
@@ -390,6 +390,29 @@ describe('src/components/AutoSuggest', () => {
...
+      const { container } = render(
+        withAppContext(
+          <form onSubmit={mockedOnSubmit}>
+            <AutoSuggest {...props} />
+            <input type="text" name="foo" />
+          </form>
+        )
+      );
+
+      const input = container.querySelector('input[aria-autocomplete=list]');
```
MauritsR, line 404: It should be possible to use screen to select this input

MauritsR (changes requested):

src/components/AutoSuggest/__tests__/AutoSuggest.test.js
aditudorache: image

aditudorache: I would gladly do but is not possible :-)

MauritsR (approved):

PR #1302, opened 2020-12-09
[SIG-3333] Container selection component

aditudorache:
This PR:
- adds a generic definition for the map items
- adds the metadata configuration for the ContainerSelect
- adds svg icons for the specific implementations (will later be moved to the backoffice)
- add temporary (test) functionality for adding containers
- renders the added containers in the Summary component


MauritsR (approved):
Looks good :)

PR #1310, opened 2020-12-14
[FE#94] Filter on assignee

boris-arkenaar:
closes Signalen/frontend#91
closes Signalen/frontend#94

This PR makes the following changes to users assigned to an incident:

- Changes `assigned_user_id` to `assigned_user_email`.
- Adds a filter (using `AutoSuggest`).
- Uses a checkbox to allow filtering on signals that have not been assigned.
- Allows sorting.

The following additions are made to the `AutoSuggest` component:

- `disabled` prop has been added to be able to disable the form component.
- `id` prop has been added to be able to target the component with a label (`htmlFor`).

Screenshot 2021-01-04 at 14 14 14
Screenshot 2021-01-04 at 14 15 06
Screenshot 2021-01-04 at 14 15 25
Screenshot 2021-01-04 at 14 15 40


package.json
```diff
@@ -89,6 +89,7 @@
     "react-reactive-form": "1.0.27",
     "react-redux": "^7.2.2",
     "react-router-dom": "^5.2.0",
+    "react-select": "^3.1.1",
```
aditudorache, line 92: Why do we need this package? It increases the bundle by 27k zipped! Let's find another solution that is not that expensive. We have an autosuggest component, maybe you can use that one?

aditudorache (changes requested):

package.json
boris-arkenaar: Hmm, that's unfortunate.
Where can I find your autosuggest component?

boris-arkenaar: Done. Updated the initial message of this PR at the top of this page.

src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -451,6 +486,36 @@ const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, o
...
+              </Label>
+              <div>
+                <Checkbox
+                  data-testid="filterNotAssigned"
+                  checked={state.options.assigned_user_email === 'null'}
+                  id="filter_not_assigned"
+                  name="notAssigned"
+                  onClick={onNotAssignedChange}
+                />
+                <AscLabel htmlFor="filter_not_assigned" label="Niet toegewezen" />
```
aditudorache, line 503: I would prefer not to use `Asc<Component>` throughout the code but define it as a variant of the own `components/Label` component

MauritsR (changes requested):

src/signals/incident-management/components/FilterForm/index.js
boris-arkenaar: It is already used in this component. Besides that, if I use the `components/Label` component the alignment between the label and the checkbox is off.

src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -227,6 +239,29 @@ const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, o
...
+    [dispatch]
+  );
+
+  const onAssignedClear = useCallback(() => {
+    dispatch(setGroupOptions({ assigned_user_email: '' }));
+  }, [dispatch]);
+
+  const onNotAssignedChange = useCallback(
+    event => {
+      event.persist();
```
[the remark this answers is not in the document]
boris-arkenaar: I copied it from the `AutoSuggest` component I guess. I'll try and remove it.

MauritsR (approved):

src/signals/incident-management/components/FilterForm/__tests__/mockResponse.json
```diff
@@ -0,0 +1,57 @@
+{
```
aditudorache, line 1: we already have an `src/utils/__tests__/fixtures/user.json` mock, could you use that one please?

src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -40,6 +41,15 @@ import {
 
 import reducer, { init } from './reducer';
 
+const USERS_AUTO_SUGGEST_URL = `${configuration.USERS_ENDPOINT}?is_active=true&username=`;
```
aditudorache, line 44: The `UserAutoSuggest` is like `PDOKAutosuggest` an specific implementation of the `AutoSuggest` control. Shell we extract this to a separate component? This would keep the FilterForm component a little bit cleander.

aditudorache (changes requested):

src/signals/incident-management/components/FilterForm/index.js
boris-arkenaar: That will only extract these lines of code: https://github.com/Amsterdam/signals-frontend/pull/1310/files#diff-985e301408102dce7b18001ad6c36fedab4917573589c06456ba167ce9ee3a37R44-R52
It is not being used anywhere else (yet), so I'm not sure if it's worth extracting.

aditudorache (approved):

PR #1317, opened 2020-12-17
[SIG-3370] Hide attribution on screen readers

aditudorache:
This PR hides the map attribution text for the screen readers


MauritsR (approved):
Nice find :D 
I did not expect the leaflet attribution option to accept html, but it works well.

PR #1318, opened 2020-12-17
[SIG-3309] more specific error messages

MauritsR:
### Changes

To improve user experience, this PR adds (1) more specific error messages, and (2) the possibility to add a custom error message for the required validator.

### DIFF
--- a/src/signals/incident/components/form/Header/index.js
+++ b/src/signals/incident/components/form/Header/index.js
@@ -60,12 +60,16 @@ const Header = ({ className, meta, options, touched, hasError, getError, childre
 
       {touched && containsErrors && (
         <Fragment>
-          {hasError('required') && <ErrorItem>Dit is een verplicht veld</ErrorItem>}
+          {hasError('required') && <ErrorItem>{meta?.requiredErrorMessage || 'Dit is een verplicht veld'}</ErrorItem>}
 
-          {hasError('email') && <ErrorItem>Het moet een geldig e-mailadres zijn</ErrorItem>}
+          {hasError('email') && (
+            <ErrorItem>
+              Vul een geldig e-mailadres in, met een @ en een domeinnaam. Bijvoorbeeld: naam@domein.nl
+            </ErrorItem>
+          )}
 
           {hasError('maxLength') && (
-            <ErrorItem>U kunt maximaal {getError('maxLength').requiredLength} tekens invoeren.</ErrorItem>
+            <ErrorItem>U heeft meer dan de maximale {getError('maxLength').requiredLength} tekens ingevoerd</ErrorItem>
           )}
 
           {hasError('custom') && <ErrorItem>{getError('custom')}</ErrorItem>}
@@ -86,6 +90,7 @@ Header.propTypes = {
   meta: PropTypes.shape({
     label: PropTypes.string,
     subtitle: PropTypes.string,
+    requiredErrorMessage: PropTypes.string,
   }),
   options: PropTypes.shape({
     validators: PropTypes.arrayOf(PropTypes.any),
--- a/src/signals/incident/components/form/Header/index.js
+++ b/src/signals/incident/components/form/Header/index.js
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import styled, { css } from 'styled-components';
 import { themeSpacing, themeColor } from '@amsterdam/asc-ui';
 
+import { validateRequired } from '../../../services/custom-validators';
 import { Validators } from 'react-reactive-form';
+
+const customRequiredFnName = validateRequired().name;
 
 const Children = styled.div`
   display: flex;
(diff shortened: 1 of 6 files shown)


src/signals/incident/components/form/Header/index.test.js
```diff
@@ -19,32 +19,34 @@ describe('signals/incident/components/form/Header', () => {
     const options = {
       validators: [Validators.required],
     };
-    const { rerender, queryByText } = render(withAppContext(<Header hasError={() => {}} meta={{ label }} options={{}} />));
+    const { rerender, queryByText } = render(
```
boris-arkenaar, line 22: So, you can remove this `queryByText` const now, right?

src/signals/incident/definitions/wizard-step-1-beschrijf.js
```diff
@@ -33,6 +33,7 @@ const getControls = memoize(
           label: 'Waar is het?',
           subtitle: 'Typ het dichtstbijzijnde adres of klik de locatie aan op de kaart',
           path: 'location',
+          requiredErrorMessage: 'Kies een locatie op de kaart of vul een adres in',
```
aditudorache, line 36: This construction is too specific. Please use either an custom validator here (`src/signals/incident/services/custom-validators/index.js`) or an override of the `getError` meta function?

aditudorache (changes requested):

aditudorache:
can we keep the PR convention here? [SIG-...]

src/signals/incident/definitions/wizard-step-1-beschrijf.js
MauritsR: I started out with a custom validator, but got stuck - I gave it another try and have got a working version with custom validator (see last commit)

MauritsR:
Refactored the implementation. @boris-arkenaar @aditudorache Please have another look

src/signals/incident/components/form/Header/index.js
```diff
@@ -62,10 +67,14 @@ const Header = ({ className, meta, options, touched, hasError, getError, childre
         <Fragment>
           {hasError('required') && <ErrorItem>Dit is een verplicht veld</ErrorItem>}
```
aditudorache, line 68: How about:
```
{hasError('required') && <ErrorItem>{getError('required')} || Dit is een verplicht veld</ErrorItem>}
```

src/signals/incident/services/custom-validators/index.js
```diff
@@ -44,3 +50,8 @@ export const validatePhoneNumber = control => {
 
   return { custom: 'Ongeldig telefoonnummer, alleen cijfers, spaties, haakjes, + en - zijn toegestaan.' };
 };
+
+export const validateRequired = message =>
```
aditudorache, line 54: This is a factory not a validator, right? I would use `create` prefix or Factory suffix here.

src/signals/incident/services/custom-validators/index.js
```diff
@@ -44,3 +50,8 @@ export const validatePhoneNumber = control => {
...
+  function customRequired({ value }) {
```
aditudorache, line 55: if the name here would be `required` it will fit the required validation pattern

src/signals/incident/services/custom-validators/index.js
```diff
@@ -44,3 +50,8 @@ export const validatePhoneNumber = control => {
...
+    return value === null || value === undefined || value.length === 0 ? { custom: message } : null;
```
aditudorache, line 56: if the returnn would be `{required: message}`, this will be used in the required validator as error message

aditudorache (changes requested):

src/signals/incident/services/custom-validators/index.js
MauritsR: 😃 nice one! Makes it much better

aditudorache (approved):

src/signals/incident/components/form/Header/index.js
```diff
@@ -44,7 +42,9 @@ const SubTitle = styled.div`
 const Header = ({ className, meta, options, touched, hasError, getError, children }) => {
   const containsErrors =
     touched && (hasError('required') || hasError('email') || hasError('maxLength') || hasError('custom'));
-  const isOptional = !options?.validators?.includes(Validators.required);
+  const isOptional = !options?.validators
+    ?.map(validator => validator.name)
+    .includes('required');
```
boris-arkenaar, line 47: I'd say you could write this even simpler and more explicit with `some`:
```
  const isOptional = !options?.validators?.some(({ name }) => name === 'required');
```

src/signals/incident/components/form/Header/index.js
```diff
@@ -60,7 +58,9 @@ const Header = ({ className, meta, options, touched, hasError, getError, childre
 
       {touched && containsErrors && (
         <Fragment>
-          {hasError('required') && <ErrorItem>{getError('required') || 'Dit is een verplicht veld'}</ErrorItem>}
+          {hasError('required') && (
+            <ErrorItem>{getError('required') === true ? 'Dit is een verplicht veld' : getError('required')}</ErrorItem>
```
boris-arkenaar, line 62: This is confusing, I have to say. It can be either a boolean or a string?

MauritsR: I agree it's confusing - but in case of an error, this is what getError of react-reactive-form returns

boris-arkenaar: I see, it's because of how react-reactive-form works.

boris-arkenaar (approved):

PR #1319, opened 2020-12-17
[SIG-3395] cypress test stub PDOK calls

arjandata:
* PDOK calls are stubbed
* small fix for deelmeldingen
* small fix for manage categories


MauritsR (approved):
Looks good

PR #1323, opened 2020-12-29
[FE#97] Allow markdown for PlainText value

boris-arkenaar:
closes Signalen/frontend#97

Parse and render markdown for the value of the `PlainText` field in the incident wizard.
Do this only when the `fetchQuestionsFromBackend` feature flag is enabled.


MauritsR (commented):
Implementation looks good. What's the reason for putting it behind a feature flag?

boris-arkenaar:
@MauritsR When the feature flag `fetchQuestionsFromBackend` is enabled and questions are coming from the database we expect the value to be a string. With this PR we parse that sting for possible markdown.
In case the feature flag is disabled the hardcoded questions are used and those use react components directly for some `PlainText` values. This logically needs to be processed differently.

MauritsR (approved):
Nice work!

aditudorache (approved):

PR #1324, opened 2020-12-29
[FE#96] Update type with feature flag

boris-arkenaar:
closes Signalen/frontend#96

Adds a feature flag called `useProjectenSignalType` which defaults to `false` and does not introduce any breaking changes.
When `true`, will make a textual change to the signal type 'Groot onderhoud', becoming 'Projecten'. This is for Den Bosch and is temporarily needed until the signal types have been moved to the database.


MauritsR (approved):

internals/schemas/app.schema.json
```diff
@@ -89,6 +89,10 @@
           "description": "When true, will show extra questions after the first page of questions in the incident submission form",
           "type": "boolean"
         },
+        "useDenBoschSignalType": {
```
aditudorache, line 92: I would prefer not to use client name in the code, how about `useProjectenSingalType`?

aditudorache (approved):

aditudorache:
LGTM, could you look at het unittests as wel?

PR #1326, opened 2021-01-05
[SIG-3417] Fix 'Enter' behaviour for filtering on Firefox/Safari on Mac

MauritsR:
### Fixes

- Clicking on a checkbox and then pressing 'Enter' did not submit the filter selection
- Clicking on 'Alles selecteren' or 'Niets selecteren' for Category filters did not work
- Clicking on 'Alles selecteren' or 'Niets selecteren' and then pressing 'Enter' did not submit the filter selection

### DIFF
--- a/src/signals/incident-management/components/CheckboxList/CheckboxList.js
+++ b/src/signals/incident-management/components/CheckboxList/CheckboxList.js
@@ -185,7 +185,7 @@ const CheckboxList = ({
       const { checked: targetIsChecked, id } = target;
 
       // Firefox & Safari on Mac do not hold focus for non-input elements when clicked
-      // Force focus so that 'Enter' will submit the checkbox selection
+      // Force focus so that 'Enter' will submit the current selection
       document.getElementById(id).focus();
 
       setChecked(state => {
@@ -257,7 +257,7 @@ const CheckboxList = ({
           indent={Boolean(title)}
           tabIndex={0}
           htmlFor={groupName}
-          onClick={groupName ? null : handleToggle}
+          onClick={handleToggle}
           onKeyDown={handleKeyDown}
         >
           {toggled ? toggleNothingLabel : toggleAllLabel}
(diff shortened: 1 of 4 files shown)


src/signals/incident-management/components/CheckboxList/CheckboxList.js
```diff
@@ -182,7 +182,11 @@ const CheckboxList = ({
    */
   const handleIndividualCheck = useCallback(
     ({ target }) => {
-      const { checked: targetIsChecked } = target;
+      const { checked: targetIsChecked, id } = target;
+
+      // Firefox & Safari on Mac do not hold focus for non-input elements when clicked
+      // Force focus so that 'Enter' will submit the current selection
+      document.getElementById(id).focus();
```
aditudorache, line 189: DOM manipulation is not a good practice. Please use `useRef` here or maybe this can be done through the event object?

src/signals/incident-management/components/CheckboxList/CheckboxList.test.js
```diff
@@ -409,6 +409,9 @@ describe('signals/incident-management/components/CheckboxList', () => {
 
     expect(randomCheckbox.checked).toEqual(false);
 
+    const focusedCheckbox = container.querySelector('input[type="checkbox"]:focus');
```
aditudorache, line 412: is this not to generic? How do I know the focused element is the right one?

aditudorache (changes requested):

src/signals/incident-management/components/CheckboxList/CheckboxList.js
MauritsR: I didn't use `useRef` to avoid creating 100+ refs on rendering the filter. But good suggestion about the event object, makes a lot more sense to use that directly.

src/signals/incident-management/components/CheckboxList/CheckboxList.test.js
MauritsR: I've changed it to use `document.activeElement`

aditudorache (approved):

PR #1331, opened 2021-01-07
[SIG-3141] Cypress tests resolutions

arjandata:
* Made tests for different resolutions based on matamo data
* Stubbed all maps in SIA
* Activated test for deelmeldingen filter because bug SIG-3415 is solved
* Activated video recording for failed tests, this should help debug test failures
* Excluded some tests which can only run on a local machine
* Update Cypress to version 6.2.1
* Some small changes


MauritsR (approved):
Looks good!

PR #1335, opened 2021-01-11
Reduce categories_structured fixture

boris-arkenaar:
Reduce the size of the fixtures and the runtime of the unit tests will reduce as well.
Especially on #1310 I noticed a long-running time for the `FilterForm` unit test suite and found out this was reduced by reducing the size of the `categories_structured` fixture. This change might also fix the flaky test at #1310.


aditudorache (approved):
Good job!

MauritsR (approved):

PR #1336, opened 2021-01-11
(Chore) Convert the App container to ts

aditudorache:
This PR:
- Converts the App container component to ts
- Unused actions are removed from the app reducer
- The global state members are converted from `invariant` type to `typescript`
- The jest thresholds are lowered for branches and functions for now


boris-arkenaar (approved):
Great work!
I would still prefer types over interfaces, though.

src/containers/App/selectors.ts
```diff
@@ -30,7 +32,7 @@ export const makeSelectUserPermissions = createSelector(makeSelectUser, user =>
  * @returns {String[]} - All permissions from assigned roles combined with extra permissions
  */
 export const makeSelectUserPermissionCodeNames = createSelector(makeSelectUserPermissions, permissions =>
-  permissions.map(({ codename }) => codename)
+  permissions.map(({ codename }) => codename) ?? []
```
MauritsR, line 35: Is `?? []` necessary here?

package.json
```diff
@@ -120,9 +120,12 @@
     "@testing-library/react-hooks": "^3.4.2",
     "@testing-library/user-event": "^12.2.2",
     "@types/enzyme": "^3.10.8",
+    "@types/invariant": "^2.2.34",
     "@types/jest": "^26.0.19",
     "@types/jsdom": "^16.2.5",
     "@types/leaflet": "^1.5.19",
+    "@types/lodash": "^4.14.167",
+    "@types/lodash.clonedeep": "^4.5.6",
```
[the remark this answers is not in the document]
aditudorache: Yes, lodash.clonedeep is an add-on with separate types

src/containers/App/actions.ts
```diff
@@ -17,28 +15,29 @@ import {
   GET_SOURCES_FAILED,
   GET_SOURCES_SUCCESS,
 } from './constants';
+import type { GlobalNotification, UserCredentials, User } from './types';
 
-export const loginFailed = payload => ({
+export const loginFailed = (payload: string) => ({
```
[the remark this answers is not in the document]
aditudorache: Great suggestion, I wanted that but I thought was too much extra work. I'll give it a try!

src/containers/App/selectors.ts
aditudorache: removed

src/containers/App/reducer.ts
```diff
@@ -0,0 +1,136 @@
...
+    message: '',
+    title: '',
+    variant: VARIANT_NOTICE,
+    type: TYPE_LOCAL,
+  },
+  searchQuery: '',
+  sources: [],
+};
+
+type ReducerActionTypes = AppActionTypes | ApplyFilterActionType | Action<null, never>;
```
MauritsR, line 44: What's the use for `Action<null, never>` ?

MauritsR (approved):
Looks good!

src/containers/App/reducer.ts
aditudorache: This is for handling the default branch of the switch. Maybe we can find a better pattern for this. Any suggestions are welcome!

PR #1342, opened 2021-01-12
Parse markdown for all PlainText values

boris-arkenaar:
Markdown has been implemented in #1323 in case extra questions come from the backend. This gave a problem on the final step of the wizard where a link is shown to the incident just reported, in case the user is authenticated. This was done using a plain text HTML anchor tag and was, therefore, displayed verbatim by the markdown parser.

To solve this problem I thought it best to switch to markdown for all `PlainText` values altogether, regardless of where the extra questions come from. This has the following consequences:

* The `PlainText` component is cleaned up
* No more `dangerouslySetInnerHTML` is being used in the `PlainText` component
* Also for the hardcoded extra questions, the `PlainText` values are now easier to read and write
* The final step needed a link only in case the user is authenticated, so I introduced `valueAuthenticated` beside the already existing `value` for the `PlainText` component, to be able to provide a different value in case the user is authenticated


MauritsR (changes requested):
Good job! Looks like the definition components can also be removed in this PR, if they're not being used anymore.

boris-arkenaar:
@MauritsR Of course, you're right. I'm gonna look into that.

aditudorache:
Great work! If you would remove the components from `src/signals/incident/definitions/components` and their references, this can be merged!

aditudorache (approved):

MauritsR (approved):

MauritsR (approved):

PR #1345, opened 2021-01-14
(Chore) Improves the esling typescript configuration

aditudorache:
This PR 
- adds the improvements to the eslint typscript configuration found during the implementation of the 'ContainerSelect' feature. 
 - adds a default `tsconfig.json` to the plugin-eslint-amsterdam package as one is required for the typescript eslint configuration.
 - improves the test configurations in the package.json


.eslintrc.js
```diff
@@ -31,6 +31,26 @@ module.exports = {
         'promise/prefer-await-to-callbacks': 'off',
       },
     },
+    {
```
MauritsR, line 34: Shouldn't these be implemented in `typescript.js` where all other typescript eslint config is placed?

aditudorache: This requires the tsconfig.json so this is required. We can look in a later stage if we want to promote some of these settings to the plugin config.

MauritsR (approved):

PR #1356, opened 2021-01-18
ADR - replace fetch mocking with msw 

aditudorache:
This PR is a ADR change proposal for using the popular msw for mocking the fetch request in the unit tests.

### DIFF
--- a/docs/adr/0010-stop-mocking-fetch-in-tests.md
+++ b/docs/adr/0010-stop-mocking-fetch-in-tests.md
@@ -18,7 +18,7 @@ Here are few advantages of introducing this library:
 - All the fetch/mocking can moved to one place (internals/testing)
 - The tests involving fetch will be simpler and easier to read/write
 - Enforces using the correct data/types, no half mock objects
-
+- E2e tests can also benefit from the msw setup and can use fixtures, the same way the unit tests can
 
 ## Decision
 The `jest-fetch-mock` will be removed and the mocking will be done by using the the `msw` package
--- a/docs/adr/0010-stop-mocking-fetch-in-tests.md
+++ b/docs/adr/0010-stop-mocking-fetch-in-tests.md
@@ -8,7 +8,7 @@ Date: 2021-01-14
 
 ## Context
 
-For mocking unittests we are using at this moment the `jest-fetch-mock` package. This works but there are new develoments on the market. As the creator of the [Testing Library](https://testing-library.com/) suggests in this [article](https://kentcdodds.com/blog/stop-mocking-fetch), there is a new strategy available that would simplify testing of components that are making fetch calls.
+For mocking unittests we are using at this moment the `jest-fetch-mock` package. This works but there are new develoments on the market. As the creator of the [Testing Library](https://testing-library.com/) suggests in this [article](https://kentcdodds.com/blog/stop-mocking-fetch), there is a new strategy available that would simplify testing of components that are making fetch calls. In this [article](https://testing-library.com/docs/react-testing-library/example-intro/) is recomanded to use `msw` for fetch mocking.
 
 ## Technical description
 The `msw` server intercepts the fetch calls and returns the configured response. This is done outside the test file and only overrides of the standard responses from the default configuration need to be added to the respective test file. [Here is the documentation](https://github.com/mswjs/msw).
--- a/docs/adr/0010-stop-mocking-fetch-in-tests.md
+++ b/docs/adr/0010-stop-mocking-fetch-in-tests.md
@@ -1,0 +1,36 @@
+# Replace mocking fetch with `msw`
+
+Date: 2021-01-14
+
+## Status
+
+2021-01-18 Proposed
+
+## Context
+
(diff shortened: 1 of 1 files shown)


docs/adr/0010-stop-mocking-fetch-in-tests.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Technical description
+The `msw` server intercepts the fetch calls and returns the configured response. This is done outside the test file and only overrides of the standard responses from the default configuration need to be added to the respective test file. [Here is the documentation](https://github.com/mswjs/msw).
+
+## Consideratons
+Here are few advantages of introducing this library:
+- All the fetch/mocking can moved to one place (internals/testing)
+- The tests involving fetch will be simpler and easier to read/write
+- Enforces using the correct data/types, no half mock objects
+
```
janjaap, line 21: Plus, the e2e tests can also benefit from the msw setup and can use fixtures, the same way the unit tests can.

boris-arkenaar (approved):

docs/adr/0010-stop-mocking-fetch-in-tests.md
```diff
@@ -0,0 +1,36 @@
...
+## Context
+
+For mocking unittests we are using at this moment the `jest-fetch-mock` package. This works but there are new develoments on the market. As the creator of the [Testing Library](https://testing-library.com/) suggests in this [article](https://kentcdodds.com/blog/stop-mocking-fetch), there is a new strategy available that would simplify testing of components that are making fetch calls.
+
+## Technical description
+The `msw` server intercepts the fetch calls and returns the configured response. This is done outside the test file and only overrides of the standard responses from the default configuration need to be added to the respective test file. [Here is the documentation](https://github.com/mswjs/msw).
+
+## Consideratons
+Here are few advantages of introducing this library:
+- All the fetch/mocking can moved to one place (internals/testing)
```
MauritsR, line 18: What would this look like in tests that require different responses to the same call? Magic input that returns a specific response?

MauritsR: Random example I found: 

`fetch.mockResponses([JSON.stringify({ detail: 'filled out' }), { status: 410 }]);`

This is already pretty succinct - how would this be changed with `msw`?

docs/adr/0010-stop-mocking-fetch-in-tests.md
```diff
@@ -0,0 +1,36 @@
...
+- The tests involving fetch will be simpler and easier to read/write
+- Enforces using the correct data/types, no half mock objects
```
MauritsR, line 20: Typing the mocks, does that mean the fixtures itself will become .ts files?

docs/adr/0010-stop-mocking-fetch-in-tests.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Refactoring plan
+- Add the `msw` package and create the configuration in the `internals folder`
+- Use for the configuration the existing features that are already present in `utils/__tests__/fixtures`
+- Start small, replace the `jest-fetch-mock` only in the changed files.
+
+
+## Consequences
+
+- There will be a transition period where both `jest-fetch-mock` and `msw` will coexist.
```
MauritsR, line 35: I'm a bit worried about adding yet another tool to run alongside a similar one. 

We're already in a transition phase for 
- JavaScript -> TypeScript
- naming conventions
- form handling
- usage of hooks/redux

If we have a running example, we might be able to see if we can implement it in the entire app at once

docs/adr/0010-stop-mocking-fetch-in-tests.md
janjaap: See https://kentcdodds.com/blog/stop-mocking-fetch

aditudorache: Will workout a poc asap so we solve this kind of scenarios

docs/adr/0010-stop-mocking-fetch-in-tests.md
aditudorache: No, they will stay .json but can be typechecked.

docs/adr/0010-stop-mocking-fetch-in-tests.md
aditudorache: At atlas, they have already solved some of the challenges we might have. The aim is to migrate gradually as files are touched, will have to set it up though.

docs/adr/0010-stop-mocking-fetch-in-tests.md
MauritsR: Looking forward to it! Would like to doublecheck the advantages list and see how this list holds up (and if we cannot simply do it with the current solution as well)

aditudorache:
Although it looks promising, there are some issues I couldn't yet solve. There seem to be some async issues when running the tests in our project. Please see here the progress I made: https://github.com/Amsterdam/signals-frontend/tree/SIG-3485-add-msw-for-testing.
We might consider drop this proposal when no easy solution can be found soon.

MauritsR (approved):

PR #1359, opened 2021-01-19
(Chore) Refactor wfs layer

MauritsR:
### Changes

- Refactors wfs-layer

Some props for the WfsLayer were computed based on a dependency it itself uses as well. I moved that part to the WfsLayer itself.

### DIFF
--- a/src/signals/incident/components/form/ContainerSelect/WfsLayer/WfsLayer.tsx
+++ b/src/signals/incident/components/form/ContainerSelect/WfsLayer/WfsLayer.tsx
@@ -1,27 +1,43 @@
-import type { FunctionComponent } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { useMapInstance } from '@amsterdam/react-maps';
 import type { FeatureCollection } from 'geojson';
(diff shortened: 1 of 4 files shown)


src/signals/incident/components/form/ContainerSelect/WfsLayer/WfsLayer.tsx
```diff
@@ -1,51 +1,67 @@
...
 };
 
-export interface Props extends WfsLayerProps {
+export interface WfsLayerProps {
   children: React.ReactElement<DataLayerProps>;
+  zoomLevel?: ZoomLevel;
 }
 
-const WfsLayer: FunctionComponent<Props> = ({ url, options, zoomLevel, children }) => {
+const WfsLayer: React.FC<WfsLayerProps> = ({ children, zoomLevel = DEFAULT_ZOOM_LEVEL }) => {
```
aditudorache, line 29: I would prefer to keep using FunctionComponent here as this is the current trend.

aditudorache (approved):
As discussed, this can be used for the moment but will be need changing in the future as the BBox format can vary between different wfs implementations

PR #1360, opened 2021-01-21
[SIG-3337] Select container item in the map

aditudorache:
This PR 
- adds the container selection functionality for the `ContainerSelect` component.
- adds types for wfs features
- some parts needed to be excluded from coverage due to testing limitations of leaflet. 

`Note`: After selecting a item in a cluster, the cluster closes. A solution for keeping the cluster open will be added in another PR.


src/signals/incident/components/form/ContainerSelect/ContainerLayer/ContainerLayer.tsx
```diff
@@ -32,49 +34,86 @@ export const ContainerLayer: FunctionComponent<DataLayerProps> = ({ featureTypes
...
-  const options: L.GeoJSONOptions<Record<string, string | undefined>> = useMemo(
+  type GeoJSONOptions = {
+    pointToLayer: (feature: Feature, latlng: LatLng) => L.Layer;
+  };
+  const options = useMemo<GeoJSONOptions>(
     () => ({
       pointToLayer: (feature, latlng: LatLng) => {
         const featureType = getFeatureType(feature);
         if (!featureType) return L.marker({ ...latlng, lat: 0, lng: 0 });
+        const selected = selection.some(
```
MauritsR, line 49: After panning the map I get an error: 'TypeError: selection.some is not a function'

src/signals/incident/components/form/ContainerSelect/ContainerLayer/ContainerLayer.tsx
```diff
@@ -32,49 +34,86 @@ export const ContainerLayer: FunctionComponent<DataLayerProps> = ({ featureTypes
...
+            iconUrl,
           }),
           // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
           alt: feature.properties[featureType.idField],
         });
+
+        marker.on(
+          'click',
+          /* istanbul ignore next */ () => {
+            const { description, typeValue, idField }: Partial<FeatureType> = featureType;
```
MauritsR, line 72: Why use `Partial`?

src/signals/incident/components/form/ContainerSelect/ContainerLayer/ContainerLayer.tsx
```diff
@@ -32,49 +34,86 @@ export const ContainerLayer: FunctionComponent<DataLayerProps> = ({ featureTypes
...
+            const item = {
```
MauritsR, line 73: Better to do something like 

`const item: Item = {`

and then cast the id field to string (turn off _non-nullable-type-assertion-style_?). That makes it more clear what part of the object needs to be casted.

It is also possible (maybe even better) to check if _id_ is defined.

MauritsR (changes requested):

.eslintrc.js
```diff
@@ -50,6 +50,7 @@ module.exports = {
           },
         ],
         '@typescript-eslint/no-extra-parens': 'off',
+        '@typescript-eslint/consistent-type-definitions': 'off',
```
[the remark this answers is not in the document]
aditudorache: Because it forces me to uses interfaces at places where I want to use types.

src/signals/incident/components/form/ContainerSelect/ContainerLayer/ContainerLayer.tsx
aditudorache: Good point! I'm not against turning of this rule, let's have a chat about that.

src/signals/incident/components/form/ContainerSelect/ContainerLayer/ContainerLayer.tsx
aditudorache: Is indeed not needed, type inference works good here.

MauritsR (approved):

PR #1362, opened 2021-01-21
[SIG-3484] Remove icon url on save

aditudorache:
This PR implements the configuration of the data that can be posted to the api in the extra questions. The only accepted format for the arrays is `string` or object of type:
```typescript
 { id: string, type: string, description: string }
```
This format is configurable in the metadata definition of the question


src/signals/incident/services/convert-value/index.ts
```diff
@@ -0,0 +1,27 @@
+import { isArray } from 'utils';
+
+const convertValue = (value: unknown = undefined, postPropertyNames: string[] = []): any => {
```
MauritsR, line 3: The returnType of this function is also `unknown`

MauritsR: Not sure if defaulting to undefined is useful, since by definition a prop that is not passed is undefined

src/signals/incident/services/convert-value/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (value === true) {
+    return 'ja';
+  }
+
+  if (value === false) {
+    return 'nee';
+  }
+
+  // when property names are specified, create an array with objects with only those properties
+  if (isArray(value) && postPropertyNames.length) {
```
MauritsR, line 17: Can this logic be implemented in a different utility function? It's not really _converting_ a value but it is filtering allowed properties

MauritsR, line 17: Can't we use `Array.isArray`? That would also remove the need to cast to an array below

MauritsR: It's preferred to use `unknown` instead of `any`

src/signals/incident/definitions/wizard-step-2-vulaan/afval.ts
```diff
@@ -52,6 +52,7 @@ export const controls = {
       label: 'Kies de container waar het om gaat',
       shortLabel: 'Container(s)',
       pathMerge: 'extra_properties',
+      postPropertyNames: ['id', 'type', 'description'],
```
MauritsR, line 55: Something like `allowedPropertyNames` would me more descriptive (same below in the utility function).

MauritsR: When should properties be added to the `additional-questions.md` docs?

MauritsR (changes requested):

MauritsR:
Is it not possible to add only the required data to the `addContainer` function and then (when icons need to be rendered) look up the icon url in the `meta` object when required?

aditudorache:
Great idea! I'll remove th iconUrl from the item and will lookup the icon in when it is used!

aditudorache:
Closed in favor of #1366

PR #1363, opened 2021-01-21
(Chore) Fix call in saga

aditudorache:
This PR replaces an unnecessary saga call with a synchronous normal function call.


MauritsR (approved):
Looks good, but has a failing test

PR #1367, opened 2021-01-25
[SIG-3339] Container map drawer

MauritsR:
This PR fixes UX and styling based on a review by our designer

### Fixes

- Map drawer should be open by default
- Map drawer subtitle should be shown as a paragraph in the content area
- Selection submit button should be active always
- Remove vertical scroll on map
- Fix map buttons shadow and font
- Add slide-in animation to legend
- Prevent icons shrinking when labels in icon list are long


aditudorache (approved):
Nice work!

PR #1368, opened 2021-01-26
[SIG-3462] Removes the ADW/Grip button references

aditudorache:
This PR removes the ADW button and references. The existing `Inloggen Keycloak` button is renamed to `Inloggen ADW`


MauritsR (approved):
Nice!

PR #1378, opened 2021-01-29
(Chore) Restructure container select according to the ADR

aditudorache:
This PR restructures the files inside the ContainerSelect component according to the Naming conventions ADR (#1291) and converts `Intro` and `Summary` components to typescript.


MauritsR (approved):
Looks good 👍

PR #1382, opened 2021-02-02
(Chore) Fix Docker build

MauritsR:
### Fixes

- Fix Docker build

Since the file handling application configuration was converted to TypeScript, the example file from which the config type is inferred needs to be present during build time. 

This PR fixes the deploy (and e2e tests).


boris-arkenaar (approved):
🙏🏼

PR #1383, opened 2021-02-02
[SIG-3433] Improve message visibility

MauritsR:
### Changes

- Improves info message visibility when closing incident with open split incident(s)


aditudorache (approved):
👍🏼

PR #1384, opened 2021-02-02
[FE#102] Source in saved filter

boris-arkenaar:
closes Signalen/frontend#102

- The selected `source` was not maintained in a saved filter.
- The same goes for `area` in case the feature flag `fetchDistrictsFromBackend` is turned on. This still requires a change in the backend as well, though. (Signalen/backend#111)

### DIFF
--- a/src/signals/incident-management/selectors.js
+++ b/src/signals/incident-management/selectors.js
@@ -29,8 +29,15 @@ export const makeSelectDistricts = createSelector([selectIncidentManagementDomai
 );
 
 export const makeSelectAllFilters = createSelector(
-  [selectIncidentManagementDomain, makeSelectMainCategories, makeSelectSubCategories, makeSelectDirectingDepartments],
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/selectors.js
```diff
@@ -28,9 +28,26 @@ export const makeSelectDistricts = createSelector([selectIncidentManagementDomai
     : null
 );
 
+export const selectFixtures = createSelector(
```
aditudorache, line 31: Please keep the selectors naming convention here: `makeSelectFixtures`.

aditudorache (changes requested):

src/signals/incident-management/selectors.js
boris-arkenaar: Ah, yes. I'll make it consistent.
I find it a very strange way of naming, though. 😅

aditudorache (approved):

MauritsR (approved):

PR #1385, opened 2021-02-03
[SIG-3380] Show containers in back office detail list

MauritsR:
### Changes

- Adds container list to extra properties in incident detail component


src/signals/incident-management/containers/IncidentDetail/components/Detail/components/ExtraProperties/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  if (Array.isArray(answer)) {
+    return answer.map((item, index) => {
+      if (typeof item !== 'string') {
+        if (item.description) {
+          return <div key={item.id}>{`${item.description}${item.id && ` - ${item.id}`}`}</div>;
+        }
+      }
+
+      // eslint-disable-next-line react/no-array-index-key
+      return <div key={index}>{item}</div>;
```
aditudorache, line 15: It is not recommended to use index as key in react lists. If you would use the item as a key here, the eslint-disable won't be needed.

src/signals/incident-management/containers/IncidentDetail/components/Detail/components/ExtraProperties/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+const ExtraProperties: FunctionComponent<ExtraPropertiesProps> = ({ items = [] }) => {
+  // Some incidents have been stored with values for their extra properties that is incompatible with the current API
+  // We therefore need to check if we're getting an array or an object
+  const itemList = Array.isArray(items)
+    ? items
+    : Object.entries(items).map(([question, answer], index) => ({ label: question, answer, id: index.toString() }));
+
+  const extraProperties: JSX.Element[] = [];
+
+  itemList.forEach((item: Item | MappedLegacyItem) => {
```
aditudorache, line 43: Try not to use forEach as is not immutable. `map` would be a better choice here.

aditudorache (changes requested):
Nice work! A couple of suggestions though.

src/signals/incident-management/containers/IncidentDetail/components/Detail/components/ExtraProperties/index.tsx
MauritsR: Initially I had map, but it was difficult to get TypeScript to work well with that solution (Union types can be a pain in a `.map` function). Looked at it again and found a way to use map cleanly 👍

aditudorache (approved):

PR #1386, opened 2021-02-04
[SIG-3479] Migrate cypress tests to typescript

arjandata:


aditudorache (approved):
Nice work!

e2e-tests/cypress/support/commandsGeneral.ts
```diff
@@ -0,0 +1,17 @@
+Cypress.Commands.add('setResolution', size => {
```
MauritsR, line 1: Can you define these functions (and those in other similar files) separately? Then it's possible to type them - right now `size` is an implicit `any` instead of a `Cypress.ViewportPreset.`

e2e-tests/cypress/support/index.d.ts
```diff
@@ -0,0 +1,250 @@
+declare namespace Cypress {
+  interface Chainable {
+
+    /**
+     * Custom command to add a note to a signal.
+     * @example cy.addnote('This is a note');
+    */
+    addNote: (noteText: string) => Chainable<Element>;
```
MauritsR, line 8: Are these really chainable? That would mean you could do `cy.addNote('Foo').addNote('Bar')`.

e2e-tests/cypress/support/commandsGeneral.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (Cypress._.isArray(size)) {
+    cy.viewport(size[0], size[1]);
+  } else {
+    cy.viewport(size);
```
MauritsR, line 5: Returning `cy.viewport(size);` makes the function chainable (see other comment in `index.d.ts` file)

e2e-tests/cypress/support/index.d.ts
MauritsR: What do you think about doing these declarations in the files themselves? 

Then you could do the following:

```   
   /**
     * Custom command to add a note to a signal.
     * @example cy.addnote('This is a note');
    */
    addNote: typeof addNote
```

And then import that file in this index.d.ts file. The benefit is having both the type, implementation and documentation colocated.

MauritsR (changes requested):

PR #1388, opened 2021-02-05
[SIG-3340] Add container not visible in map

MauritsR:
### Changes

- Allow adding a container that is not visible on the map


src/signals/incident/definitions/wizard-step-2-vulaan/afval.constants.ts
```diff
@@ -0,0 +1 @@
+export const CONTAINER_NOT_ON_MAP_TYPE_NAME = 'not-on-map';
```
aditudorache, line 1: I would place this in `...ContainerSelect/constants.ts`  the afval configuration is to be moved to the backend.

aditudorache, line 1: CONTAINER_NOT_ON_MAP could be UNREGISTERED_CONTAINER

aditudorache (changes requested):
Great work, one suggestion: The reason the containers not on map is because they are unregistered so lets rename `NotOnMapContainer` with something like 'UnregisteredContainer'.

src/signals/incident/components/form/ContainerSelect/Selector/SelectionPanel/SelectionPanel.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+      {selectionOnMap.length ? (
+        <StyledContainerList selection={selectionOnMap} onRemove={removeContainer} featureTypes={featureTypes} />
+      ) : (
+        <EmptySelectionWrapper>
+          <StyledParagraph>Maak een keuze op de kaart</StyledParagraph>
+        </EmptySelectionWrapper>
+      )}
+
+      <form>
```
aditudorache, line 116: Is this needed?

MauritsR: Not necessarily. I changed it so that the 'enter' key behavior closes the panel without needing a form.

aditudorache (approved):

PR #1389, opened 2021-02-08
[SIG-3414] Add Zoom message

aditudorache:
This PR adds a zoom message when the containers are not visible due to the current zoom level.


src/signals/incident/components/form/ContainerSelect/Selector/ZoomMessage/ZoomMessage.tsx
```diff
@@ -0,0 +1,54 @@
...
+  @media screen and ${breakpoint('max-width', 'tabletM')} {
+    left: 0;
+    width: 100%;
+    margin: 0;
+  }
+`;
+
+export interface ZoomMessageProps {
+  showZoomMessage: boolean;
+  setShowZoomMessage: (show: boolean) => void;
```
MauritsR, line 29: I'd simplify the props (e.g. only a maxZoomLevel prop). The component can then handle it's own state. 

Better yet, all logic in this component can be extracted to a hook (e.g. useIsLayerWithinZoomBoundaries)

MauritsR (changes requested):

src/signals/incident/components/form/ContainerSelect/Selector/Selector.tsx
```diff
@@ -89,6 +96,28 @@ const EmptySelectionWrapper = styled.div`
   margin: ${themeSpacing(4)} 0;
 `;
 
+const ButtonBarStyle = styled.div<{ layerVisible: boolean }>`
+  @media screen and ${breakpoint('max-width', 'tabletM')} {
```
MauritsR, line 100: This should be 

`  @media screen and (${breakpoint('min-width', 'tabletM')}) {`

Or it won't be triggerred on Mac (also, I get a tslint error for it: `( expected ts-styled-plugin(9999)`)

MauritsR (changes requested):
Looks good! Left one last remark.

src/signals/incident/components/form/ContainerSelect/Selector/Selector.tsx
aditudorache: Good catch! Added brackets! I get that the ts error as well. Not found yet a way to suppress it. It looks that with brackets is gone :-)

PR #1394, opened 2021-02-10
[SIG-3538] Allow container map navigation with one finger on touch devices 

MauritsR:
### Changes 

- Adds posibility to navigate container map with one finger


aditudorache (approved):
Seems to work on mobile (Android), but in Chrome DevTools I can drag the map in the first page of the wizard which is not intended.

src/shared/services/configuration/map-options.ts
```diff
@@ -4,10 +4,11 @@ import type { MapOptions } from 'leaflet';
 import configuration from 'shared/services/configuration/configuration';
 
 const MAP_OPTIONS: MapOptions = {
-  ...(configuration?.map.options as MapOptions),
+  ...(configuration.map.options as MapOptions),
   attributionControl: true,
   crs: getCrsRd(),
   zoomControl: false,
+  dragging: !('ontouchstart' in window),
```
aditudorache, line 11: Maybe an comment here that this works on mobile but might not work on DevTools emulator?

PR #1395, opened 2021-02-10
[FE#105] Filter users on department through endpoint

boris-arkenaar:
fixes Signalen/frontend#105

depends on Amsterdam/signals#714

**Problem**
For the form field of 'assigned user' of an incident, the users were filtered on department in the FE. This was based on a list of users fetched from the users endpoint. This is not a full list of all users but the first 'page' of 100 users. This gave problems.

**Solution**
The users are now filtered on the BE by means of a request to the user endpoint with the `profile_department_code` filter param.

### DIFF
--- a/internals/mocks/fixtures/departments.json
+++ b/internals/mocks/fixtures/departments.json
@@ -1,0 +1,409 @@
+{
+  "_links": {
+    "self": {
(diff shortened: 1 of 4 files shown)


aditudorache (approved):
Nice addition to the msw server!

MauritsR (approved):

PR #1399, opened 2021-02-11
[SIG-3434] Change text to indicate parent/child/default incident

MauritsR:
### Changes

- In the incident detail header, change the text to be standaard/hoofd/deel melding
- In the incident detail view, add a link to the parent incident


aditudorache (approved):
Great work!! Maybe porting to ts in the next iteration?

PR #1413, opened 2021-02-22
[SIG-3546] Show error when wfs call fails

aditudorache:
This PR shows a message in the container selector when the wfs service is not available


src/signals/incident/components/form/ContainerSelect/types.ts
```diff
@@ -38,9 +38,11 @@ export interface ContainerSelectValue {
   selection: Item[];
   location: LatLngExpression;
   meta: Meta;
+  message?: string;
   update: (items: Item[]) => void;
   edit: ClickEventHandler;
   close: () => void;
+  setMessage: (message: string | undefined) => void;
```
MauritsR, line 45: Can also be written as `setMessage: (message?: string) => void;`

MauritsR (changes requested):

src/signals/incident/components/form/ContainerSelect/context.tsx
```diff
@@ -6,9 +6,11 @@ export const initialValue: ContainerSelectValue = {
   selection: [],
   location: [0, 0],
   meta: { endpoint: '', featureTypes: [] },
+  message: undefined,
   update: /* istanbul ignore next */ () => {},
   edit: /* istanbul ignore next */ () => {},
   close: /* istanbul ignore next */ () => {},
+  setMessage: /* istanbul ignore next */ () => {},
 };
```
MauritsR, line 14: Not part of this PR, but we're providing a default context value that should never be used (empty functions). Instead, we could also write a `useContainerSelectContext` hook that throws an error when 'value' is not supplied. Example here: https://kentcdodds.com/blog/how-to-use-react-context-effectively#the-custom-consumer-hook

src/signals/incident/components/form/ContainerSelect/Selector/Selector.tsx
```diff
@@ -78,24 +78,19 @@ const StyledMap = styled(Map)`
   }
 `;
 
-// Added also to `arm` in #1108. Will be removed here when the next version of `arm` is released
-const StyledMapPanelDrawer = styled(MapPanelDrawer)`
```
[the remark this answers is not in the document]
aditudorache: Good catch! Thanks!

src/signals/incident/components/form/ContainerSelect/context.tsx
aditudorache: Great suggestion! I had exactly this in mind for some time now :-)

MauritsR (approved):
Looks good :)

PR #1414, opened 2021-02-22
[SIG-3565] Next incident button

MauritsR:
### Changes

- Adds button to final incident step to quickly navigate to a new incident form


src/signals/incident/components/form/HandlingMessage/__snapshots__/index.test.js.snap
```diff
@@ -2,43 +2,35 @@
 
 exports[`Form component <HandlingMessage /> rendering should render empty correctly with no handling message 1`] = `
 <div
-  className="col-12 mode_input"
+  className="col-12"
```
aditudorache, line 5: Can we delete this snap and replace the tests with testing-library?

aditudorache (changes requested):
Nice! If we can remove the old snapshot would be great!

src/signals/incident/components/form/HandlingMessage/__snapshots__/index.test.js.snap
MauritsR: Fixed it :) now in TypeScript and without the snapshots (using testing-library)

aditudorache (approved):

aditudorache:
Approved! Some tests still failing...

PR #1429, opened 2021-03-01
[SIG-3442] WCAG html textarea 

aditudorache:
This PR
- adds aria links between the textarea element and its description in the main incident form and the extra questions
- renames the Label component to match the file name
- renames the helpText property of the TextArea to infoText which is more descriptive for this property
- adds a fieldset inside the incident form; this fieldset contains of the form.

### DIFF
--- a/src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
+++ b/src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
@@ -30,8 +30,8 @@ const DescriptionInput = ({
 
   return (
     <TextArea
-      id={meta.id}
-      aria-describedby={`subtitle-${meta.id}`}
+      id={meta.name}
+      aria-describedby={`subtitle-${meta.name}`}
       data-testid="descriptionInput"
       rows={meta.rows}
       placeholder={meta.placeholder}
--- a/src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
+++ b/src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
@@ -30,12 +30,13 @@ const DescriptionInput = ({
 
   return (
     <TextArea
+      id={meta.id}
       data-testid="descriptionInput"
-      rows={meta.rows || 6}
+      rows={meta.rows}
       placeholder={meta.placeholder}
       {...handler()}
       onBlur={handleBlur}
-      helpText={<DescriptionInfo info={getCharactersInfo()} />}
+      infoText={<DescriptionInfo info={getCharactersInfo()} />}
     />
   );
 };
--- a/src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
+++ b/src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
@@ -29,16 +29,19 @@ const DescriptionInput = ({
   }, [meta, parent]);
 
   return (
-    <TextArea
-      id={meta.name}
-      aria-describedby={`subtitle-${meta.name}`}
-      data-testid="descriptionInput"
-      rows={meta.rows}
-      placeholder={meta.placeholder}
-      {...handler()}
-      onBlur={handleBlur}
-      infoText={<DescriptionInfo info={getCharactersInfo()} />}
-    />
+    <div>
+
+      <TextArea
+        id={meta.name}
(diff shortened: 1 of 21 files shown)


src/signals/incident/components/form/TextareaInput/index.js
```diff
@@ -8,13 +8,16 @@ const TextareaInput = ({ handler, touched, value, hasError, meta, parent, getErr
   meta?.isVisible && (
     <Header meta={meta} options={validatorsOrOpts} touched={touched} hasError={hasError} getError={getError}>
       <TextArea
+        id={meta.id}
```
MauritsR, line 11: Instead of assigning an id to every input, we could also use the `meta.name` property which is derived from the key of the question in the form generator. For example, the 'description' question in wizard-step-1 already passes `meta: {name: 'description'}` to the TextArea component. 

This would also prevent issues where someone forgets to add the 'id' field to the meta key of a question.

src/signals/incident/components/form/Header/index.js
```diff
@@ -47,14 +38,14 @@ const Header = ({ className, meta, options, touched, hasError, getError, childre
...
-        <Label>
+        <Label htmlFor={meta.id}>
           {meta.label}
 
           {isOptional && <Optional>(optioneel)</Optional>}
         </Label>
       )}
 
-      {meta?.subtitle && <SubTitle>{meta.subtitle}</SubTitle>}
+      {meta?.subtitle && <SubTitle id={`subtitle-${meta.id}`}>{meta.subtitle}</SubTitle>}
```
MauritsR, line 48: Better to double check that the id is passed to every component (also see comment below). Right now this results in `id="subtitle-undefined"` in the location map.

MauritsR (changes requested):

src/signals/incident/components/form/TextareaInput/index.js
aditudorache: Good suggestion! I'll use that!

src/signals/incident/components/form/DescriptionInput/DescriptionInput.js
```diff
@@ -29,14 +29,19 @@ const DescriptionInput = ({
...
-    <TextArea
-      data-testid="descriptionInput"
-      rows={meta.rows || 6}
-      placeholder={meta.placeholder}
-      {...handler()}
-      onBlur={handleBlur}
-      helpText={<DescriptionInfo info={getCharactersInfo()} />}
-    />
+    <div>
+
```
MauritsR, line 33: whitespace can be removed

src/signals/incident/components/form/Header/index.js
```diff
@@ -2,41 +2,48 @@ import React, { Fragment } from 'react';
...
 import styled, { css } from 'styled-components';
 import { themeSpacing, themeColor } from '@amsterdam/asc-ui';
+import Label from 'components/Label';
 
-const Children = styled.div`
+const Wrapper = styled.div`
   display: flex;
-  flex-flow: column;
-`;
+  flex-direction:column;
```
MauritsR, line 9: Space should be inserted after `flex-direction:`

src/signals/incident/components/form/Header/index.js
```diff
@@ -2,41 +2,48 @@ import React, { Fragment } from 'react';
...
-const Wrapper = styled.div`
   ${({ invalid }) =>
     invalid &&
     css`
       border-left: ${themeColor('support', 'invalid')} 2px solid;
       padding-left: ${themeSpacing(3)};
     `}
+
+  /* keep a 3 units margin above the last element/control  */
+  & > :last-child :not(& > :first-child) {
```
MauritsR, line 19: This should be improved: on the incident page of the "type" field (logged in), this increases the margin between between the type explanation and the radio buttons.

MauritsR (changes requested):
Added a couple more comments (styling related)

src/signals/incident/components/form/Header/index.js
aditudorache: Solved by grouping the radio buttons and the description. We assume the last element is the container with the space on top.

MauritsR (approved):

PR #1430, opened 2021-03-01
[SIG-3312] tab focus order

MauritsR:
### Changes

- Improve tabbing order of form elements in incident form
- Prevent location marker from being focusable
- Add focus outline to GPS button

### DIFF
--- a/src/components/GPSButton/index.js
+++ b/src/components/GPSButton/index.js
@@ -9,7 +9,7 @@ import LoadingIndicator from 'components/LoadingIndicator';
 import GPS from '../../shared/images/icon-gps.svg';
 
 const StyledButton = styled(Button)`
-  box-shadow: 0 0 0 2px rgb(0 0 0 / 10%);
+  box-shadow: 0 0 0 2px rgb(0, 0, 0, 0.1);
 `;
 
 const GPSIcon = styled(GPS)`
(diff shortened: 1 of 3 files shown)


src/components/Map/Map.js
```diff
@@ -84,6 +89,9 @@ const Map = ({
       options={options}
       setInstance={captureInstance}
     >
+      {children}
+
+      {/* Render GPS and zoom buttons after children to maintain correct focus order */}
```
aditudorache, line 94: 👍🏼

src/components/GPSButton/index.js
```diff
@@ -9,7 +9,7 @@ import LoadingIndicator from 'components/LoadingIndicator';
 import GPS from '../../shared/images/icon-gps.svg';
 
 const StyledButton = styled(Button)`
-  outline: 2px solid rgb(0, 0, 0, 0.1);
+  box-shadow: 0 0 0 2px rgb(0 0 0 / 10%);
```
aditudorache, line 12: Can we use here `rgba(0, 0, 0, 0.1);` for consistency?

aditudorache (changes requested):

src/components/GPSButton/index.js
aditudorache: I mean `rgba` instead `rgb` or else the last parameter is not used

MauritsR: Changed it to `rgba` instead of `rgb` to keep consistency. 

`rgba` is an alias for `rgb` so the result stays the same :)

aditudorache: :-) didn't know dat!!

aditudorache (approved):

PR #1431, opened 2021-03-02
[SIG-3595] Prevent retrieving user list when view user permission is not set

MauritsR:
### Changes

- Prevents fetching users list when user has no permission to retrieve it
- When filtering on assigned user, fall back to regular input if user has not permission to retrieve user list

After https://github.com/Amsterdam/signals/pull/734  is merged, only users with the "view_user" permission are allowed to retrieve the user list. The filter form (assigned user) and IncidentManagementContext use this data. This PR makes sure that not having the "view_user" permission set is handled correctly.


src/signals/incident-management/components/FilterForm/__tests__/FilterForm.test.js
```diff
@@ -43,10 +44,10 @@ const jsonResponse = {
 const mockResponse = JSON.stringify(jsonResponse);
 
 const formProps = {
-  onClearFilter: () => {},
-  onSaveFilter: () => {},
+  onClearFilter: () => { },
```
aditudorache, line 47: I see a lot of text editor formatting. Is something changed?

MauritsR: Didn't run prettier formatting (only TS formatting). Fixed now.

aditudorache (approved):

PR #1447, opened 2021-03-08
[SIG-3444] Fix CheckboxInput accessibility

aditudorache:
This PR fixes the accessibility and positioning of the CheckboxInput control according to WCAG.

Note that the conversion of this control to `typescript` will be done in a separate PR.


src/signals/incident/components/form/CheckboxInput/CheckboxInput.js
```diff
@@ -0,0 +1,94 @@
...
+  display: flex;
+  flex-direction: column;
+`;
+
+const CheckboxInput = ({ handler, touched, hasError, meta, parent, getError, validatorsOrOpts }) =>
+  meta.isVisible && (
+    <Header meta={meta} options={validatorsOrOpts} touched={touched} hasError={hasError} getError={getError}>
+      <CheckboxWrapper>
+        {isObject(meta.values) ? (
+          <CheckboxGroup role="group">
```
MauritsR, line 42: The label in the Header now applies a "htmlFor" property, but the id to which it points is not set. 

```suggestion
          <CheckboxGroup role="group" id={meta.name}>
```

MauritsR (changes requested):

MauritsR (approved):

PR #1458, opened 2021-03-11
[SIG-3450] Cypress tests field validations

arjandata:
* Added scenario for field validations
* Put some texts in a separate file


e2e-tests/cypress/support/selectorsCreateSignal.ts
```diff
@@ -12,7 +12,10 @@ export const CREATE_SIGNAL = {
   dropdownMinuten: '[data-testid=selectMinutes]',
   dropdownSource: '[data-testid="source"]',
   dropdownSubcategory: '[data-testid="subcategory"]',
+  errorDateTime: ':nth-child(3) > p',
```
MauritsR, line 15: These seems fragile. Can you add a data-testid attribute to the `<Header />` component? Something like 

`<ErrorItem data-testid={`${meta.name}-required`}>`

MauritsR (changes requested):

e2e-tests/cypress/support/selectorsCreateSignal.ts
arjandata: Added data-testid

MauritsR (approved):

PR #1461, opened 2021-03-15
(Chore) Add development configuration to docker-compose

aditudorache:
This PR adds an `frontend-dev` development configuration service to the docker-compose. This enables changes in the sources when running the application in development mode. 
- Use `make start` to run the application in dev mode
- Use `docker-compose up frontend` to start the application as a container. This command will be used in the `gh-actions` as well for the `e2e-tests`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ COPY .gitignore \
 RUN npm install
 
 COPY assets /app/assets
+COPY server /app/server
 COPY internals /app/internals
 COPY src /app/src
 
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,11 @@ build: pull                         ## Build docker image
 	BUILD_ENV=$(BUILD_ENV) $(dc) build
 
 start:                              ## Run frontend
-	$(dc) run --service-ports frontend
+	$(dc) run --service-ports frontend-dev
 
 stop:                               ## Clean docker stuff
 	$(dc) down -v --remove-orphans
 
 start-e2e: build                    ## Starts the application and opens cypress
-	$(dc) up -d
+	$(dc) up -d frontend-dev
 	npm run open --prefix e2e-tests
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -32,6 +32,7 @@ COPY .gitignore \
 RUN npm install
 
 COPY assets /app/assets
+COPY server /app/server
```
MauritsR, line 35: What is this used for?

docker-compose.yml
```diff
@@ -1,6 +1,22 @@
 version: '3.8'
 
 services:
+  frontend-dev:
```
MauritsR, line 4: Maybe this can be solved in another way, since the two frontend services have almost the same configuration.

It doesn't make sense to have two frontend services in a single compose file (for example, `docker-compose up`) will fail because they both try to bind to port 3001.

docker-compose.yml
```diff
@@ -11,6 +27,8 @@ services:
       - ./app.amsterdam.json:/app.base.json
     ports:
       - 3001:80
+    depends_on:
+      - backend
```
MauritsR, line 31: What's the practical use of this setting in our case?

Makefile
```diff
@@ -15,11 +15,11 @@ build: pull                         ## Build docker image
...
 start:                              ## Run frontend
-	$(dc) run --service-ports frontend
+	$(dc) run --service-ports frontend-dev
 
 stop:                               ## Clean docker stuff
 	$(dc) down -v --remove-orphans
 
 start-e2e: build                    ## Starts the application and opens cypress
-	$(dc) up -d
+	$(dc) up -d frontend-dev
```
MauritsR, line 24: Not sure if this has the intended effect. Now, the backend/database/dex services will not be started

MauritsR (changes requested):

MauritsR (approved):

PR #1463, opened 2021-03-15
[SIG-3623] Reset dynamic extra questions state

MauritsR:
### Changes

- Fixes state handling for extra questions that are dynamically added/removed

The visibility of some extra questions is dependent other questions. Previously, answers from questions that were dynamically removed (i.e. `isVisible` was set to `false`) were kept in the incident state. This caused these answers to still show up on the summary page.


src/signals/incident/components/IncidentForm/index.test.js
```diff
@@ -300,4 +302,53 @@ describe('<IncidentForm />', () => {
       expect(next).not.toHaveBeenCalled();
     });
   });
+
+  describe('Extra questions', () => {
+    it('removes extra question data when the question is not visible anymore', () => {
```
aditudorache, line 307: Can we write the tests using the @testing-libraty?

aditudorache (changes requested):
Not sure if out of scope, but would be nice to convert the IncidentForm component to a functional component.

MauritsR:
> Not sure if out of scope, but would be nice to convert the IncidentForm component to a functional component.

Out of scope for this story, but definitely something we should pick up!

aditudorache (approved):
Great job!

PR #1464, opened 2021-03-16
[SIG-3611] Hide users menu item

aditudorache:
This PR refactors the Settings menu:
- A new `ProtectedRoute` component is introduced to handle the access rights
- The 'Instellingen' menu is shown when the user has access to any of the menu items of this menu.
- A user with only `user view rights` won't see the `Gebruikers` menu item.


src/components/ProtectedRoute/ProtectedRoute.tsx
```diff
@@ -0,0 +1,38 @@
...
+import NotFoundPage from 'components/NotFoundPage';
+import { useSelector } from 'react-redux';
+import { makeSelectUserCan, makeSelectUserCanAccess } from 'containers/App/selectors';
+
+export const NO_PAGE_ACCESS_MESSAGE = 'U heeft geen toegang tot deze pagina';
+
+interface ProtectedRouteProps extends RouteProps {
+  role?: string;
+  roleGroup?: string;
+  component?: any;
```
MauritsR, line 15: Can be typed like
```suggestion
  component: (props: RouteComponentProps<any>) => React.ReactNode;
```

src/containers/App/selectors.ts
```diff
@@ -33,6 +35,16 @@ export const makeSelectUserPermissions = createSelector(makeSelectUser, (user: P
  */
 export const makeSelectUserPermissionCodeNames = createSelector(makeSelectUserPermissions, permissions =>
   permissions.map(({ codename }) => codename)
+  // [
```
MauritsR, line 38: To be removed?

MauritsR (approved):
Looks good! I did add two minor comments.

PR #1465, opened 2021-03-16
Show an error above submit button on form errors

boris-arkenaar:
fixes Signalen/frontend#111

Couldn't find out how to work with `react-reactive-form` to easily scroll to the top of the page or to the input field that has an error. I worked around that by displaying an error message at the bottom of the form, directly above the submit button. It's not perfect, but it's better. I hope good enough (for now). 😛 

Maybe someone wants to think about the exact error message to show to the user.

Screenshot 2021-03-16 at 08 21 04


src/signals/incident/definitions/wizard-step-2-vulaan.js
```diff
@@ -45,6 +45,12 @@ const expandQuestions = memoize(
         }),
         {}
       ),
+      error: {
+        meta: {
+          label: 'Nog niet alles is correct ingevuld. Volg de meldingen in het rood. Scroll eventueel weer naar boven.',
```
MauritsR, line 50: To skip this step when no additional questions are required:

```suggestion
          label: 'Nog niet alles is correct ingevuld. Volg de meldingen in het rood. Scroll eventueel weer naar boven.',
          ignoreVisibility: true,
```

MauritsR (commented):
Looked into it, and to scroll to top you could do the following in the `handleSubmit` function of `IncidentForm.js`:

```
if (!this.form.valid) {
  window.scrollTo({ top: 0, behavior: 'smooth' });
}
```

boris-arkenaar:
Scroll to top on error + global error message on top of form (step 1 and step 2).
On the KTO form scrolling to the field that errors directly (no global error message).

Screenshot 2021-03-18 at 08 35 29

src/signals/incident/components/form/Error/Error.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Header from '../Header';
+import { InvalidatedProjectKind } from 'typescript';
```
MauritsR, line 5: Unused import

src/signals/incident/components/form/Error/Error.js
```diff
@@ -0,0 +1,23 @@
...
+
+const Error = ({ meta, parent: { touched, valid } }) => {
+  const hasError = errorType => errorType === 'global' && !valid;
+  const getError = errorType => (errorType === 'global' && meta.label) || true;
+  return <Header meta={{ name: 'global' }} touched={touched} hasError={hasError} getError={getError} />;
```
MauritsR, line 10: Do we really need the Header component here? How about simply returning a <Paragraph /> with text "Er zijn vragen niet (of niet juist) ingevuld. Vul de vragen hieronder op de goede manier in."? And perhaps a rename of the component to GlobalError?

MauritsR (changes requested):

src/signals/incident/components/form/Error/Error.js
boris-arkenaar: Makes sense actually. Since not much of the `Header` is being used anymore and even exceptions have been added.

aditudorache:
It seems there are some unit tests failing, can you check that @boris-arkenaar?

MauritsR (approved):

aditudorache (approved):

PR #1466, opened 2021-03-17
[SIG-3446] Improve a11y for the DateTimeInput control

aditudorache:
This PR links the description of the hour and minute to the select lists. It also removes the link from the Time label to the hours select control

### DIFF
--- a/src/signals/incident/components/form/DateTimeInput/index.js
+++ b/src/signals/incident/components/form/DateTimeInput/index.js
@@ -31,8 +31,18 @@ const FieldWrapper = styled.div`
   }
 `;
 
-const TimeWrapper = styled.div`
+const TimeFieldset = styled.fieldset`
+  border: 0;
+  padding: 0;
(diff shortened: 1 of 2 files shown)


src/signals/incident/components/form/DateTimeInput/index.js
```diff
@@ -86,21 +86,23 @@ const DateTimeInput = ({ touched, hasError, meta, parent, getError, validatorsOr
 
         <TimeWrapper>
           <div>
-            <StyledLabel htmlFor="hours">Tijd</StyledLabel>
+            <StyledLabel>Tijd</StyledLabel>
```
MauritsR, line 89: Wrapping `uur` and `min` in a `fieldset` with `legend` (Tijd) would associate the `Tijd` label better with the input components. What do you think?

MauritsR (changes requested):

MauritsR (approved):
Looks good :)

PR #1468, opened 2021-03-17
[SIG-3566] - Add filter for punctuality

MauritsR:
### Changes

- Adds a filter for punctuality to the incident management view

Screenshot 2021-03-17 at 16 56 05

### DIFF
--- a/src/shared/types/index.js
+++ b/src/shared/types/index.js
@@ -66,6 +66,7 @@ export const apiFilterType = PropTypes.shape({
     created_after: PropTypes.string,
     created_before: PropTypes.string,
     feedback: PropTypes.string,
+    punctuality: PropTypes.string,
     maincategory_slug: PropTypes.arrayOf(PropTypes.string),
     priority: PropTypes.arrayOf(PropTypes.string),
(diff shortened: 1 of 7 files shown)


src/shared/types/index.js
```diff
@@ -66,6 +66,7 @@ export const apiFilterType = PropTypes.shape({
     created_after: PropTypes.string,
     created_before: PropTypes.string,
     feedback: PropTypes.string,
+    punctuality: PropTypes.string,
```
aditudorache, line 69: As I can see, in the backend they use `deadline` for this. Can we use that as well? Punctuality let me think at something else :-).

aditudorache (changes requested):

src/shared/types/index.js
MauritsR: The backend has named this filter punctuality so I don't understand where deadline would come into play. Can you clarify?

aditudorache: I see. At category level they call this deadline 
image

aditudorache (approved):

PR #1472, opened 2021-03-18
(Chore) Trigger reloading of the ServiceWorker on new release

aditudorache:
This PR solves the problem of force reloading the application when a new version is released and the service worker is updated.

The added console.log statements are just temporary for process monitoring purpose and will be removed in a later version.


MauritsR (approved):
Should resolve some of the issues we've had with older versions 👍 

Best to pass this solution by Herjen to verify the behavior (clients with slow networks can get a hard refresh several seconds after the first render)

PR #1476, opened 2021-03-22
[SIG-3302] Set aria-label on the autosuggest input

aditudorache:
This PR set the aria-label for the MapInput on the autosuggest input element for a11y.

### DIFF
--- a/src/components/AutoSuggest/index.js
+++ b/src/components/AutoSuggest/index.js
@@ -242,6 +242,7 @@ const AutoSuggest = ({
           ref={inputRef}
           disabled={disabled}
           id={id}
+          autoComplete="off"
           {...rest}
         />
       </div>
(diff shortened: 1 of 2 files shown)


src/components/AutoSuggest/index.js
```diff
@@ -241,6 +242,7 @@ const AutoSuggest = ({
           ref={inputRef}
           disabled={disabled}
           id={id}
```
MauritsR, line 244: How about preventing auto-complete here, since it interferes with our own version of autocomplete/suggest? 

Screenshot 2021-03-22 at 14 37 02


```suggestion
          id={id}
          autoComplete="off"
```

MauritsR (changes requested):

MauritsR (approved):

PR #1481, opened 2021-03-22
[SIG-3626] List departments responsible for categories 

MauritsR:
### Changes

- Show responsible departments in categories view


aditudorache (approved):
General remarks: no typescript, space 12px between labels and input controls, in the tests there is one enzyme/mount test.

src/signals/settings/categories/Detail/components/CategoryForm/CategoryForm.tsx
```diff
@@ -83,45 +97,41 @@ const CategoryForm = ({ data, history, onCancel, onSubmitForm, readOnly }) => {
...
                 readOnly={readOnly}
-                rows="8"
+                rows={8}
               />
             </FieldGroup>
 
             {responsibleDepartments.length > 0 ? (
               <FieldGroup>
-                <dt><strong>Verantwoordelijke afdeling</strong></dt>
+                <StyledDefinitionTerm><strong>Verantwoordelijke afdeling</strong></StyledDefinitionTerm>
```
aditudorache, line 112: The FiledGroup is a `div`. I wouldn't use here `dt/dd` unless was a list and I had used dl as parent. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

aditudorache (changes requested):
Nice! Just one more comment.

aditudorache (approved):

PR #1494, opened 2021-03-25
[SIG-3571] Upgrade to webpack 5

aditudorache:
This PR makes the necessary changes and upgrades webpack to version 5 including the plugins and specific dependencies


MauritsR (approved):
Seems to work well!

PR #1495, opened 2021-03-25
(Chore) Update SSRI

arjandata:
Update SSRI to 8.0.1


MauritsR (dismissed):

MauritsR:
Looks like the security alert is due to webpack requiring an insecure version of SSRI. Therefore, webpack should be updated to fix the alert.

aditudorache (approved):

PR #1501, opened 2021-03-30
Clean up linting configuration

boris-arkenaar:
The configuration for prettier and eslint has been simplified, making use of the standards and defaults as much as possible. I'd like your input on this configuration before I start fixing the code.

Formatting (prettier) and linting (eslint) have been separated.

- `npm run format` formats the code using prettier.
- `npm run lint` runs eslint without any rules that would be corrected by prettier (no red squiggly lines in vscode for any formatting issue, which will be auto fixed anyway)

ESLint plugins I'd still like to add but which are giving too many errors at the moment:

- [sonarjs](https://github.com/SonarSource/eslint-plugin-sonarjs)
- [unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn)
- [jest](https://github.com/jest-community/eslint-plugin-jest)

Other remarks:

- Disabled a few rules for `*.test.*` files specifically.
- Removed rules for Cypress since the e2e tests run separately and have their own linting configuration (haven't done anything with that).

Running `npm run lint:nibble` gives:

```
  @typescript-eslint/explicit-module-boundary-types:  24|
  jsx-a11y/aria-role:                                  6|
  jsx-a11y/no-autofocus:                               1|
  no-undef:                                            1|
  promise/always-return:                               3|
  promise/catch-or-return:                             2|
  promise/no-callback-in-promise:                      7|
  react/display-name:                                  4|
  react/jsx-key:                                       3|
  react/prop-types:                                  174|
  redux-saga/no-unhandled-errors:                     33|
  require-yield:                                       2|
```

So, I've disabled these rules explicitly for now.

Rules with more than 10 warnings/errors:

- `@typescript-eslint/explicit-module-boundary-types` important to specify function return types in our ts code
- `react/prop-types` important to specify our props
- `redux-saga/no-unhandled-errors` redux-saga recommends using try-catch blocks, we could decide to keep this one disabled


.eslintrc.json
```diff
@@ -0,0 +1,73 @@
+{
```
aditudorache, line 1: Please keep this file in js format. This is the standard format for other projects in Amsterdam.

MauritsR:
Definitely going in the right direction! 

Personally, I'm not a fan of the `@typescript-eslint/explicit-module-boundary-type` rule (might explain why it's giving so many errors 😅, but I don't mind if the rule gets enforced. 

Really like it if you add `sonarjs` as a plugin. However, I'm not sure about adding `security` with detect-object-injection. Wouldn't this only make sense for the server-side files?

boris-arkenaar:
@MauritsR I agree on the `security` one.
`@typescript-eslint/explicit-module-boundary-type` seems important to me though. Your module boundary is like the API of the module, you'd want that to be typed when writing typescript.

boris-arkenaar:
I've updated the top message of this PR.
Let's merge this as soon as possible so that the bulk of the changes has gone through without (too many) conflicts (hopefully)!

aditudorache:
Agree! Let's merge it but I see that the e2e are failing. When those are fixed I'll approve it and merge it

aditudorache (approved):

MauritsR (approved):
Looks good!

PR #1502, opened 2021-03-30
Make the incident list horizontally scrollable

boris-arkenaar:
closes Signalen/frontend#123


src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.js
```diff
@@ -26,6 +26,7 @@ const getDaysOpen = incident => {
 
 const StyledList = styled.div`
   width: 100%;
+  overflow: scroll;
```
aditudorache, line 29: Why not `auto`? In this way the scroll bars won't be shown when not needed

aditudorache (changes requested):

aditudorache (approved):

PR #1506, opened 2021-03-31
[SIG-3656] Show child incident history

MauritsR:
### Changes

When viewing an incident with child incidents, the last history item of the child incidents is shown. A toggle button allows the user to expand the history list.


aditudorache (changes requested):
Works nice! I have a couple remarks. I don't know if this is out of scope but I noticed that in the incident detail page the space above `Deelmelding` en `Geshidenis` subtitles is not 32px as it should.

MauritsR:
> Works nice! I have a couple remarks. I don't know if this is out of scope but I noticed that in the incident detail page the space above `Deelmelding` en `Geshidenis` subtitles is not 32px as it should.

To me it looks like the correct spacing is applied - can you post a screenshot?

aditudorache (approved):

PR #1511, opened 2021-04-02
[SIG-3411] Split incident history

MauritsR:
### Changes

- Show 'Geen nieuwe wijzigingen' when there are no recent changes for a split incident
- Show all split incident changes with occurrence later than most recent parent incident change
- Adds E2E tests (by @arjandata )


MauritsR:
- [ ] Fix styling (margins)

aditudorache (changes requested):
Looks good! One test failing in `signals/incident-management/containers/IncidentDetail › should render correctly`

aditudorache (approved):

PR #1513, opened 2021-04-06
[SIG-3572] Upgrade node and npm

MauritsR:
### Changes

- Update default node.js and npm version
- Removes error overlay which is not supported in this npm version


aditudorache:
LGTM, linting fails

MauritsR:
> LGTM, linting fails

Removed a rule that was causing false positives (and a true positive, but I'll wait for Boris' eslint PR to sort that out)

aditudorache (approved):

PR #1515, opened 2021-04-07
[SIG-3609] Radio focus 

aditudorache:
This PR sets a custom focus on the radio because on Firefox the browser default focus is not visible on Mac


src/components/RadioButton/index.tsx
```diff
@@ -22,7 +32,7 @@ interface RadioButtonProps extends InputHTMLAttributes<HTMLInputElement> {
 
 const RadioButton: FunctionComponent<RadioButtonProps> = ({ className, ...props }) => (
   <Wrapper className={className}>
-    <AscRadio {...props} />
+    <StyledRadio {...props} data-test="test" />
```
MauritsR, line 35: data-test?

MauritsR (changes requested):
Installing the new asc-ui fails. Can you synchronize with develop and try again? Might be that npm@7 is throwing a wrench into it.

src/components/RadioButton/index.tsx
aditudorache: Good catch! Thanks!

MauritsR:
Screenshot 2021-04-09 at 12 15 42
I get the result shown above, a mix of the native focus ring and a new black focus ring. Shouldn't the native ring be shown, just more visible? 

This document also describes using the default focus state: [design system - focus state](https://designsystem.amsterdam.nl/7awj1hc9f/p/17684a-focus-state)

aditudorache:
> Screenshot 2021-04-09 at 12 15 42
> I get the result shown above, a mix of the native focus ring and a new black focus ring. Shouldn't the native ring be shown, just more visible?
> 
> This document also describes using the default focus state: [design system - focus state](https://designsystem.amsterdam.nl/7awj1hc9f/p/17684a-focus-state)

It would be the best solution but I don't think is possible. The way how the native focus is shown is the browser responsibility. I couldn't find a way to do that in the application, do you have any suggestion?

MauritsR (approved):
Looks good, but can you install the packages using npm@7 and recommit the package-lock.json?

PR #1516, opened 2021-04-08
[SIG-3579] Map cluster containing selected marker should be red

MauritsR:
## Changes

- When a cluster contains a selected marker, show the cluster icon as red instead of blue


aditudorache (changes requested):
image
I don't see the blue anymore for clusters where nothing is selected. Could you test that?

MauritsR:
> image
> I don't see the blue anymore for clusters where nothing is selected. Could you test that?

Whoops - incorrectly computed a className. Fixed it with the last commit.

aditudorache (approved):
Nice job! Works like a charm!

PR #1519, opened 2021-04-12
[SIG-3672] Prevent double trigger on Toggle

samuelleeuwenburg:
Restore removed guard that prevents double triggering the `onClick` handler


aditudorache (approved):
Can we move this to `typescript`?

PR #1523, opened 2021-04-13
[SIG-3689] Add process time to MetaList

samuelleeuwenburg:
Adds process time to the `MetaList` component on the incident detail page

image

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
@@ -144,6 +144,20 @@ const MetaList = () => {
     [handlingTimesBySlug, incident]
   );
 
+  const [processTimeCopy, processTimeClass] = useMemo(() => {
+    const now = new Date();
+
+    if (now > new Date(incident.category?.deadline_factor_3)) {
+      return ['3x buiten de afhandeltermijn', 'alert'];
+    }
+
+    if (now > new Date(incident.category?.deadline)) {
+      return ['Buiten de afhandeltermijn', 'alert'];
+    }
+
+    return ['Binnen de afhandeltermijn', ''];
+  }, [incident]);
+
   const getDepartmentId = useCallback(
     () => (routingDepartments ? `${routingDepartments[0].id}` : departmentOptions && departmentOptions[0].key),
     [departmentOptions, routingDepartments]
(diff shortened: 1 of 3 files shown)


samuelleeuwenburg:
Currently it doesn't recheck the conditions when time progresses, not sure if that is really needed or how long users stay on this page generally speaking?

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
```diff
@@ -144,6 +144,20 @@ const MetaList = () => {
...
+
+    if (now > new Date(incident.category?.deadline_factor_3)) {
+      return ['3x buiten de afhandeltermijn', 'alert'];
+    }
+
+    if (now > new Date(incident.category?.deadline)) {
+      return ['Buiten de afhandeltermijn', 'alert'];
+    }
+
+    return ['Binnen de afhandeltermijn', ''];
```
MauritsR, line 158: Leaving out the empty string would prevent the rendered component from having an empty class (`class=""`)

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.js
```diff
@@ -144,6 +144,20 @@ const MetaList = () => {
     [handlingTimesBySlug, incident]
   );
 
+  const [processTimeCopy, processTimeClass] = useMemo(() => {
```
MauritsR, line 147: If you'd name this `processTimeText` instead of `processTimeCopy` that would align with naming conventions in other parts of the app

src/signals/incident-management/containers/IncidentDetail/components/MetaList/index.test.js
```diff
@@ -96,6 +96,9 @@ describe('MetaList', () => {
       expect(screen.queryByTestId('meta-list-handling-time-definition')).toHaveTextContent(/^Afhandeltermijn$/);
       expect(screen.queryByTestId('meta-list-handling-time-value')).toHaveTextContent(/^4 werkdagen$/);
 
+      expect(screen.queryByTestId('meta-list-process-time-definition')).toHaveTextContent(/^Doorlooptijd$/);
```
MauritsR, line 99: This is correct, but only because the fixture does not have a `deadline` or `deadline_factor_3`. Can you add that to the incident fixture? Perhaps in a way that you'd get "3x buiten de afhandeltermijn" in order for tests not to break in the future

MauritsR (changes requested):
Nice work :) - Added some minor comments

MauritsR (approved):
Looks good!

PR #1525, opened 2021-04-13
[SIG-3691] Clear 'name' field when reset filter button is clicked

MauritsR:
### Changes

- Fix bug where name filter was not reset after 'Nieuw filter' button was clicked

Technically, the bug is caused by the 'filter name' input not being a controlled input. However, changing this has implications to how the form behaves: Form state is handled in a react reducer, and updates to it causes the `onSubmitForm` callback to be recomputed, which in turn causes all form elements to rerender. This has a noticeable performance impact. 

The solution implemented here (with help from @samuelleeuwenburg) stores the filter data into a ref to remove the filter dependency from the `onSubmitForm` callback. There are some issues with it though: `valuesHaveChanged` is also dependent on the same data, but it's computed with a `useMemo` (thus not updated if changed to use the ref data). 

An alternative could be to use a debounce, which has some other drawbacks (e.g. the submit button label will update after some time instead of directly).

### DIFF
--- a/src/signals/incident-management/components/FilterForm/index.js
+++ b/src/signals/incident-management/components/FilterForm/index.js
@@ -302,8 +302,6 @@ const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, o
     [notRoutedOption.key, state.options.routing_department]
   );
 
-  const [nameValue, setNameValue] = useState(state.filter.name);
-
   return (
     <Form action="" novalidate>
       <ControlsWrapper>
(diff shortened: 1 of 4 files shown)


src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -123,19 +123,27 @@ const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, o
   const dateFrom = state.options.created_after && new Date(state.options.created_after);
   const dateBefore = state.options.created_before && new Date(state.options.created_before);
 
+  // Store ref to filter object (used in onSubmitForm).
```
aditudorache, line 126: Not sure this is the best solution. You store here a part of the reducer state in a ref. It doesn't fill right. If this has to be the solution, I would split the state.filter from the state.options in two separate state structures. In that way would be easier to understand what happens here.

aditudorache (changes requested):

MauritsR:
Closed in favor of https://github.com/Amsterdam/signals-frontend/pull/1527

PR #1527, opened 2021-04-14
[SIG-3691] Clear name/address/note fields when clear button is clicked

MauritsR:
### Changes

- Fix name, note and address inputs not clearing when clicking "Nieuw filter" button

Didn't go with the `createRef` solution discussed earlier - where the uncontrolled input value would be set to empty outside of react - because React complained about it (console error) and I feel like the current solution is more in line with 'normal' usage of React.

### DIFF
--- a/src/signals/incident-management/components/FilterForm/index.js
+++ b/src/signals/incident-management/components/FilterForm/index.js
@@ -196,9 +196,9 @@ const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, o
     [dispatch]
   );
 
-setControlledTextInput({ ...controlledTextInput, name: event.target?.value })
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/components/FilterForm/index.js
```diff
@@ -186,6 +196,10 @@ const FilterForm = ({ filter, onCancel, onClearFilter, onSaveFilter, onSubmit, o
     [dispatch]
   );
 
+  const onNameChange = useCallback(event => setControlledTextInput({ name: event.target?.value }), [setControlledTextInput]);
```
aditudorache, line 199: I believe you are missing the spead here:
```suggestion
setControlledTextInput({ ...controlledTextInput, name: event.target?.value })
```

aditudorache (changes requested):

aditudorache (approved):

PR #1528, opened 2021-04-14
[SIG-3682] Use Fieldset with Legend for radio group form input

MauritsR:
### Changes

- Removes role radiogroup as A11Y review found it an insufficient solution
- Updates header component to optionally render fieldset/legend instead of a label
- Use fieldset/legend for KTO form

### DIFF
--- a/src/signals/incident/components/form/Header/Header.tsx
+++ b/src/signals/incident/components/form/Header/Header.tsx
@@ -22,6 +22,16 @@ const StyledLabel = styled(Label)`
   line-height: ${themeSpacing(6)};
 `;
 
+const FieldSet = styled.fieldset`
+  border: 0;
+  padding: 0;
+  margin-bottom: 0;
+
+  & > :last-child {
+    margin-top: ${themeSpacing(3)};
+  }
+`;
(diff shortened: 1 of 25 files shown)


src/signals/incident/components/form/Header/Header.tsx
```diff
@@ -53,48 +65,51 @@ const Header: FunctionComponent<HeaderProps> = ({
...
-          )}
-
-          {hasError('custom') && <ErrorMessage message={getError('custom') as string} />}
-        </Fragment>
-      )}
-
-      {children}
+      <FieldSetWrapper>
+        {meta?.label && (
+          <StyledLabel {...(isFieldSet ? { as: 'legend' } : { htmlFor: meta.name })}>
```
aditudorache, line 74: Nice!

aditudorache (changes requested):
Great job! Can we rename this  `src/signals/incident/components/form/Header/Header.tsx` component to `FormField`? The role of this component is not in sync with its name anymore.

aditudorache (approved):

PR #1530, opened 2021-04-15
[SIG-3681] Scroll to error in the wizard form

aditudorache:
This PR 
- Implements focus on the not valid element on the reactive forms.
- Replaces the value input type for the select elements from object to array. This is needed to indepent be able to focus on not valid children when the value type is an object without knowing the input control type
- Converts the `SelectInput` to typescript


src/signals/incident/components/IncidentForm/index.js
```diff
@@ -203,7 +205,10 @@ class IncidentForm extends React.Component {
         this.setIncident(formAction);
         next();
       } else {
-        window.scrollTo({ top: 0, behavior: 'smooth' });
+        const invalidControl = Object.values(this.form.controls).find(c => c.invalid);
+        const { name, values } = invalidControl.meta;
+        const valueSelector = !Array.isArray(values) && isObject(values) ? `-${Object.keys(values)[0]}1` : '';
```
MauritsR, line 210: Feels error-prone to maintain this selector, although I can't think think of a different way to get the same functionality. 

Small improvement: looks like it is possible to remove `+ 1` at  `src/signals/incident/components/form/RadioInputGroup/RadioInputGroup.js#L45` so that you won't have to hardcode a `1` here.

MauritsR (changes requested):

src/signals/incident/components/IncidentForm/index.js
aditudorache: Agree that the solution is not ideal, but the best choice for now. Removing `+ 1` won't help. This 1 suffix is created by the reactive-forms framework.

MauritsR: I'm quite sure the suffix is created by `idAttr={`${meta.name}-${key + 1}`}` in the file I linked. As far as I can see there is no use for it.

MauritsR (approved):

PR #1541, opened 2021-04-21
(Chore) convert MetaList to TS

samuelleeuwenburg:
Converts MetaList and the bigger reducer it uses to TypeScript


src/signals/incident-management/containers/IncidentDetail/types.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type Coordinates = [number, number];
+
+export interface Incident {
+  _links: {
+    curies: { href: string; name: string };
+    _link: { self: { href: string } };
+    archives: { href: string };
+    'sia:children': { href: string }[];
```
MauritsR, line 40: sia:children and sia:parent should be typed as optional (only present for hoofdmelding / deelmelding respectively)

aditudorache:
LGTM, there are still tests failing though

aditudorache (approved):

PR #1547, opened 2021-04-28
[SIG-3674] List incidents by same reporter

MauritsR:
## Changes

- Adds list of incidents to /melder page

- [x] Add useReporter unit tests
- [x] Add SPDX license identifiers


src/signals/incident-management/containers/ReporterContainer/types.ts
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2021 Gemeente Amsterdam
+export type Feedback = {
+  is_satisfied: boolean | null
+  submitted_at: string | null
+} | null
```
samuelleeuwenburg, line 6: it would be nicer not to have `null` be part of the `Feedback` type.

That way when you see another interface that implements this type you can easily see if it is nullable/optional, and it also allows you to guard for it and pass it down as non nullable to children components etc.

samuelleeuwenburg (changes requested):

src/signals/incident-management/containers/ReporterContainer/ReporterContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+`
+
+const ReporterContainer: React.FunctionComponent = () => {
+  const {
+    reporter,
+    isLoading,
+    selectedIncident,
+    selectedIncidentId,
+    setSelectedIncidentId,
+  } = useReporter()
```
aditudorache, line 43: How about naming this hook `useFetchReporter`, Then is more clear what is intended. See for instance `useFetchUsers` hook with similar functionality.

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
```diff
@@ -0,0 +1,93 @@
...
+    originalIncidentId: string
+    email?: string
+    incidents?: Reporter
+  }
+}
+
+export const useReporter = (): ReporterHook => {
+  const storeDispatch = useDispatch()
+  const [selectedIncidentId, setSelectedIncidentId] = useState<number>()
+  const { id } = useParams<{ id: string }>()
```
aditudorache, line 29: I would move this to the parent and pass it as parameter here. Is difficult in the container to figure out where the reporter  comes from

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
```diff
@@ -0,0 +1,93 @@
...
+          variant: VARIANT_ERROR,
+          type: TYPE_LOCAL,
+        })
+      )
+    }
+  }, [getReporterError, getIncidentError, storeDispatch])
+
+  return {
+    selectedIncident,
+    selectedIncidentId,
```
aditudorache, line 84: You don't need to export this. You have it already in `selectedIncident`

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
```diff
@@ -0,0 +1,93 @@
...
+    setSelectedIncidentId,
+    isLoading: getReporterLoading || getIncidentLoading,
+    reporter: {
+      originalIncidentId: id,
+      email: incident?.reporter.email,
+      incidents: reporter,
```
aditudorache, line 90: You could add the data related properties (like `selectedIncidnent`) here.

src/components/ParentIncidentIcon/ParentIncidentIcon.tsx
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2021 Gemeente Amsterdam
```
aditudorache, line 2: Nice!

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.js
```diff
@@ -229,7 +213,7 @@ const List = ({
             return (
               <tr key={incident.id}>
                 <Td detailLink={detailLink}>
-                  {incident.has_children && <ParentIcon />}
+                  {incident.has_children && <ParentIncidentIcon />}
```
aditudorache, line 216: Nice!

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
```diff
@@ -0,0 +1,93 @@
...
+          type: TYPE_LOCAL,
+        })
+      )
+    }
+  }, [getReporterError, getIncidentError, storeDispatch])
+
+  return {
+    selectedIncident,
+    selectedIncidentId,
+    setSelectedIncidentId,
```
aditudorache, line 85: as this is an action that can be triggered from outside the hook, I would call it something like `selectIncident` for the outside world because the end result of this action is a new selected incident.

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
```diff
@@ -0,0 +1,93 @@
...
+    isLoading: getReporterLoading || getIncidentLoading,
+    reporter: {
```
aditudorache, line 87: This naming (reporter) is a bit strange: this member is called reporter and the incidents member is the returned reporter data. Calling this `data` or something like that would solve the confusion.

src/signals/incident-management/containers/ReporterContainer/components/FeedbackStatus.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Feedback } from '../types'
+
+interface FeedbackStatusProps {
+  feedback: Feedback | null
+  className?: string
+}
+
+const Status = styled.span<{ feedback: Feedback | null; theme: Theme }>`
+  ${({ feedback, theme }) => {
+    if (feedback === null || !feedback.submitted_at) return
```
aditudorache, line 16: We avoid as much as possible using snake casing in the code. Converting the api results to own defined structures is preffered. In this case parameter like 
`isSatisfied? boolean` would be enough and would make this component reusable in the `detail incident` story

aditudorache (changes requested):

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
MauritsR: I refactored the data quite a bit, can you take another look at it?

src/signals/incident-management/containers/ReporterContainer/useReporter.ts
MauritsR: Refactored!

aditudorache (approved):

samuelleeuwenburg (approved):

PR #1550, opened 2021-04-28
[SIG-3688] samenhang pagination

samuelleeuwenburg:
Adds pagination to the incident list from the same reporter

### DIFF
--- a/src/signals/incident-management/containers/ReporterContainer/ReporterContainer.tsx
+++ b/src/signals/incident-management/containers/ReporterContainer/ReporterContainer.tsx
@@ -64,14 +64,15 @@ const ReporterContainer: FunctionComponent = () => {
     <LoadingIndicator />
   )
 
-  const pagination = incidents?.data?.count && (
-    <StyledCompactPager
-      collectionSize={incidents.data.count}
-      pageSize={PAGE_SIZE}
-      page={currentPage}
-      onPageChange={setCurrentPage}
-    />
-  )
+  const pagination = incidents?.data?.count &&
+    incidents.data.count / PAGE_SIZE > 1 && (
+      <StyledCompactPager
+        collectionSize={incidents.data.count}
+        pageSize={PAGE_SIZE}
+        page={currentPage}
+        onPageChange={setCurrentPage}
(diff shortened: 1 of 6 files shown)


src/signals/incident-management/containers/ReporterContainer/ReporterContainer.tsx
```diff
@@ -4,6 +4,7 @@ import React from 'react'
 import { themeColor, themeSpacing } from '@amsterdam/asc-ui'
 import styled from 'styled-components'
 import LoadingIndicator from 'components/LoadingIndicator'
+import Pagination from 'components/Pagination'
```
MauritsR, line 7: You're able to use [the ASC](https://amsterdam.github.io/amsterdam-styled-components/?path=/story/ui-compactpager--default-story) component instead (this wasn't available earlier)

MauritsR (commented):
Looks good so far 👍

src/signals/incident-management/containers/ReporterContainer/__tests__/useFetchReporter.test.ts
```diff
@@ -26,7 +30,69 @@ jest.mock('react-router-dom', () => ({
...
+    const { result, waitForNextUpdate } = renderHook(
+      () => useFetchReporter(INCIDENT_ID),
+      {
+        wrapper: Provider,
+        initialProps: { store },
+      }
+    )
+
+    await waitForNextUpdate()
+    await waitForNextUpdate()
```
aditudorache, line 80: Why do you need 2 cicli here?

aditudorache (commented):
LGTM

src/signals/incident-management/containers/ReporterContainer/__tests__/useFetchReporter.test.ts
samuelleeuwenburg: is there a way to prevent this? If I don't wait twice there are unhandled async events and it will complain about code not being wrapped in `act()` calls

MauritsR (approved):

PR #1559, opened 2021-04-28
Convert AmsterdamLogo component to TypeScript

jonkoops:


src/components/AmsterdamLogo/index.tsx
```diff
@@ -33,7 +38,9 @@ const AmsterdamLogoStyle = styled.a`
     `};
 `
 
-const AmsterdamLogo = ({ ...otherProps }) => (
+const AmsterdamLogo: FunctionComponent<AmsterdamLogoProps> = ({
+  ...otherProps
+}) => (
   <AmsterdamLogoStyle {...otherProps}>
```
samuelleeuwenburg, line 44: if it is only one prop anyway, its better to not spread it imho

jonkoops: Yeah, but this might break existing logic elsewhere because other props are being passed in JS files. I'd rather not change implementation.

samuelleeuwenburg: but only this project uses this component right? I see no references using any other props, or did I miss one?

samuelleeuwenburg: it is being passed around a lot, but it is basically being used like this:

https://github.com/Amsterdam/amsterdam-styled-components/blob/5b7a907ba6b0090cfb195d308738a7db02f3aa1c/packages/asc-ui/src/components/Header/HeaderLogoText.tsx#L10-L14

jonkoops: Yeah, doesn't seem to be used. Updated.

jonkoops:
@samuelleeuwenburg Removing the spreading broke the tests, looks like more props are passed somewhere implicitly. I've restored the orignal commit.

MauritsR (approved):

PR #1580, opened 2021-05-03
[SIG-3719] Fix redirect before fetching permissions on settings pages

MauritsR:
### Changes

- Fixes bug where user was redirected after refreshing on a settings page (e.g. users page)

### DIFF
--- a/src/containers/App/selectors.ts
+++ b/src/containers/App/selectors.ts
@@ -56,13 +56,13 @@ export const makeSelectUserPermissionCodeNames = createSelector(
 export const makeSelectUserCan = createSelector(
   [makeSelectUser, makeSelectUserPermissionCodeNames],
   (user, permissions) => {
-    const is_superuser = user?.is_superuser ?? false
+    const is_superuser = user?.is_superuser
     /**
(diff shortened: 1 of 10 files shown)


aditudorache (approved):
Nice! Can you check also the unittests? They are failing now

src/containers/App/selectors.ts
```diff
@@ -77,13 +77,13 @@ export const makeSelectUserCan = createSelector(
 export const makeSelectUserCanAccess = createSelector(
   [makeSelectUser, makeSelectUserPermissionCodeNames],
   (user, permissions) => {
-    const is_superuser = user?.is_superuser ?? false
+    const is_superuser = user?.is_superuser
```
samuelleeuwenburg, line 80: I am curious to know if `Boolean(user?.is_superuser)` would work here as wel

samuelleeuwenburg (approved):

src/containers/App/selectors.ts
MauritsR: It wouldn't - `is_superuser` should be `undefined` when the user data is not fetched yet.

PR #1611, opened 2021-05-20
Feature publiccode

ginokok1996:
Added publiccode.yaml following the standard of:
https://docs.italia.it/italia/developers-italia/publiccodeyml-en/en/master/index.html

We have filled in as much information as we could.
But we suggest to check if they are correct.

Adding the publiccode.yaml file allows Dash Kube to correctly index this component at the commonground page found at:
https://dashkube.com/commonground

It also allows for one button installation of the signalen application via Dash Kube.

Changes suggested by @bartjkdp and @Ainali have been done.

### DIFF
--- a/publiccode.yaml
+++ b/publiccode.yaml
@@ -1,44 +1,49 @@
 publiccodeYmlVersion: "0.2"
 name: Signalen frontend
 url: "https://github.com/signalen/frontend"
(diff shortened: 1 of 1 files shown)


ginokok1996:
@MauritsR Thanks for the comments, i have made the changes.

MauritsR (approved):

PR #1628, opened 2021-05-26
(Feature) Add area page for incident 

samuelleeuwenburg:


aditudorache:
Could you add aa [SIG-...] ticket number in the description?

src/components/OverviewMap/OverviewMap.tsx
```diff
@@ -59,7 +59,7 @@ interface Data {
   features: Feature[]
 }
 
-export const POLLING_INTERVAL = 5000
+export const POLLING_INTERVAL = 50000000
```
MauritsR, line 62:
```suggestion
export const POLLING_INTERVAL = 5000
```

MauritsR (approved):

PR #1648, opened 2021-06-04
[SIG-3784] Incident area map

MauritsR:
## Changes

- Adds area map with markers (closed or open), with ability to select a marker
- Adds map close button (used in multiple maps)
- Allows importing svg as `url` (necessary for Leaflet maps) or `ReactComponent`
- Add cluster drawing functionality to `MarkerCluster` - might need more work later since the container map still uses it's own drawing/interaction implementation


custom.d.ts
```diff
@@ -2,8 +2,10 @@
 // Copyright (C) 2020 - 2021 Gemeente Amsterdam
 declare module '*.svg' {
   import { FunctionComponent, SVGProps } from 'react'
-  const Component: FunctionComponent<SVGProps<SVGSVGElement>>
-  export default Component
+  export const ReactComponent: FunctionComponent<SVGProps<SVGSVGElement>>
+
+  const url: string
+  export default url
```
samuelleeuwenburg, line 8: I like having the `url` as the default export, however the rest of JS codebase exports the react component as the default doesn't it?

samuelleeuwenburg (approved):

PR #1666, opened 2021-06-16
[SIG-3806] HTML label structure

MauritsR:
## Changes 

- FileInput `label` now does not contain `div` element
- 'Uur' en 'min' select inputs are now _labelled_ instead of _described_


samuelleeuwenburg:
This probably needs jest threshold adjustment

samuelleeuwenburg (approved):

PR #1682, opened 2021-06-23
[SIG-3800] A11Y: Removes alt text from preview (text alternative is present)

MauritsR:


src/components/MapSelect/index.js
```diff
@@ -131,7 +131,7 @@ const MapSelect = ({
                 feature.properties[iconField],
                 selection.current.has(feature.properties[idField])
               ),
-              alt: feature.properties.objectnummer,
+              alt: selectionOnly ? '' : feature.properties.objectnummer,
```
samuelleeuwenburg, line 134: doesn't this give you an empty property? `alt=""` or does it not set it with an empty string?

MauritsR: Yes it is set to `alt=""`. This is a valid method to prevent the screenreader from reading out lout other info contained in the tag.

samuelleeuwenburg (approved):

PR #1703, opened 2021-07-05
(Chore) Prevents flash of content on initial render of signaling dashboard

MauritsR:


src/signals/incident-management/containers/Signaling/components/GraphDescription/GraphDescription.tsx
```diff
@@ -50,7 +50,7 @@ const GraphDescription: FunctionComponent<GraphDescriptionProps> = ({
   description,
   total,
 }) => (
-  <Wrapper>
+  <Wrapper data-testid="graphDescription">
```
samuelleeuwenburg, line 53: kebak-case-perhaps

src/signals/incident-management/containers/Signaling/Signaling.tsx
```diff
@@ -111,15 +111,18 @@ const Signaling: FunctionComponent = () => {
       {heading}
       <Row>
         <StyledColumn span={6} wrap>
-          <GraphDescription
-            title={`Openstaande meldingen tot en met 2020`}
-            description={`
+          {totalOpen !== null ? (
```
samuelleeuwenburg, line 114: if its data or null, might as well go `totalOpen && (....`   vs   `totalOpen !== null ? (..) : null`

src/signals/incident-management/containers/Signaling/Signaling.tsx
```diff
@@ -131,14 +134,17 @@ const Signaling: FunctionComponent = () => {
           )}
         </StyledColumn>
         <StyledColumn span={6} wrap>
-          <GraphDescription
-            title={`Verzoek tot heropenen tot en met Q1 2021`}
-            description={`
+          {totalReopenRequested !== null ? (
```
samuelleeuwenburg, line 137: Same as above

samuelleeuwenburg (changes requested):
Small details

src/signals/incident-management/containers/Signaling/Signaling.tsx
MauritsR: That's not possible: if `totalOpen === 0` the component would not be rendered.

samuelleeuwenburg: in the `useMemo` above you use `total_signal_count || null`, if that number is `0` it will default to `null` as well, maybe that needs a more precise check in that case

MauritsR: Sharp! Fixed it

samuelleeuwenburg (approved):

PR #1719, opened 2021-07-13
[SIG-3895] Override min-width for the remove button on the Caterpillar map

samuelleeuwenburg:


MauritsR (changes requested):
Missing change for container map

MauritsR (approved):

PR #1741, opened 2021-07-20
Fix publiccode.yml

bfabio:
* mv publiccode.yaml publiccode.yml
* Use the correct SPDX tag for the license
* Force dependencies versions as strings


MauritsR (approved):
Thanks for submitting this PR! 👍

PR #1769, opened 2021-09-14
[SIG-3995] Update incident overview table

MauritsR:
Changes

- By default, a row has a height of two lines, and text is truncated on the second line
- Switches up the ordering of the columns 
- Because the address formatting is changed, it now uses a formatting function instead of the address display value


src/signals/incident-management/containers/IncidentOverviewPage/components/List/styles.ts
```diff
@@ -31,25 +31,49 @@ export const Th = styled.th`
...
     a {
       text-decoration: none;
       color: black;
-      display: flex;
-      padding: ${themeSpacing(2)};
+      overflow: hidden;
+      text-overflow: ellipsis;
+
+      // Show ellipsis on second line
+      @supports (-webkit-line-clamp: 2) {
```
janjaap, line 72: No real need to use this here; both `@supports` and `-webkit-line-clamp` are supported by the same set of browsers. You can just use
```css
display: -webkit-box;
-webkit-line-clamp: 2;
-webkit-box-orient: vertical;
```

PR #1771, opened 2021-09-14
Changes header texts

janjaap:
This PR changes signaling column header texts.

### DIFF
--- a/src/signals/incident-management/containers/Signaling/Signaling.tsx
+++ b/src/signals/incident-management/containers/Signaling/Signaling.tsx
@@ -47,12 +47,9 @@ const Signaling: FunctionComponent = () => {
   } = useGetReportReopenRequested()
 
   useEffect(() => {
-    getReportOpen({ end: new Date().toISOString() })
-  }, [getReportOpen])
-
-  useEffect(() => {
+    getReportOpen()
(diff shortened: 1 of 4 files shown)


MauritsR (changes requested):
If you remove the end date that is used in the request for this data, the data will reflect the updated text.

janjaap:
> If you remove the end date that is used in the request for this data, the data will reflect the updated text.

Done

src/signals/incident-management/containers/Signaling/Signaling.tsx
```diff
@@ -49,7 +47,7 @@ const Signaling: FunctionComponent = () => {
   } = useGetReportReopenRequested()
 
   useEffect(() => {
-    getReportOpen({ end: endOpen })
+    getReportOpen({ end: new Date().toISOString() })
```
MauritsR, line 50: The end parameter is optional in the API, so the addition is unnecessary. 

Looks like `getReportOpen` should be typed differently and/or set a default parameter to handle empty input.

MauritsR (dismissed):

src/signals/incident-management/containers/Signaling/Signaling.tsx
janjaap: Good catch. Changed it.

PR #1778, opened 2021-09-27
[SIG-3946] Refactor auth implementation

MauritsR:
Changes

- Rename 'authz' file to 'implicit-auth' to better communicate what this file implements (implicit auth flow)
- Simplifies application configuration to prevent double config for oidc providers
- Convert auth files to TypeScript
- Adds silent SSO using IFrame (see [docs](https://github.com/keycloak/keycloak-documentation/blob/master/securing_apps/topics/oidc/javascript-adapter.adoc))

Relates to PR: https://github.com/Amsterdam/signalen/pull/82


src/shared/services/auth/services/keycloak.ts
```diff
@@ -0,0 +1,127 @@
...
+  private refreshIntervalId: number | null = null
+
+  constructor() {
+    this.keycloak = keycloakJS({
+      clientId: configuration.oidc.clientId,
+      realm: (configuration.oidc as any).realm as string,
+      url: configuration.oidc.authEndpoint,
+    })
+
+    this.keycloak.onAuthSuccess = () => this.startRefreshInterval()
```
janjaap, line 23: Shouldn't this also work:
```JS
this.keycloak.onAuthSuccess = this.startRefreshInterval
```

Or
```JS
this.keycloak.onAuthSuccess.bind(this.startRefreshInterval)
```

src/shared/services/auth/services/keycloak.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    return token ? { Authorization: `Bearer ${token}` } : {}
+  }
+
+  getAccessToken(): string | null {
+    return this.keycloak.token || null
+  }
+
+  login() {
+    if (typeof global.Storage === 'undefined') {
```
janjaap, line 87: When this condition evaluates to true, line 10 could already have thrown a `DOMException`.

src/shared/services/auth/services/keycloak.ts
```diff
@@ -0,0 +1,127 @@
...
+  startRefreshInterval() {
+    // Refresh the access token periodically
+    const minValidity = 30 // Token should be valid for at least the next 30 seconds
+    const updateInterval = minValidity * 0.75 // Keep token valid by checking regularly
+
+    // Start a token updater, if not yet running
+    if (!this.refreshIntervalId) {
+      this.refreshIntervalId = setInterval(() => {
+        this.keycloak.updateToken(minValidity)
+      }, updateInterval * 1000) as unknown as number
```
janjaap, line 115: Is it necessary to check that often? Roughly every twenty seconds. I thought that the token was valid for five minutes, or so?

src/shared/services/auth/auth.ts
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2018 - 2021 Gemeente Amsterdam
+import { UserCredentials } from 'containers/App/types'
+import configuration from '../configuration/configuration'
+import Keycloak from './services/keycloak'
+import ImplicitAuth from './services/implicit-auth'
+
+const useKeycloak =
```
janjaap, line 8: I choose a different name for this variable; in the React context, this name signifies a hook.

src/shared/services/auth/services/keycloak.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        Object.assign({}, keycloakJSMock, {
+          authenticated: true,
+          isTokenExpired: () => false,
+        })
+      )
+
+      expect(new Keycloak().getIsAuthenticated()).toBe(true)
+    })
+
+    it('returns false if user is not authenticaed is expired', () => {
```
janjaap, line 108: Is this sentence correctly formulated? 🤔

src/shared/services/auth/services/keycloak.ts
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2020 - 2021 Gemeente Amsterdam
+import keycloakJS, { KeycloakInitOptions, KeycloakInstance } from 'keycloak-js'
+import configuration from 'shared/services/configuration/configuration'
+
+const AUTH_REDIRECT_URI = `${global.location.protocol}//${global.location.host}/manage/incidents`
```
janjaap, line 6: Why not use `window.location.origin` here as well?

src/shared/services/auth/services/keycloak.ts
```diff
@@ -0,0 +1,127 @@
...
+const OAUTH_DOMAIN_KEY = 'oauthDomain' // Domain that is used for login
+
+const storage = global.localStorage || global.sessionStorage
+
+class Keycloak {
+  private keycloak: KeycloakInstance
+  private refreshIntervalId: number | null = null
+
+  constructor() {
+    this.keycloak = keycloakJS({
```
janjaap, line 17: Shouldn't `this.keycloak` be a singleton?

janjaap (commented):
Nice work on the refactor; definitely improves readability. A couple of remarks on the proposed changes.

src/shared/services/auth/services/keycloak.test.ts
MauritsR: :sweat_smile: Most definitely not, haha. Will rewrite

src/shared/services/auth/services/keycloak.ts
MauritsR: Removed this check

src/shared/services/auth/services/keycloak.ts
MauritsR: I changed it to check less often. It's correct that the token is valid for five minutes.

src/shared/services/auth/services/keycloak.ts
MauritsR: `window` is indeed better in this context, changed it

janjaap: I actually meant using `window.location.origin` instead of `${global.location.protocol}//${global.location.host}`

src/shared/services/auth/services/keycloak.ts
MauritsR: By making it a singleton, this class becomes really tricky to unit test. I prefer keeping the current implementation in place and only use the `Keycloak` class in the `auth.ts` file, which is our go-to file for authentication in the app.

src/shared/services/auth/services/keycloak.ts
MauritsR: Whoops, I read that wrong, fixed it

janjaap (approved):

PR #1781, opened 2021-09-27
[SIG-3975] Change caterpillar map question to be required

MauritsR:


src/signals/incident/components/form/MapSelectors/Caterpillar/CaterpillarSelect.tsx
```diff
@@ -57,26 +57,28 @@ const CaterpillarSelect: FunctionComponent<CaterpillarSelectProps> = ({
...
-        update,
-        edit,
-        close,
-        setMessage,
-      }}
-    >
-      {!showMap && value.length === 0 && (
-        <Intro edit={edit} location={location} />
-      )}
+    <div id={meta.name}>
```
janjaap, line 60: Why the extra `<div>`?

MauritsR: The `id` is set for the error handler to 'find' the map select element and scroll to it. I added a `<div>` because there are no other elements where this is element makes sense to place. E.g. if  it was placed on the `<Intro />` it would be confusing: the `<Intro />` element is not always shown (even though it is always shown when the input is invalid - which is not something that is immediately obvious).

janjaap (approved):

PR #1789, opened 2021-09-29
[SIG-3952] Hide header in app mode

janjaap:
Adds `appMode` feature flag and prevents rendering of the site header when flag is set.

### DIFF
--- a/internals/schemas/app.schema.json
+++ b/internals/schemas/app.schema.json
@@ -111,6 +111,10 @@
         "useStaticMapServer": {
           "description": "When true, will use the static map server defined by map.staticUrl to get a png image for a small map preview, otherwise leaflet without controls will be used.",
           "type": "boolean"
+        },
+        "appMode": {
+          "description": "When true, will render the application in app mode, hiding specific UI elements",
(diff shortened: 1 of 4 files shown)


internals/schemas/app.schema.json
```diff
@@ -111,6 +111,10 @@
         "useStaticMapServer": {
           "description": "When true, will use the static map server defined by map.staticUrl to get a png image for a small map preview, otherwise leaflet without controls will be used.",
           "type": "boolean"
+        },
+        "appMode": {
+          "description": "When true, will render the application in app mode, hiding specific UI elements",
```
MauritsR, line 116: `hiding specific UI elements` will not be the only effect of this flag - perhaps better to mention that this is an example. Alternatively, we update the description when we implement another of the upcoming `app` features.

MauritsR (approved):

internals/schemas/app.schema.json
janjaap: I'm not aware of any changes to need to be applied to the code base to accommodate the app frame. Would indeed be better to update the description as we go.

PR #1790, opened 2021-09-29
Chore replace custom dev server

janjaap:
This PR replaces the custom dev server with Webpack's internal dev server. Also, the npm script command `start:prod` has been removed since `docker-compose` can be used to spin up the production version of the application.


package.json
```diff
@@ -29,9 +29,8 @@
     "preanalyze": "npm run analyze:clean",
     "prebuild": "npm run build:clean",
     "pretest": "npm run test:clean",
-    "start:prod": "cross-env BABEL_ENV=production NODE_ENV=production node server",
-    "start:tunnel": "cross-env BABEL_ENV=production NODE_ENV=production ENABLE_TUNNEL=true node server",
-    "start": "cross-env NODE_ENV=development BABEL_ENV=development node server -- --port=3001",
+    "start:prod": "npm run build && cross-env NODE_ENV=production node server",
+    "start": "cross-env NODE_ENV=development npx webpack s --config ./internals/webpack/webpack.dev.babel --port=3001 --hot --compress --history-api-fallback --watch-files=./src/**/*",
```
MauritsR, line 33: Don't think `npx` is necessary here - `webpack` is already installed as a devDependency

MauritsR (approved):
Approved - but placed a comment in the `package.json`

package.json
```diff
@@ -12,9 +12,9 @@
   "license": "MPL-2.0",
   "scripts": {
     "analyze:clean": "rimraf stats.json",
-    "analyze": "BABEL_ENV=production NODE_ENV=production node ./internals/scripts/analyze.js",
+    "analyze": "NODE_ENV=production node ./internals/scripts/analyze.js",
```
MauritsR, line 15:
```suggestion
    "analyze": "ANALZYE=true npm run build",
```
Would make more sense now that we have the BundleAnalyzerPlugin directly in the webpack base config. Then the `analyze:clean` line as wel as `analyze.js` can also be removed.

PR #1791, opened 2021-09-30
(Chore) Update docker compose usage

MauritsR:
- It is unnecessary to install `docker-compose`, since it's possible to use `docker compose` instead 
- Prevent running pipeline on `release` build, which triggered two GitHub action 'checks'.


MauritsR:
Closed because Docker does not come with the `compose` command on Linux systems

PR #1804, opened 2021-10-04
(Chore) Prevent keycloak redirect to overview page on login

MauritsR:
On login, the application should return to the same page as where the login flow was started. 

This also fixes an issue where when running the application on `localhost`, the user was not able to reach the incident form page.


janjaap (approved):
Surprised to see that this fix doesn't require a change in one of the tests. Approved nonetheless.

PR #1807, opened 2021-10-05
(Chore) Change GitHub actions config to break pipeline on broken build

MauritsR:
E2E job should
- [Fail](https://github.com/Amsterdam/signals-frontend/actions/runs/1304048385) step when build fails
- [Fail](https://github.com/Amsterdam/signals-frontend/actions/runs/1303502021) step when a job fails 
- [Succeed](https://github.com/Amsterdam/signals-frontend/actions/runs/1303504515) when build+tests succeed

Removed step to print dashboard URL because this URL is already printed in every E2E step


janjaap (approved):
Won't

PR #1846, opened 2021-11-08
(Hotfix) contain:content blocker

janjaap:
This PR fixes an issue where pages could not be scrolled where the height of the content exceeded the height of the viewport.


kiruys (commented):
helemaal top!

PR #1849, opened 2021-11-08
[SIG-4045] Filter source options

kiruys:
A source will only be shown as option in case its can_be_selected property is true.

Please note that this change depends on SIG-4044.
Before merging this PR, it should be tested when SIG-4044 is available on acceptance.

### DIFF
--- a/src/signals/incident/definitions/wizard-step-1-beschrijf.js
+++ b/src/signals/incident/definitions/wizard-step-1-beschrijf.js
@@ -21,8 +21,8 @@ const typesValuesList = typesList.reduce(
   (acc, { key, value, info }) => ({ ...acc, [key]: { value, info } }),
   {}
 )
-const filterSources = (sources) =>
-  sources.filter((source) => source.can_be_selected === true)
+const selectableSources = (sources) =>
+  sources.filter((source) => source.can_be_selected)
 const reduceSources = (sources) =>
   sources.reduce(
     (acc, { value }) => [...acc, { [value]: value }],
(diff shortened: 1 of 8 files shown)


src/signals/incident/definitions/wizard-step-1-beschrijf.js
```diff
@@ -21,6 +21,8 @@ const typesValuesList = typesList.reduce(
   (acc, { key, value, info }) => ({ ...acc, [key]: { value, info } }),
   {}
 )
+const filterSources = (sources) =>
```
janjaap, line 24: No need to use `=== true`:

```javascript
const filterSources = (sources) =>   sources.filter(({ can_be_selected }) => can_be_selected)
```

Personally, I would name the function `selectableSources`.

src/signals/incident/definitions/wizard-step-1-beschrijf.js
```diff
@@ -37,7 +39,7 @@ const getControls = memoize(
         meta: {
           label: 'Hoe komt de melding binnen?',
           path: 'source',
-          values: sources ? reduceSources(sources) : [],
+          values: sources ? reduceSources(filterSources(sources)) : [],
```
janjaap, line 42: Would be good to add a test for this change.

kiruys: with the next commit the can_be_selected prop is tested in wizard_step_1

src/signals/incident/definitions/wizard-step-1-beschrijf.js
kiruys: both done👍

janjaap (approved):

boris-arkenaar:
@janjaap @kiruys
This has been merged while it still has the label `do not merge yet`.
Now no sources show up for me locally.

boris-arkenaar:
I see https://github.com/Amsterdam/signals/pull/879 has already been merged, so should be fine.
I've updated backend master and now everything works again locally.

kiruys:
@boris-arkenaar I wanted to remove the do not merge label, but didn't know how... The sources that need to be filtered can be adjusted in the django admin.

PR #1850, opened 2021-11-08
[SIG-4099] Markdown parse

janjaap:
This PR contains changes that allow for Markdown as input format for any form that renders into a `History` component as well as the 'Thank you' page of the incident form.
Note that a change to the `mapDynamicFields` function was necessary to prevent the Markdown parser to trip over the string that was returned by that function whenever an entry could not be returned.


kiruys (approved):
looks good! I have checked out the branch, added a markdown link to the thank you text, and it worked! ;-)

PR #1855, opened 2021-11-09
[SIG-4116] Replace incident overview table header

janjaap:
This PR replaces the table header sorting options with the Sort component

### DIFF
--- a/src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
+++ b/src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
@@ -107,25 +107,24 @@ const List: FunctionComponent<ListProps> = ({
       <Table cellSpacing="0">
         <thead>
           <tr>
-            <ThParent data-testid="parent" />
-            <ThPriority data-testid="priority" />
-            <Th data-testid="sortId">Id</Th>
-            <Th data-testid="sortDaysOpen">Dag</Th>
-            <ThDate data-testid="sortCreatedAt">Datum en tijd</ThDate>
-            <ThSubcategory data-testid="sortSubcategory">
-              Subcategorie
-            </ThSubcategory>
-            <ThStatus data-testid="sortStatus">Status</ThStatus>
+            <ThParent />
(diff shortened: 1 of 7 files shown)


kiruys:
The date ordering seems to be reversed, so 'nieuw - oud' gives old incidents on top, and 'oud - nieuw' gives new incidents on top

kiruys:
I'm not sure what "toegewezen aan" ordering does exactly.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
```diff
@@ -137,55 +109,23 @@ const List: FunctionComponent<ListProps> = ({
...
-            <ThSubcategory
-              data-testid="sortSubcategory"
-              onClick={onSort('sub_category,-created_at')}
-            >
-              Subcategorie {renderChevron('sub_category')}
+            <Th data-testid="sortId">Id</Th>
+            <Th data-testid="sortDaysOpen">Dag</Th>
+            <ThDate data-testid="sortCreatedAt">Datum en tijd</ThDate>
+            <ThSubcategory data-testid="sortSubcategory">
+              Subcategorie
```
kiruys, line 116: Is it possible to remove the data-testid attributes now that the sorting is done differently?

kiruys (approved):

janjaap:
> I'm not sure what "toegewezen aan" ordering does exactly.

Good point; I forgot to make that option conditional base on the value of the `assignSignalToEmployee` feature flag.

janjaap:
> The date ordering seems to be reversed, so 'nieuw - oud' gives old incidents on top, and 'oud - nieuw' gives new incidents on top

Ouch, good one; I'll change that.

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
janjaap: I guess so. They don't seem to be used at all.

PR #1859, opened 2021-11-11
[SIG-1460] Replace fonts

janjaap:
This PR:
- Removes the fonts from this repo and replaces them with a `<link />` tag to get the `fonts.css` file from static.amsterdam.nl
- Removes as much custom font sizes and line heights so that the settings from Amsterdam Styled Components are used
- Removes unused CSS rules from `global.scss` and removes the last SCSS mixins


kiruys:
How can I test this? The footer links for example don't look the way the are supposed to.
Do I need to update other stuff as well?

kiruys (approved):

PR #1863, opened 2021-11-11
[SIG-4046] Add source param in form to indicate that incident was reported through app

kiruys:
I had to create a new form component, StaticHiddenInput, because the HiddenInput component is being used for dynamic updates of the datetime component and the image file upload.


janjaap:
I don't think we need a new component; the references in the form configuration do not render a component, so we can leave that out. The parser for the form configuration needs an entry and (apparently) not necessarily a component to render. You can apply your changes to the `HiddenInput` component and change the references in the form configuration from
```
incident_time_hours: {
  meta: {
    label: 'Incident time hours',
    readOnly: true,
  },
  render: FormComponents.HiddenInput,
},
```
to
```
incident_time_hours: {
  meta: {
    label: 'Incident time hours',
    readOnly: true,
  },
},
```
That should also work

src/signals/incident/components/form/StaticHiddenInput/StaticHiddenInput.js
```diff
@@ -0,0 +1,35 @@
...
+  const { parent, meta } = props
+
+  if (!meta.name || !meta.value) return null
+
+  const Render = ({ handler }) => {
+    useEffect(() => {
+      parent.meta.updateIncident({
+        [meta.name]: meta.value,
+      })
+    }, [])
```
janjaap, line 17: This part isn't covered by a test case.

kiruys:
> I don't think we need a new component; the references in the form configuration do not render a component, so we can leave that out. The parser for the form configuration needs an entry and (apparently) not necessarily a component to render. You can apply your changes to the `HiddenInput` component and change the references in the form configuration from
> 
> ```
> incident_time_hours: {
>   meta: {
>     label: 'Incident time hours',
>     readOnly: true,
>   },
>   render: FormComponents.HiddenInput,
> },
> ```
> 
> to
> 
> ```
> incident_time_hours: {
>   meta: {
>     label: 'Incident time hours',
>     readOnly: true,
>   },
> },
> ```
> 
> That should also work

good point!

src/signals/incident/definitions/wizard-step-5-samenvatting.js
```diff
@@ -173,6 +173,7 @@ export default {
       page_summary: {
         meta: {
           value: 'summary',
+          name: 'page_summary',
         },
         render: FormComponents.HiddenInput,
```
janjaap, line 178: This line can be removed, right?

kiruys: No it cannot. To make it work, I added the name prop

janjaap: I'm pretty sure the line should be removed; leaving it in will dispatch the `UPDATE_INCIDENT` action and set the `summary` prop in the state with value 'page_summary'. Setting that prop has no effect nor has it any business in the state. As mentioned before, the `page_summary` prop in the `controls` configuration could also be `page_summary: {}`.

janjaap (approved):

PR #1893, opened 2021-11-23
[SIG-4146] Combine fields

janjaap:
This PR combines the incident form steps `telefoon` and `email` into a single step; `contact`. Since the new step needed more than one heading and since this wasn't supported, I added a component `WithHeading` to be able to support that. The `WithHeading` component wraps another component and renders that component with an `h2` heading element.


docs/adr/0006-forms.md
```diff
@@ -67,9 +67,8 @@ A form components specifically using the dependency, which can easily be updated
 The definitions for all steps in the wizard can be turned into components, defining the form directly.
 
 - [wizard-step-1-beschrijf.js](../../src/signals/incident/definitions/wizard-step-1-beschrijf.js)
-- [wizard-step-3-telefoon.js](../../src/signals/incident/definitions/wizard-step-3-telefoon.js)
-- [wizard-step-4-email.js](../../src/signals/incident/definitions/wizard-step-4-email.js)
-- [wizard-step-5-samenvatting.js](../../src/signals/incident/definitions/wizard-step-5-samenvatting.js)
+- [wizard-step-3-contact.js](../../src/signals/incident/definitions/wizard-step-3-contact.js)
+- [wizard-step-4-summary.js](../../src/signals/incident/definitions/wizard-step-5-summary.js)
 - [wizard-step-6-bedankt.js](../../src/signals/incident/definitions/wizard-step-6-bedankt.js)
 - [wizard-step-7-fout.js](../../src/signals/incident/definitions/wizard-step-7-fout.js)
```
kiruys, line 73: Shouldn't  steps 6 and 7 also be renamed to 5 and 6?

janjaap: Good catch! I have those links fixes in an upcoming PR

src/signals/incident/components/form/WithHeading/WithHeading.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+type WithHeadingProps = FormOptions & { meta: WrapMeta }
+
+const StyledHeading = styled(Heading)`
+  margin: ${themeSpacing(2, 0, 5)};
+`
+
+const WithHeading: FC<WithHeadingProps> = (props) => {
+  const { wrap, heading } = props.meta
+
```
kiruys, line 20: wrap refers to the wrapped component, right?
if so, this is a bit confusing.
I would prefer something like content, or input, or question...

janjaap: Yeah, I was debating a more descriptive name. Maybe `wrappedComponent`?

src/signals/incident/definitions/wizard-step-3-contact.js
```diff
@@ -0,0 +1,76 @@
...
+          heading: 'Mogen we u bellen voor vragen?',
+          label: 'Wat is uw telefoonnummer?',
+          path: 'reporter.phone',
+          subtitle:
+            'Zo kunt u ons helpen het probleem sneller of beter op te lossen.',
+          type: 'tel',
+          wrap: FormComponents.TextInput,
+        },
+        render: FormComponents.WithHeading,
+        options: {
```
kiruys, line 32: I would prefer to see the actual component here, so in this case TextInput, and then have withHeading as a meta property. However, I understand that if the preference is to have a separate component that can be used for more form components that this is the pragmatic solution.

src/signals/incident/components/form/WithHeading/WithHeading.tsx
kiruys: 👍

kiruys (approved):

src/signals/incident/definitions/wizard-step-3-contact.js
janjaap: That's not possible with the current setup; `react-reactive-form` expect a component to render. It cannot be reversed.

PR #1894, opened 2021-11-23
Sig 4147 adjust summary page

janjaap:
This PR:
- Adds a `Contact` section to the incident form summary page
- Renders the `WithHeading` component in a `<div />` wrapper to make sure that the parent fieldset CSS rules don't mess up the alignment
- Makes sure that the `WithHeading` component pass on the correct values to its child component (the `parent` prop is required by `react-reactive-form`, but isn't automatically passed down when creating a wrapper component)
- Fixes an unrelated test (for the `Map` component) that polluted the test output


kiruys (approved):
Looks good!

PR #1912, opened 2021-12-07
[SIG-4185] Surrounding area SVG icons fix

janjaap:
This PR fixes an issue where icons, that are referenced by URL and are not loaded as React components, were not shown. As it turned out, the `@svgr` package was't capable of rendering both React components and SVG images by URL. Switching to `svg-url-loader` solved that issue. I was also able to get rid of the `webpack-image-loader` dependency; Webpack has builtin support for image handling.

When an SVG needs to be referenced by its URL, Webpack's [`resourceQuery`](https://webpack.js.org/configuration/module/#ruleresourcequery) option needs to be used, like so:
```javascript
import incidentSvgUrl from '../../incident.svg?url'
```

Rendering an SVG as a React component:
```javascript
import IncidentIcon from '../../incident.svg'
```


kiruys (approved):
goed gevonden!

kiruys (approved):

PR #1915, opened 2021-12-07
[SIG-4182] AutoSuggest to ts

janjaap:
This PR contains the conversion to TS for the `SuggestList`, `AutoSuggest` and `PDOKAutoSuggest` components. Next to that, the appropriate tests have been cleaned


src/components/AutoSuggest/AutoSuggest.test.tsx
```diff
@@ -99,35 +115,33 @@ describe('src/components/AutoSuggest', () => {
...
     expect(fetch).not.toHaveBeenCalled()
 
     act(() => {
       fireEvent.change(input, { target: { value: 'Ams' } })
     })
 
-    await findByTestId('autoSuggest')
+    await screen.findByTestId('autoSuggest')
 
     expect(fetch).not.toHaveBeenCalled()
```
kiruys, line 129: I don't understand why fetch should not be called here.

kiruys (approved):

src/components/AutoSuggest/AutoSuggest.test.tsx
janjaap: Fetch should not be called when the input delay timer hasn't passed

PR #1917, opened 2021-12-10
[SIG-4092] Convert coordinates from array to literal

janjaap:
This PR contains a rewrite of the global incident state so that the location is stored as a latlong literal 
(`{ lat: number, lng: number }`) and not as a tuple (`[number, number]`). The latter is used in the output of the incident detail API and leads to confusion; the API returns the values in the `lng, lat` order where the rest of the application and its components uses the form `lat, lng`.

 Where possible, the literal type is used and only in the communication with the API, the tuple type is used. Also, the term `location` has been replaced with `coordinates` where applicable.


src/shared/services/map-location/map-location.test.ts
```diff
@@ -21,56 +21,22 @@ const testAddress = {
...
+const coordinates = { lng: 4, lat: 52 }
 
 const testFeature = {
   type: 'Point',
-  coordinates: [4, 52],
+  coordinates: [coordinates.lat, coordinates.lng] as LatLngTuple,
 }
 
 describe('locationToFeature', () => {
   it('should convert', () => {
```
kiruys, line 32: maybe we should rename these methods to coordinatesToFeature, and featureToCoordinates?

src/signals/incident/components/IncidentPreview/components/MapSelect/MapSelect.test.js
```diff
@@ -26,15 +26,15 @@ describe('signals/incident/components/IncidentPreview/components/MapSelect/MapSe
...
   })
 
   it('returns default coords', () => {
     expect(getLatlng()).toEqual({
-      latitude: DEFAULT_COORDS[1],
-      longitude: DEFAULT_COORDS[0],
+      lat: DEFAULT_COORDS[0],
+      lng: DEFAULT_COORDS[1],
     })
   })
```
kiruys, line 39: is it correct that the indexes should be reversed here?

kiruys:
In response to your PR description: when is location applicable?

janjaap:
> In response to your PR description: when is location applicable?

Location, in the context of SIA, is an entity that has both coordinates and address values. The term 'location' is used in more than one context which is confusing.

src/signals/incident/components/IncidentPreview/components/MapSelect/MapSelect.test.js
janjaap: Yes, the default coordinates that were used in the `MapSelect` component were set in an order that differs from the usual order. Like the API, the order in the `MapSelect` component was set to `lng,lat` instead of `lat,lng`. I removed the default coordinates from the `MapSelect` implementation and replaced them with the coordinates from the configuration. The configuration uses the 'correct' order. That is why the change in the test was required.

src/shared/services/map-location/map-location.test.ts
janjaap: Absolutely. Good catch. Is it okay if I tackle that in an upcoming PR?

kiruys (approved):

PR #1924, opened 2021-12-15
[SIG-4197] Limit selection on object map to single object or location

janjaap:
Changes the functionality, that allows selecting multiple objects in an object selector map, to selecting a single object.
To make this work, most of the components in the `Asset` component's folder have been touched.

### Most notable changes
- `CaterpillarListPreview` component has been removed since it can be replaced by `AssetList`
- The `update` callback function in the `AssetSelect` context has been replaced with `setItem` and `removeItem`
- The props `address` and `coordinates` have been added to the `AssetSelect`context
- Objects that are selected now have a `location` object. That object contains the coordinates of the selected object and, when the address can be resolved, also the address. The address need to be shown in the summary.
- The `IncidentContainer` saga converts the lat/lng object back to a format that the API understands

### General improvements
- Test files have been moved out of their `__tests__` folders for a cleaner file structure
- Major components have been mocked in test suites to simplify testing and to ensure that test only cover the unit under test
- For some components, the amount of required props has been reduced in favor of using the `AssetSelect` context
- The `withAssetSelectContext` function has been moved out of the `Asset/__tests__/context.test` file; every time that file was imported by another test, the whole context test suit was executed


kiruys:
did you test if the caterpillar preview indeed works with the asset list component?

src/signals/incident/components/form/MapSelectors/Asset/AssetList/AssetList.tsx
```diff
@@ -57,40 +64,35 @@ const AssetList: FunctionComponent<AssetListProps> = ({
...
+              data-testid={`assetListRemove-${item.id}`}
+              aria-label="Verwijder"
+              icon={<Close />}
+              onClick={onRemove}
+            />
+          )}
+        </ItemWrapper>
+      </IconListItem>
     </IconList>
   )
```
kiruys, line 97: If it is only possible to have one item here, shouldn't the list be removed?

janjaap: We can, but the list component has styling for the icons and it was easier to leave it as it is.

src/signals/incident/components/form/MapSelectors/Asset/Intro/Intro.test.tsx
```diff
@@ -4,15 +4,16 @@ import { fireEvent, render, screen } from '@testing-library/react'
...
 
 const contextValue: AssetSelectValue = {
-  selection: [],
+  selection: undefined,
   meta: {} as Meta,
   coordinates: { lat: 0, lng: 0 },
-  update: jest.fn(),
+  setItem: jest.fn(),
+  removeItem: jest.fn(),
   edit: jest.fn(),
```
kiruys, line 17: much more clear like this. Before I was already wondering what update meant

src/signals/incident/components/form/MapSelectors/Asset/Intro/Intro.tsx
```diff
@@ -38,9 +38,7 @@ const Intro = () => {
   const center =
     coordinates || (configuration.map.options.center as LatLngTuple)
 
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const mapOptions = useMemo<MapOptions>(
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
     () => ({
```
kiruys, line 42: 👍

src/signals/incident/components/form/MapSelectors/Caterpillar/CaterpillarLayer/CaterpillarLayer.tsx
```diff
@@ -1,42 +1,36 @@
...
       const feature = feat as Feature
-
-      const [lng, lat] = feature.geometry.coordinates
+      const coordinates = featureTolocation(feature.geometry as Geometrie)
       // Caterpillar layer renders only a single feature type (oak tree)
-      const featureType = meta.featureTypes[0]
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const featureType = meta.featureTypes.find(
+        ({ label }) => label === 'Eikenboom'
+      )!
```
kiruys, line 30: Is this a safe way to identify the Eikenboom featureType? If anything from the feature type would change, it would be the label. I would suggest to take the typeValue

janjaap: Good catch. Very true! I'll change it.

kiruys (approved):

PR #1928, opened 2021-12-16
(Fix) Asset selector position

janjaap:
This PR contains a fix for the `AssetSelector` position on the screen after the page has been scrolled down. This would render the portal an amount of pixels up instead of filling the screen top to bottom and left to right.


kiruys (approved):
👍

PR #1929, opened 2021-12-17
[SIG-4182] Object map auto suggest

janjaap:
This PR adds the `PDOKAutoSuggest` component to the `AssetSelect` component.


src/components/AutoSuggest/AutoSuggest.tsx
```diff
@@ -289,7 +278,7 @@ const AutoSuggest: FC<AutoSuggestProps> = ({
         role="combobox"
       >
         <StyledInput
-          aria-activedescendant={activeId}
+          aria-activedescendant={activeId.toString() ?? null}
           aria-autocomplete="list"
```
kiruys, line 282: this can never be null, right? since on line 82 activeId is always defined

src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
```diff
@@ -68,14 +68,6 @@ describe('components/PDOKAutoSuggest', () => {
...
 
   describe('municipality', () => {
-    it('should call fetch without municipality by default', async () => {
-      await renderAndSearch()
-      expect(fetch).toHaveBeenCalledWith(
-        expect.not.stringContaining(municipalityQs),
-        expect.objectContaining({ method: 'GET' })
-      )
-    })
-
```
kiruys, line 78: Is this assumption no longer valid?

src/components/AutoSuggest/AutoSuggest.tsx
janjaap: Good catch. I forgot to revert that change.

src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
janjaap: No, the default municipality is always Amsterdam

src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
```diff
@@ -131,14 +135,17 @@ const Selector = () => {
   const [showSelectionPanel, setShowSelectionPanel] = useState(true)
   const [pinMarker, setPinMarker] = useState<MarkerType>()
   const [map, setMap] = useState<MapType>()
+  const addressValue = address ? formatAddress(address) : ''
+
   const showMarker =
-    coordinates && (!selection || selection.type === UNREGISTERED_TYPE)
+    coordinates !== undefined &&
+    (!selection || selection.type === UNREGISTERED_TYPE)
```
kiruys, line 142: is the `!== undefined` really necessary here?

kiruys:
Nice work, many tests for this feature! A small thing that is missing however, is if you are in the summary and do 'Wijzig melding', the address is not available in the address field.

kiruys (changes requested):
see separate comments

src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
janjaap: Not necessarily, no.

kiruys (approved):

PR #1933, opened 2021-12-21
[SIG-4091] Restructure summary page

janjaap:
Removes columns, adds margins and section counters.


kiruys (approved):
looks good! nice that you removed the dependency on being logged in or not.

PR #1944, opened 2021-12-23
[SIG-4095] Move location map

janjaap:
This PR:
- Removes the map from step 1 in the incident form
- Adds the `location` prop to all configurations for step 2
- Adds an extra definitions file for `straatverlichting` and `klokken`
- Moves the map input question up in the order of questions so that it's always the first item in the list
- Adjusts the `SelectionPanel` component so that the checkbox and input field are only rendered when a list of feature types is available
- Removes the `autoFocus` attribute from fields in step two
- Rewrites the `checkVisibility` service for readablity


src/signals/incident/services/checkVisibility/index.js
```diff
@@ -4,21 +4,40 @@ import isEqual from 'lodash/isEqual'
...
+      return true
+  }
+
+  return false
+}
+
+const comparisonFuncs = {
+  ifAllOf: Array.prototype.every,
+  ifOneOf: Array.prototype.some,
+}
```
kiruys, line 40: A big improvement 🎉, although still quite complex

kiruys:
What is strange in the interaction now, is that when entering the asset select, the map is zoomed in on dam square. The easiest way to go to a neighborhood outside the center is to type in the address. As a result, the checkbox that the container is not on the map is already checked.

janjaap:
> What is strange in the interaction now, is that when entering the asset select, the map is zoomed in on dam square. The easiest way to go to a neighborhood outside the center is to type in the address. As a result, the checkbox that the container is not on the map is already checked.

Indeed, intended behavior.

src/signals/incident/components/form/MapSelectors/Asset/Summary/Summary.tsx
```diff
@@ -36,7 +36,7 @@ const Summary: FC = () => {
   const { description } =
     meta.featureTypes.find(({ typeValue }) => typeValue === type) ?? {}
 
-  const summaryDescription = `${description}${id ? ` - ${id}` : ''}`
+  const summaryDescription = [description, id].filter(Boolean).join(' - ')
   let summaryAddress = coordinates ? 'Locatie is gepind op de kaart' : ''
```
kiruys, line 40: This change does not make it more readable to me.

src/signals/incident/components/form/index.ts
```diff
@@ -38,7 +37,6 @@ export default {
   HandlingMessage,
   HiddenInput,
   LinkButton,
-  MapInput,
   MapSelect,
```
kiruys, line 40: is it possible to remove the MapInput component as well?

janjaap: No, it's used in the incident detail page.

src/signals/incident/components/form/MapSelectors/Asset/Summary/Summary.tsx
janjaap: What would you suggest?

src/signals/incident/definitions/wizard-step-2-vulaan/civieleConstructies.test.ts
```diff
@@ -0,0 +1,9 @@
+import civieleConstructies from './civieleConstructies'
+
+describe('definition civieleConstructiesn', () => {
```
kiruys, line 3: typo

kiruys:
> > What is strange in the interaction now, is that when entering the asset select, the map is zoomed in on dam square. The easiest way to go to a neighborhood outside the center is to type in the address. As a result, the checkbox that the container is not on the map is already checked.
> 
> Indeed, intended behavior.

I would make more sense to show a zoomed out map in a first step

src/signals/incident/components/form/MapSelectors/Asset/Summary/Summary.tsx
kiruys: to keep it the way it was

janjaap: That's not possible, because in some cases, `description` is `undefined`. That means we need to filter it out. And that's exactly what this change does.

kiruys (approved):

PR #1949, opened 2021-12-24
[SIG-4273] Request appropriate dataset for openbare straatverlichting

janjaap:
This PR adds a property equalizer filter for both clocks and street lights WFS calls. This ensures that only the required data is provided and not the entire dataset for both clocks and street light.


kiruys (approved):
Awesome!

PR #1950, opened 2021-12-24
[SIG-4260] Fix map center

janjaap:
This PR:
- contains a fix for setting the map center for a location that only has coordinates and no address
- renames functions in the `map-location` module so that they correctly reflect their purpose and return value


kiruys (approved):
Ja die stond nog op het lijstje. Beter zo

PR #1951, opened 2021-12-24
[SIG-4001] Error message component already in Dom to help screen readers

kiruys:
FileUpload, FileInput, textarea, and FormField component


src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -10,8 +10,15 @@ import {
...
-}))`
+}))<{ invalid: boolean }>`
   margin-top: 0;
+  white-space: pre-wrap;
+  display: none;
+  ${({ invalid }) =>
+    invalid &&
+    css`
+      display: flex;
+    `}
```
janjaap, line 21: This change won't solve the issue at hand; the point is that the element that holds the error message, should be present in the DOM. Setting its `display` value to 'none' will effectively hide it for browsers and screen readers which defeats the purpose.
See https://www.w3.org/WAI/WCAG21/working-examples/aria-role-status-searchresults/ for an example.

janjaap (changes requested):

src/components/FileInput/FileInput.tsx
```diff
@@ -84,14 +84,11 @@ const FileInput: FunctionComponent<FileInputProps> = ({
...
-          <StyledErrorMessage
-            key={index}
-            data-testid="fileInputError"
-            message={message}
-          />
-        ))}
+      <StyledErrorMessage
+        data-testid="fileInputError"
+        message={errorMessages.join('\n') || ''}
+        invalid={errorMessages.length > 0}
```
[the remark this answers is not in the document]
kiruys: invalid props is removed

src/signals/incident/components/form/FormField/FormField.tsx
```diff
@@ -81,59 +81,60 @@ const FormField: FunctionComponent<FormFieldProps> = ({
...
+                      .requiredLength
+                  )} tekens ingevoerd`
+                : ''
+            }
+          />
+
+          <ErrorMessage
+            data-testid="customError"
+            message={hasError('custom') ? getError('custom') : ''}
+          />
```
janjaap, line 136: These only need to be rendered when there is an actual error to display.

src/components/TextArea/TextArea.tsx
```diff
@@ -78,13 +78,13 @@ const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
         )}
-
-        {errorMessage && (
+        <div role="status">
           <StyledErrorMessage
             id="textareaErrorMessage"
-            message={errorMessage}
+            data-testid="textareaError"
+            message={errorMessage || ''}
           />
```
janjaap, line 86: This component should only be rendered when there is an actual error message that needs to be shown.

janjaap (changes requested):

kiruys:
Created new PR to fix this

PR #1953, opened 2022-01-04
[SIG 4095] subtasks move location map

kiruys:


src/components/LocationMarker/index.js
```diff
@@ -38,19 +38,30 @@ const LocationMarker = ({ geolocation }) => {
 
   const { accuracy, latitude, longitude } = geolocation
 
+  const removeLocationDot = () => {
+    locationDot.remove()
+  }
+
```
janjaap, line 44: This function can be moved out of the scope of the `LocationMarker` component

src/components/LocationMarker/index.js
```diff
@@ -38,19 +38,30 @@ const LocationMarker = ({ geolocation }) => {
...
   useEffect(() => {
     if (!mapInstance) return undefined
 
+    const addLocationDot = () => {
+      locationDot.setLatLng([latitude, longitude])
+      locationDot.addTo(mapInstance)
+    }
+
```
janjaap, line 52: Either move this function out of the scope of the `LocationMarker` component or wrap it in a `useCallback`.

src/components/LocationMarker/LocationMarker.test.js
```diff
@@ -53,14 +70,17 @@ describe('components/LocationMarker', () => {
     ])
     expect(Leaflet.Circle.prototype.setRadius).toHaveBeenCalledWith(accuracy)
 
+    act(() => {
+      fireEvent.click(screen.queryByTestId('gpsButton'))
+    })
+
```
janjaap, line 76: No need to wrap a call to `fireEvent` in `act`. Also, the preferred way of handling user interaction, is to use `userEvent` instead of `fireEvent`.

janjaap (commented):
Looks good. Just a couple of comments.

src/components/LocationMarker/LocationMarker.test.js
kiruys: ✅

src/components/LocationMarker/index.js
kiruys: ✅

src/components/LocationMarker/index.js
kiruys: ✅

PR #1956, opened 2022-01-08
Use id instead of _display to avoid 'SIA'

boris-arkenaar:
closes signalen/frontend#159

### DIFF
--- a/src/signals/incident/containers/IncidentReplyContainer/__tests__/IncidentReplyContainer.test.tsx
+++ b/src/signals/incident/containers/IncidentReplyContainer/__tests__/IncidentReplyContainer.test.tsx
@@ -47,7 +47,7 @@ describe('IncidentReplyContainer', () => {
       await waitFor(() => {
         screen.getByRole('heading', { name: 'Aanvullende informatie' })
         screen.getByRole('heading', { name: 'Uw melding' })
-        screen.getByText('Nummer: SIA-1234')
+        screen.getByText('Nummer: 1234')
         screen.getByText('Gemeld op: 26 juli 2021, 17:43 uur')
         screen.getByRole('textbox', { name: 'Wat voor kleur heeft de auto?' })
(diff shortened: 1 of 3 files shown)


src/types/api/public-incident.ts
```diff
@@ -6,6 +6,7 @@ enum State {
 }
 export interface PublicIncident {
   _display: string
+  id: number
```
kiruys, line 9: More straightforward if _display is removed all together then.

src/signals/incident/containers/IncidentReplyContainer/__tests__/IncidentReplyContainer.test.tsx
```diff
@@ -47,7 +47,7 @@ describe('IncidentReplyContainer', () => {
       await waitFor(() => {
         screen.getByRole('heading', { name: 'Aanvullende informatie' })
         screen.getByRole('heading', { name: 'Uw melding' })
-        screen.getByText('Nummer: SIA-1234')
```
janjaap, line 50: I'm not sure if this particular format is used anywhere else in the whole process of incident management. Can we maybe move the `SIA-` prefix to the configuration?

boris-arkenaar: You mean, the `SIA-` prefix is used in other parts of the app, for Amsterdam, but configurable? And so it should be

janjaap: The prefix is indeed used in the administration of Amsterdam, so it needs to be configurable.

janjaap:
@boris-arkenaar Can we close this PR and move the responsibility for providing the correct label to the backend?

boris-arkenaar:
Yes, let's close it!

PR #1958, opened 2022-01-08
Add small crs value for use with bbox query string

boris-arkenaar:
Necessary to make the following wfs service work:

`https://data.haarlem.nl/geoserver/wfs?request=GetFeature&service=wfs&outputFormat=json&typeNames=gemeentehaarlem:bor_afvalinzamel&version=2.0.0&srsName={srsName}&bbox={west},{south},{east},{north},{srsNameSmall}`

### DIFF
--- a/src/signals/incident/components/form/MapSelectors/Asset/Selector/WfsLayer/WfsLayer.tsx
+++ b/src/signals/incident/components/form/MapSelectors/Asset/Selector/WfsLayer/WfsLayer.tsx
@@ -13,8 +13,8 @@ import useLayerVisible from '../../../hooks/useLayerVisible'
 import type { DataLayerProps } from '../../types'
 import { NO_DATA, WfsDataProvider } from './context'
 
-const SRS_NAME = 'urn:ogc:def:crs:EPSG::4326'
-const SRS_NAME_SMALL = 'EPSG:4326'
+export const SRS_NAME = 'urn:ogc:def:crs:EPSG::4326'
+export const SRS_NAME_SMALL = 'EPSG:4326'
 
 interface Bbox {
   east: string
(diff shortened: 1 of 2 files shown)


kiruys (approved):
Maybe worth while to add this to the Wfs layer test? That would also serve as documentation.

src/signals/incident/components/form/MapSelectors/Asset/Selector/WfsLayer/WfsLayer.tsx
```diff
@@ -14,6 +14,7 @@ import type { DataLayerProps } from '../../types'
 import { NO_DATA, WfsDataProvider } from './context'
 
 const SRS_NAME = 'urn:ogc:def:crs:EPSG::4326'
+const SRS_NAME_SMALL = 'EPSG:4326'
```
janjaap, line 17: What makes this differentiation necessary?

boris-arkenaar: As per the url shown in the description above, only this value seems to work, not the long one.

janjaap: To me, the name `srsNameSmall` doesn't convey any meaning; without documentation it is not clear what the difference is and why its introduction is required. I would strongly suggest looking into using the current `wfsFilter` value. The WFS standard offers a lot of filter options. There has to be an option that allows us to query GeoServer services without having to add parameters for each and every exception.

boris-arkenaar: I agree. For me WFS services are just always very difficult to get working. Maybe it's this specific service of Haarlem not allowing everything. I'll discuss with them.

boris-arkenaar: How about this. One SRS value that seems to work everywhere.

janjaap: Looks good. Seems to work as well 😄

janjaap (approved):

PR #1960, opened 2022-01-10
[SIG 4209] Only show link when phone number is visible

kiruys:
checks if phone number value of melding reporter has astrix.
If so, value is rendered in a link.
If not, value is rendered as text.

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/Detail/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/Detail/index.js
@@ -80,10 +80,10 @@ const Detail = ({ attachments, context }) => {
       ),
     [context?.near?.signal_count]
   )
-  const showPhone = useMemo(
-    () => !/[ *]/.test(incident.reporter.phone.replace(/ /g, '')),
-    [incident.reporter.phone]
-  )
+  const showPhone = useMemo(() => {
+    const phoneNumber = incident.reporter.phone?.replace(/ /g, '')
+    return !/[ *]/.test(phoneNumber)
+  }, [incident.reporter.phone])
 
   return (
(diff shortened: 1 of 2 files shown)


src/signals/incident-management/containers/IncidentDetail/components/Detail/index.js
```diff
@@ -80,6 +80,10 @@ const Detail = ({ attachments, context }) => {
       ),
     [context?.near?.signal_count]
   )
+  const showPhone = useMemo(
+    () => /^\d+$/.test(incident.reporter.phone),
```
janjaap, line 84: This will not work. Valid phone number values are also `+312012345678`,  `020-12345678` and `020 12345678`

janjaap (changes requested):
The proposed regex won't cover all use-cases

src/signals/incident-management/containers/IncidentDetail/components/Detail/index.js
kiruys: Herjen said that only numbers could be present. Perhaps I should have tested that...

janjaap (approved):

src/signals/incident-management/containers/IncidentDetail/components/Detail/index.js
```diff
@@ -80,6 +80,10 @@ const Detail = ({ attachments, context }) => {
       ),
     [context?.near?.signal_count]
   )
+  const showPhone = useMemo(() => {
+    const phoneNumber = incident.reporter.phone?.replace(/ /g, '')
```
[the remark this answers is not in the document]
kiruys: I will change it for the sake of performance

PR #1961, opened 2022-01-11
[SIG 4291] Update npm packages

kiruys:
- Updated e2e packages
- Updated packages


package.json
```diff
@@ -119,7 +119,7 @@
     "@typescript-eslint/eslint-plugin": "^4.14.1",
     "@typescript-eslint/parser": "^4.23.0",
     "@wojtekmaj/enzyme-adapter-react-17": "^0.6.0",
-    "add-asset-html-webpack-plugin": "^3.2.0",
+    "add-asset-html-webpack-plugin": "^2.1.2",
```
janjaap, line 122: Why the downgrade?

kiruys: That's strange indeed

kiruys: that was the result from npm audit fix.. I have updated the add-asset-html-webpack-plugin. However, putting back the eslint-nibble update was not successful.

janjaap (approved):

PR #1975, opened 2022-01-18
[SIG-4323] Move images to webroot

janjaap:
This PR moves all of the SVG files, that are referenced by their URL, to the web root. See the ADR for details.


docs/adr/0013-move-images-to-webroot.md
```diff
@@ -0,0 +1,28 @@
+# Move images to webroot
+
+Date: 2022-01-24
+
```
kiruys, line 4: nice!

kiruys (approved):

boris-arkenaar (approved):

PR #1995, opened 2022-01-22
Clean up config

boris-arkenaar:
Update description for `staticUrl` config.


janjaap:
@boris-arkenaar Can you update the PR, describing why the proposed change is required?

janjaap:
@boris-arkenaar The branch has conflicts with the base branch. Can you resolve them?

src/signals/incident/definitions/wizard-step-2-vulaan/afval-container.ts
```diff
@@ -44,7 +48,7 @@ export const controls = {
       pathMerge: 'extra_properties',
       wfsFilter:
         '<PropertyIsEqualTo><PropertyName>status</PropertyName><Literal>1</Literal></PropertyIsEqualTo><BBOX><PropertyName>geometrie</PropertyName><gml:Envelope srsName="{srsName}"><lowerCorner>{west} {south}</lowerCorner><upperCorner>{east} {north}</upperCorner></gml:Envelope></BBOX>',
-      endpoint: configuration.map.layers.containers,
+      endpoint: configuration.map.layers?.containers,
```
janjaap, line 51: The `containers` props is a required prop from the perspective of Amsterdam, which means we're using the wrong type if we need to use optional chaining.

**Food for thought:**
For each of the domains that are used by the municipality of Amsterdam (amsterdam, weesp, amsterdamsebos and app), we can provide different configuration and assets for the acceptance and production environments. We use the [`signalen`](https://github.com/amsterdam/signalen) repository for that purpose.

We could also make it possible to do that with the JSON schema that is used for configuration validation. That way we can use the Amsterdam JSON schema in the `signals-frontend` code base and have types according to the corresponding configuration. We can avoid having to make exceptions for other use-cases which keeps the code clean and clear.

Each individual implementation (VNG) will have to take care of providing its own applicable JSON schema and configuration. The latter should already be the case.

janjaap (approved):

PR #1998, opened 2022-01-24
[SIG-4352] Close button position

janjaap:
This PR contains a change that will position the map close button to the top right of its parent viewercontainer component, effectively positioning the auto suggest field between the GPS button and the map close button.


kiruys (approved):
Better like this!

PR #2003, opened 2022-01-25
[SIG-4274] Optional checked box

janjaap:
This PR removes the direct dependency between selecting a location on the map and checking the 'The object is not on the map' box in the selection panel.


kiruys (approved):
👍

PR #2012, opened 2022-02-01
[SIG-4274] Object on map label

janjaap:
This PR contains a change that simplifies the logic that shows the location description in the `Summary` component. For it to work, the "De container staat niet op de kaart" checkbox needed to clear the selection in the state when unchecked. That way it became possible to distinguish between no selected object and a selected, unknown, object.


kiruys (approved):
applause for adding these tests!
seems to work nicely.

PR #2013, opened 2022-02-03
[SIG-427] Object not on map fix

janjaap:
This PR fixes an issue where selecting a location on the map and ticking the 'not on the map' checkbox conflicted. To make it work, a new object type was added and the existing type was renamed. The asset select component now distinguishes `OBJECT_UNKNOWN` (was `UNREGISTERED_TYPE`) and `OBJECT_NOT_ON_MAP`. Also, because the map can be closed without having selected coordinates, the submit button

Please refer to the individual commits in this PR for all the information necessary for review.


janjaap:
Closing; going to make it work in another, less convoluted, way.

PR #2026, opened 2022-02-09
[SIG-4122] Moving assets

janjaap:
This PR removes some of the config entries that hold references to image locations. Those location can be implied by always having assets in the web root. This way, there doesn't need to be a configuration entry per asset and individual implementations of the SIA application can overwrite those images (like the app's logo) by mounting an image with the same name in the `signals-frontend` container. The same approach has been taken with map icons (see https://github.com/Amsterdam/signals-frontend/pull/1975 for reference).

The proposed change requires implementation (like [Den Bosch](https://meldingen.s-hertogenbosch.nl)) to provide their own assets and to have those assets have the exact same name as the ones provided in the `signals-frontend` repository. Failing to do so will expose the images in this repo. Like [the Amsterdam logo](https://meldingen.s-hertogenbosch.nl/assets/images/amsterdam-logo.svg).

Each implementation will have to provide the following images:
- `favicon.png`
- `icon_180x180.png`
- `icon_192x192.png`


src/components/Logo/index.js
```diff
@@ -54,7 +54,7 @@ export const Logo = ({ tall, ...props }) => (
       data-testid="logo"
       alt={configuration.language.logoDescription}
       tall={tall}
-      src={configuration.logo.url}
+      src="/assets/images/logo.svg"
```
boris-arkenaar, line 57: This way only svg logos can be used, not png logos, right?

janjaap: Yeah, I realized that. I think that the logo URL is something we cannot change that easily. If at all.

src/components/Logo/__tests__/Logo.test.js
```diff
@@ -36,9 +35,6 @@ describe('components/Logo', () => {
     expect(
       container.querySelector(`a[href="${configuration.links.home}"]`)
     ).toBeInTheDocument()
-    expect(
-      container.querySelector(`img[src="${configuration.logo.url}"]`)
```
boris-arkenaar, line 40: Shouldn't we keep these tests now?

kiruys: the url definition on line 18 should be removed, but we could put in the fixed logo url..

kiruys (approved):

boris-arkenaar:
BREAKING: configurations need to be updated.

src/components/Logo/__tests__/Logo.test.js
janjaap: I'll put the tests back. Good catch.

PR #2027, opened 2022-02-09
(Fix) Function reference

janjaap:
Fixes an issue where an optional prop was referenced directly as a function instead of checking if the prop is a function.


jpoppe (approved):
👍🏻

PR #2036, opened 2022-02-15
[SIG-4368] New map detail panel

janjaap:
This PR:
- Renames the `SelectionPanel` component to `DetailPanel
- Removes the `@amsterdam/arm-core` `MapPanel` component and the dependency on its context provider
- Moves the responsibility of handling the legend panel visibility to the `LegendPanel` component itself
- Moves all styled components to their corresponding `styled.tsx` files and
- Renames 'Legenda' to 'Uitleg'
- Rewrites the legend button interaction so that a click will only influence the legend panel and not also the detail panel


src/signals/incident/components/form/MapSelectors/Asset/Selector/LegendPanel/LegendPanel.test.tsx
```diff
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright (C) 2021 Gemeente Amsterdam
 import { render, screen } from '@testing-library/react'
+import 'jest-styled-components'
+
```
kiruys, line 5: what is this import for?

janjaap: This wil export the `toHaveStyleRule` assertion as used in the added test cases in this suite

kiruys (approved):

PR #2038, opened 2022-02-16
[SIG-4426] Move autosuggest

janjaap:
This PR moves the `PDOKAutoSuggest` component from the map (asset `Selector`) to the detail panel. Note that this change isn't the full set of changes that are required to completely move the functionality. Just an intermediate step.


kiruys:
the back shadow of the address input can be removed.

kiruys (approved):

PR #2041, opened 2022-02-16
Remove unnecessary `sort`

boris-arkenaar:
The lat lng conversion function unnecessarily used `sort` while `reverse` alone is enough to reverse lat and lng.
The `sort` caused a problem when the lng would be alphabetically 'bigger' than the lat.


boris-arkenaar:
Convert to float first. Added UTs.

janjaap (approved):

boris-arkenaar:
closes https://github.com/Signalen/frontend/issues/168

PR #2042, opened 2022-02-16
[SIG-4428] Full-screen auto suggest

janjaap:
This PR:
- Renders a full-screen address panel with auto suggest field on smaller screens
- Adds the [react-responsive](https://www.npmjs.com/package/react-responsive) dependency. This is meant to replace the [react-media](https://www.npmjs.com/package/react-media)`dep, which is not maintained anymore.
- Adjusts the `AutoSuggest` component so that it can render a list of suggestions (default) or return the list via callback so that implementing components can render the list in their own way.

### DIFF
--- a/src/components/AutoSuggest/AutoSuggest.tsx
+++ b/src/components/AutoSuggest/AutoSuggest.tsx
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright (C) 2020 - 2021 Gemeente Amsterdam
-import { useCallback, useEffect, useState, useRef } from 'react'
+import { useCallback, useEffect, useState, useRef, useMemo } from 'react'
 
 import type { FC } from 'react'
 import type { PdokResponse } from 'shared/services/map-location'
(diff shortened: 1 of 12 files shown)


src/components/AutoSuggest/AutoSuggest.tsx
```diff
@@ -75,6 +86,10 @@ const AutoSuggest: FC<AutoSuggestProps> = ({
   const clearInput = useCallback(() => {
     if (inputRef.current) {
       inputRef.current.value = ''
+
+      if (!showInlineList) {
+        inputRef.current.focus()
```
kiruys, line 91: is it possible to avoid 2 deep nesting here?

janjaap: Not really, because the field should only get the focus when `inputRef` has a value AND `showInlineList` equals to false. I could flatten the assertion like so `if (inputRef.current && !showInlineList)`, but that doesn't add much to the readability.

kiruys: ok

kiruys (approved):

PR #2046, opened 2022-02-21
[SIG-4292] Add datetime question

janjaap:
This PR:
- Adds a `timestamp` prop to the incident reducer. This prop replaces the props `incident_date`, `incident_time_hours`, `incident_time_minutes` and `datetime`
- Stores the incident date and time as a timestamp value instead of separate string values
- Separates the `DateTime` component render logic into a separate component. Keeping it in the context of a `react-reactive-form` function doesn't allow for using hooks.
- Adds configuration for several subcategories that need to show the `DateTime` component
- Enables the prop `ignoreVisibility` for the form controls `meta` object; this prop will override the visibility checks (that are hidden somewhere deep down in a dark corner of the application's logic) so that an element that is linked to a `null` value prop in the global state, will still be rendered. This is necessary, because `react-reactive-form` refuses tot render any element that gets `null` as a value :shrug: 

To do:
- [x] Enable for Overlast Bedrijven en Horeca
- [x] Enable for Openbaar groen en water
- [x] Enable for Overlast in de openbare ruimte
- [x] Enable for Wonen


src/signals/incident/components/form/DateTimeInput/styled.tsx
```diff
@@ -40,6 +40,6 @@ export const TimeWrapper = styled.div`
   }
 
   select {
-    width: 70px; /* fixed value from design */
+    width: 75px; /* fixed value from design */
```
kiruys, line 43: magic number ;-)

janjaap: That is why the comment is there; it explains where the value comes from.

src/signals/incident/components/IncidentPreview/components/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,43 @@
...
+  value: Incident['timestamp']
+  incident?: Incident
+}
+
+const getValue = (value: Incident['timestamp'], incident?: Incident) => {
+  if (value === null) {
+    return 'Nu'
+  }
+
+  if (!value || !incident) {
```
kiruys, line 20: shouldn't this statement come before the check if `value === null`?

src/signals/incident/components/IncidentPreview/components/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,43 @@
...
+import type { FC } from 'react'
+
+interface DateTimeProps {
+  value: Incident['timestamp']
+  incident?: Incident
+}
+
+const getValue = (value: Incident['timestamp'], incident?: Incident) => {
+  if (value === null) {
+    return 'Nu'
```
kiruys, line 17: is it possible to use constants for 'Nu' and 'Eerder'?

src/signals/incident/components/IncidentPreview/components/DateTime/DateTime.tsx
janjaap: No, because the value `null` is an intended value. Having that value in the state is equivalent to now (`Date.now()`)

src/signals/incident/components/IncidentPreview/components/DateTime/DateTime.tsx
janjaap: Ah, good one. I'll make a type for that.

src/types/incident.ts
```diff
@@ -21,15 +21,12 @@ type ExtraProps = {
 export interface Incident extends Record<string, any>, ExtraProps {
   category: string
   classification: Classification | null
-  datetime: Datetime
+  dateTime: number | null
```
kiruys, line 24: why is this a number and not a datetime type?

src/signals/incident/components/form/DateTimeInput/styled.tsx
kiruys: You don't need to change this, but to me the remark 'from design' is not really informative. I would assume that such a value is always included to conform with the design.

src/types/incident.ts
janjaap: Because `Datetime` type is an object that is constructed to meet the data structure that is used in radio button list components. Using an object to determine something as basic as a point in time is superfluous. Timestamps can be used to calculate diffs and can be formatted to readable strings.

kiruys (approved):

PR #2048, opened 2022-02-21
(Chore) Show autosuggest clear button

janjaap:
This PR:
- always shows the auto suggest input clear button when user types in the input field
- makes sure that the list of auto suggest options is removed whenever the clear button is clicked


kiruys:
Schermafbeelding 2022-02-21 om 16 28 53
the close button needs to be smaller. when the input has focus, the erase button sticks out of the border

janjaap:
@kiruys Good catch on the focus styling. Fixed it by making the button smaller so that it doesn't overlap the focus ring in Chrome.

kiruys (approved):

kiruys:
I approved. However I really don't prefer the position absolute solution.

PR #2051, opened 2022-02-22
Postpone submit on loading questions from backend

boris-arkenaar:
closes Signalen/product-steering#93

Make sure to postpone submitting in step 1 in case the extra questions are coming from the backend and are still loading.

### DIFF
--- a/src/signals/incident/components/IncidentForm/index.js
+++ b/src/signals/incident/components/IncidentForm/index.js
@@ -33,7 +33,9 @@ class IncidentForm extends Component {
       return null
     }
 
-    const loading = get(props, props.postponeSubmitWhenLoading)
+    const loading = Array.isArray(props.postponeSubmitWhenLoading)
+      ? props.postponeSubmitWhenLoading.some((prop) => get(props, prop))
+      : get(props, props.postponeSubmitWhenLoading)
     if (loading !== prevState.loading) {
       return {
         loading,
@@ -219,7 +221,10 @@ IncidentForm.propTypes = {
   removeQuestionData: PropTypes.func.isRequired,
   updateIncident: PropTypes.func.isRequired,
   createIncident: PropTypes.func.isRequired,
-  postponeSubmitWhenLoading: PropTypes.string,
+  postponeSubmitWhenLoading: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.arrayOf(PropTypes.string),
+  ]),
 }
 
 export default IncidentForm
--- a/src/signals/incident/components/IncidentForm/index.js
+++ b/src/signals/incident/components/IncidentForm/index.js
@@ -3,7 +3,6 @@
 import { createRef, Component } from 'react'
 import PropTypes from 'prop-types'
 import { FormGenerator } from 'react-reactive-form'
-import get from 'lodash/get'
 import isEqual from 'lodash/isEqual'
 import isObject from 'lodash/isObject'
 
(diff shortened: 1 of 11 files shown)


src/signals/incident/components/IncidentForm/index.js
```diff
@@ -33,7 +33,9 @@ class IncidentForm extends Component {
       return null
     }
 
-    const loading = get(props, props.postponeSubmitWhenLoading)
+    const loading = Array.isArray(props.postponeSubmitWhenLoading)
+      ? props.postponeSubmitWhenLoading.some((prop) => get(props, prop))
+      : get(props, props.postponeSubmitWhenLoading)
     if (loading !== prevState.loading) {
```
kiruys, line 39: The code becomes overly complex this way. I would prefer to always use an array.

src/signals/incident/components/IncidentForm/index.test.js
```diff
@@ -285,6 +285,97 @@ describe('<IncidentForm />', () => {
 
         expect(nextSpy).toHaveBeenCalledTimes(2)
       })
+
+      it('should postpone submit when postponeSubmitWhenLoading is defined for loadingQuestions', () => {
```
kiruys, line 289: I don't understand what this means. Does it mean "it postpones submit in case questions need to be fetched from the backend but haven't loaded yet"?

src/signals/incident/components/IncidentForm/index.test.js
```diff
@@ -285,6 +285,97 @@ describe('<IncidentForm />', () => {
...
+        expect(nextSpy).toHaveBeenCalledTimes(1)
+        userEvent.click(screen.getByText(mockForm.nextButtonLabel))
+        expect(nextSpy).toHaveBeenCalledTimes(1)
+
+        renderIncidentForm(props, rerender)
+
+        expect(nextSpy).toHaveBeenCalledTimes(2)
+      })
+
+      it('should postpone submit when postponeSubmitWhenLoading is defined for both loadingClassification and loadingQuestions', () => {
```
kiruys, line 309: Same here: Does this mean: "it postpones submit until the classification and questions have loaded" ?

src/signals/incident/containers/IncidentContainer/reducer.test.js
```diff
@@ -197,10 +204,22 @@ describe('signals/incident/containers/IncidentContainer/reducer', () => {
             handling_message,
           },
           loadingClassification: false,
+          loadingQuestions: false,
           classificationPrediction: classification,
         })
       })
 
+      it('sets loadingQuestions with feature flag enabled', () => {
```
kiruys, line 212: We can see that the feature flag is enabled. What does it mean? That it is specified that questions are retrieved from the backend?

src/signals/incident/definitions/wizard.ts
```diff
@@ -28,7 +28,7 @@ export type WizardSection = {
     formFactory?: any
     label?: string
     subHeader?: string
-    postponeSubmitWhenLoading?: string
+    postponeSubmitWhenLoading?: string | string[]
```
kiruys, line 31: As I mentioned before: I would prefer to always use an array

janjaap (changes requested):
Seems to me that the proposed solution is quit fragile and can be better handled by setting the `loading` prop in the global state. This boolean prop can be used to check if the (or any) request is still pending. Using the global state is also more future-proof than hacking it into the react-reactive-form way of working.

src/signals/incident/components/IncidentForm/index.test.js
boris-arkenaar: It means it should postpone in case `loadingQuestions` is present in `postponeSubmitWhenLoading`.

boris-arkenaar:
That sounds good @janjaap I'll look into that.

src/signals/incident/containers/IncidentContainer/saga.js
```diff
@@ -66,15 +65,16 @@ export function* getClassification(action) {
...
+  const { category, subcategory } = getIncidentClassification(
+    incident,
+    action.payload
+  )
   if (
     !configuration.featureFlags.fetchQuestionsFromBackend ||
     !category ||
     !subcategory
   ) {
+    yield put(getQuestionsError())
```
janjaap, line 77: I get that the state needs to be reset, but dispatching the `getQuestionsError` when the feature flag `fetchQuestionsFromBackend` is 'false' doesn't seem right; an error didn't occur, functionality just isn't used. Dispatching an error action doesn't give a developer the right feedback in the Redux dev console. Also, semantically it's not the correct thing to do.
I would suggest dispatching something like `setLoadingData()`.

janjaap (changes requested):

janjaap (approved):

PR #2052, opened 2022-02-22
Use zoom levels from config for asset select

boris-arkenaar:
closes Signalen/frontend#166

Have the Asset Select map use the min and max zoom levels as defined in the config.

### DIFF
--- a/src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
+++ b/src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
@@ -50,7 +50,12 @@ const MAP_CONTAINER_ZOOM_LEVEL: ZoomLevel = {
   max: 13,
 }
 
-const MAP_LOCATION_ZOOM = 14
+export const MAP_LOCATION_ZOOM = Math.min(
+  14,
+  MAP_OPTIONS.maxZoom || Number.POSITIVE_INFINITY
+)
+
+const MAP_NO_LOCATION_ZOOM = MAP_OPTIONS.zoom
 
 const Selector: FC = () => {
   // to be replaced with MOUNT_NODE
@@ -78,12 +83,7 @@ const Selector: FC = () => {
       dragging: true,
       zoomControl: false,
       scrollWheelZoom: true,
-      zoom: coordinates
-        ? Math.min(
-            MAP_LOCATION_ZOOM,
-            MAP_OPTIONS.maxZoom || Number.POSITIVE_INFINITY
-          )
-        : MAP_OPTIONS.zoom,
+      zoom: coordinates ? MAP_LOCATION_ZOOM : MAP_NO_LOCATION_ZOOM,
     }),
     [center, coordinates]
   )
--- a/src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
+++ b/src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
@@ -50,12 +50,7 @@ const MAP_CONTAINER_ZOOM_LEVEL: ZoomLevel = {
   max: 13,
 }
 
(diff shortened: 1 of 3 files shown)


src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
```diff
@@ -72,14 +71,19 @@ const Selector: FC = () => {
...
-      minZoom: 7,
-      maxZoom: 16,
-      zoom: coordinates ? MAP_LOCATION_ZOOM : MAP_NO_LOCATION_ZOOM,
+      zoom: coordinates
+        ? Math.min(
+            MAP_LOCATION_ZOOM,
+            MAP_OPTIONS.maxZoom || Number.POSITIVE_INFINITY
+          )
+        : MAP_OPTIONS.zoom,
     }),
```
kiruys, line 87: Can you explain why you have dropped the `MAP_NO_LOCATION_ZOOM` const? It would be more readable if the MAP_LOCATION_ZOOM and `MAP_NO_LOCATION_ZOOM` constants are kept and the computation is done over there. Also, I don't see the use of the Math.min part with the Number.POSITIVE_INFINITY. A much simpler and readable solution would be to say ```
const MAP_LOCATION_ZOOM = MAP_OPTIONS.maxZoom ? Math.min(
            MAP_LOCATION_ZOOM,
            MAP_OPTIONS.maxZoom) : MAP_LOCATION_ZOOM 
```

Comment (kiruys) [reply src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx:87]: And the MAP_OPTIONS_NO_LOCATION_ZOOM is replaced by MAP_OPTIONS.zoom. Are you sure that value can be used? In the amsterdam repo, that value is different than the MAP_NO_LOCATION_ZOOM.

Comment (boris-arkenaar) [reply src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx:87]: We could still put it in a const, that's true.

When there is no location set, you want the map to open at the default zoom level. That's `MAP_OPTIONS.zoom`.

Comment (janjaap) [reply src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx:87]: I would also like to see a test for this change

Comment (boris-arkenaar) [reply src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx:87]: That's a bit difficult since it's map options passed on to the `Map` component.

File: src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx [hunk 2]
```diff
@@ -50,8 +50,12 @@ const MAP_CONTAINER_ZOOM_LEVEL: ZoomLevel = {
   max: 13,
 }
 
-const MAP_LOCATION_ZOOM = 14
-const MAP_NO_LOCATION_ZOOM = 9
+export const MAP_LOCATION_ZOOM = Math.min(
```
kiruys, line 53: `export` is not used, right?

src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
```diff
@@ -50,8 +50,12 @@ const MAP_CONTAINER_ZOOM_LEVEL: ZoomLevel = {
...
+  14,
+  MAP_OPTIONS.maxZoom || Number.POSITIVE_INFINITY
+)
+
```
kiruys, line 57: you could add a test for computing `MAP_LOCATION_ZOOM`

src/signals/incident/components/form/MapSelectors/Asset/Selector/Selector.tsx
boris-arkenaar: It is now used in the unit test

janjaap (approved):

PR #2053, opened 2022-02-23
Set maxNativeZoom for PDOK tiles

boris-arkenaar:
closes Signalen/frontend#166

The map went gray when zooming in past level 14.
Setting the `maxNativeZoom` to `14` on the PDOK tiles layer fixes this.

Better solution than https://github.com/Amsterdam/signals-frontend/pull/2052 for Signalen/frontend#166.
https://github.com/Amsterdam/signals-frontend/pull/2052 is still a bug fix though.

### DIFF
--- a/app.base.json
+++ b/app.base.json
@@ -71,7 +71,7 @@
           5.10737
         ]
       ],
-      "maxZoom": 14,
+      "maxZoom": 16,
       "minZoom": 8,
       "zoom": 10
     },
@@ -86,6 +86,7 @@
       ],
       "options": {
         "attribution": "Kaartgegevens &copy; Kadaster",
+        "maxNativeZoom": 14,
         "tms": false
       }
     },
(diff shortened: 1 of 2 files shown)


kiruys (commented):
Why did you change the maxZoom value?

janjaap:
@boris-arkenaar Can't the issue be solved by decreasing the value for `maxZoom` in the individual configs? If a tile server doesn't return tiles beyond a specific zoom level, then each municipality should have their own zoom levels configured.

boris-arkenaar:
@janjaap No, we want to be able to zoom in further than the tile server allows. Specifically for the asset select of containers that are positioned very closely together. Hence, `maxZoom` should stay high (`16`). `maxNativeZoom` is necessary to make sure that beyond the max possible zoom available on the tile server (`14`) leaflt will use that last zoom level and stretch it out, instead of showing gray tiles.

janjaap (approved):

PR #2056, opened 2022-02-25
[SIG 4425] Container-bijplaatsen subcategory activates container map

kiruys:


src/signals/incident/definitions/wizard-step-2-vulaan/afval-container.ts
```diff
@@ -19,6 +19,7 @@ export const controls = {
     meta: {
       ifOneOf: {
         subcategory: [
+          'container-bijplaatsing',
```
janjaap, line 22: This change shouldn't be necessary, since the logic that loads these controls already looks if a subcategory start with the value `container`.

janjaap: In fact, the whole `ifOneOf` should be removable.

kiruys: I wasn't sure if there was a reason for the double check. I will remove the part that checks the ifOneOf array then

janjaap (approved):

PR #2059, opened 2022-03-01
Allow a negative term for municipality

boris-arkenaar:
closes Signalen/frontend#167

For Groningen we ran into a problem.
`configuration.map.municipality` was set to `groningen`. Turns out there is also a municipality called "Midden-Groningen", having the same keyword `groningen` in there. Therefore the autosuggest of the address lookup gave results of both municipalities, while only results from the municipality of Groningen are desired.
Since an exact match on municipality is not possible with PDOK, the workaround could be to exclude the term "midden" in this case. This is done with a dash, as in `-midden`.

Double quotes are put around each term to accommodate spaces in a term, like `den bosch`. Putting double quotes around the term `-midden` would prevent it from being a negative search term. It should be `-midden`, or `-"midden"`. Not `"-midden"`.

## Thougt process

To fix this I chose to not put double quotes around a term in case it begins with a dash.

Other possible solutions could be:

- Never put quotes around a term. Then they need to be put explicitly in the configuration in case the term contains spaces.
- Always put quotes around a term. In case the term begins with a dash, put the dash in front of the quote.

## Final implementation

Finally decided to go for a fix that removes all magic. More power in the configuration.
For clarity, only string values are allowed now. Not an array anymore.

## BREAKING:

- Any configuration setup that has a municipality defined with spaces in it now needs to put double quotes around it.
So, `"den bosch"` should become `"\"den bosch\""`.
- Any configuration setup that has municipality defined as an array needs to convert it into a string, separating each municipality with a space.

### DIFF
--- a/src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
+++ b/src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
@@ -76,17 +76,17 @@ describe('components/PDOKAutoSuggest', () => {
       )
     })
 
-    it('should work with an array for municipality', async () => {
-      await renderAndSearch('Dam', { municipality: ['utrecht', 'amsterdam'] })
+    it('should work with multiple municipalities', async () => {
+      await renderAndSearch('Dam', { municipality: 'utrecht amsterdam' })
       expect(fetch).toHaveBeenCalledWith(
         expect.stringContaining(`${municipalityQs}(utrecht amsterdam)`),
         expect.objectContaining({ method: 'GET' })
       )
     })
 
-    it('should work with quotes', async () => {
+    it('should work with quotes and negations', async () => {
       await renderAndSearch('Dam', {
-        municipality: ['"den bosch"', '-amsterdam'],
+        municipality: '"den bosch" -amsterdam',
       })
       expect(fetch).toHaveBeenCalledWith(
         expect.stringContaining(`${municipalityQs}("den bosch" -amsterdam)`),
@@ -97,36 +97,6 @@ describe('components/PDOKAutoSuggest', () => {
     it('should ignore an empty string', async () => {
       await renderAndSearch('Dam', {
         municipality: '',
-      })
-      expect(fetch).toHaveBeenCalledWith(
-        expect.not.stringContaining(municipalityQs),
-        expect.objectContaining({ method: 'GET' })
-      )
-    })
-
-    it('should ignore an empty array', async () => {
-      await renderAndSearch('Dam', {
-        municipality: [],
-      })
-      expect(fetch).toHaveBeenCalledWith(
-        expect.not.stringContaining(municipalityQs),
-        expect.objectContaining({ method: 'GET' })
-      )
-    })
-
-    it('should ignore an empty string in an array', async () => {
-      await renderAndSearch('Dam', {
-        municipality: ['', 'amsterdam', ''],
-      })
-      expect(fetch).toHaveBeenCalledWith(
-        expect.stringContaining(`${municipalityQs}(amsterdam)`),
-        expect.objectContaining({ method: 'GET' })
-      )
-    })
-
-    it('should ignore an array with only empty strings', async () => {
-      await renderAndSearch('Dam', {
-        municipality: ['', ''],
       })
       expect(fetch).toHaveBeenCalledWith(
         expect.not.stringContaining(municipalityQs),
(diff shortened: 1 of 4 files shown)


janjaap:
@boris-arkenaar The query parameter `gemeentenaam` is hardcoded in the `PDOKAutoSuggest` component. Wouldn't it make more sense to make that parameter configurable? Because, maybe, using the parameter `woonplaatsnaam` instead would solve the Groningen issue.

boris-arkenaar:
@janjaap The municipality has quite a list of 'woonplaatsen'. I think it would be more error prone and cumbersome to test a big list of 'woonplaatsen' than having the municipality 'Groningen' and excluding 'midden'.

boris-arkenaar:
And with each and every one of these 'woonplaatsen' you can have the same issue we now have with the municipality 'groningen'.

janjaap:
> And with each and every one of these 'woonplaatsen' you can have the same issue we now have with the municipality 'groningen'.
Did you test the effect of using `woonplaats` instead of using `gemeentenaam`?

kiruys:
What I miss in this discussion, is that in a couple of months no one will remember why these quotes are added. I think such kinds of changes should be well documented, and that is why jan jaap's solution might be more robust, even though it is less efficient.

boris-arkenaar:
The quotes were there already and @janjaap's solution is not changing that.
If we want to remove all the 'magic' we should go for one string entry in the configuration where you have all the freedom to write a solr expression for 'gemeentenaam'. Optionally we could also allow to set this property ('gemeentenaam', 'woonplaatsnaam') in another configuration entry, but that's not necessary for now.

So, something like:
```
    "municipality": "groningen -midden",
```
or
```
    "municipality": "boxtel \"den bosch\"",
```
instead of:
```
    "municipality": [
      "boxtel",
      "den bosch"
    ],
```

boris-arkenaar:
All magic now removed. This does mean a breaking change for the configuration.

src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
```diff
@@ -71,23 +71,65 @@ describe('components/PDOKAutoSuggest', () => {
...
+        expect.stringContaining(`${municipalityQs}(utrecht amsterdam)`),
         expect.objectContaining({ method: 'GET' })
       )
     })
 
-    it('should not put quotes around a value starting with "-"', async () => {
-      await renderAndSearch('Dam', { municipality: ['utrecht', '-amsterdam'] })
+    it('should work with quotes', async () => {
+      await renderAndSearch('Dam', {
+        municipality: ['"den bosch"', '-amsterdam'],
```
janjaap, line 89: This looks pretty error prone, having to wrap any string with spaces with double quotes. Can't we have this handled by the `PDOKAutoSuggest` component logic?

boris-arkenaar: This way you have the freedom to compose this part of the query string as needed. Otherwise you get magic in the code, like we had before.

janjaap: True, but this construct requires extra documentation to explain how to handle name that have spaces.

boris-arkenaar: Yes, because it also allows for instance a dash, to ignore a search term.
All magic in the config, or all magic in the code, I would say. Half half would be even more confusing.

To keep all magic in the code, we could introduce an optional second config parameter, like `map.municipalityExclude`.

Then you can do something like:

```
"map": {
  "municipality": "groningen",
  "municipalityExclude": "midden"
}
```

To get a query string like:

```
fq=gemeentenaam:("groningen" -"midden")
```

In the code we'll put double quotes around every term, like before. And put a dash in front of each term to exclude, putting those at the end.

boris-arkenaar: Another option, to keep all magic and flexibility in the config, is to only allow a string, not an array, to make it more clear that terms with spaces need quotes.

Like I explained above https://github.com/Amsterdam/signals-frontend/pull/2059#issuecomment-1057830548

janjaap: :thinking: Might indeed be better to use a string instead of an array, since the PDOK service also uses that. Then our implementation is in sync with what PDOK is offering and we don't have to document on top of what PDOK is already doing.

boris-arkenaar: Yes, exactly. I think that would be the cleanest approach.

janjaap (approved):

PR #2060, opened 2022-03-01
(Chore) Allow Browserstack tunnel

janjaap:
The proposed change in this PR allows testing local URLs on iOS with Browserstack. iOS doesn't allow the use of `localhost`, but this can be circumvented by using the bs-local.com domain. See https://www.browserstack.com/docs/live/local-testing/test-using-local-testing for reference on local testing.
Note that, for this change to work, the hosts file on the development machine needs to be changed so that it accepts requests for bs-local.com and routes these to `127.0.0.1`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "prebuild": "npm run build:clean && npm run proj4compile",
     "pretest": "npm run test:clean",
     "prestart:prod": "npm run build",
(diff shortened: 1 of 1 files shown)


kiruys (approved):

kiruys:
nice!

PR #2073, opened 2022-03-07
Remove area_type_code when area is null

boris-arkenaar:
closes Signalen/frontend#118

In case the option 'Niet toegewezen' is selected to filter incidents on district, `area_code` will be `'null'`. `area_type_code` still defaulted to 'district' though. For this filter to work `area_type_code` must also be `'null'`, or not be present at all. In this change I make sure that it is not present when `area_code` is `'null'`.


janjaap:
@boris-arkenaar Test are breaking

janjaap (approved):

PR #2077, opened 2022-03-07
[SIG-4462] Status form alert spacing

janjaap:
Ensures spacing between notification element in the status form


kiruys (commented):
In the ticket it says margin bottom. So is this really the correct change?

janjaap:
The change in this PR covers the use-case of having multiple adjacent elements with spacing in-between them.

kiruys (approved):

PR #2078, opened 2022-03-08
[SIG-4375] Move StatusLayer component

janjaap:
Since the `StatusLayer` component hasn't any dependency on the `WfsLayer` component, nor does it use its context, this PR  moves the component to another folder. That folder, `VectorLayer` will be the context in which other related layer components will be placed.


kiruys:
The choice of 'vectorlayer' is unclear to me. The statuslayer can also stay in its own folder outside the wsf layer.

kiruys (approved):

kiruys (approved):

PR #2098, opened 2022-03-23
Update questions on change of category

boris-arkenaar:
Additional questions from the backend were not updated anymore when you are logged in and change the category manually in the dropdown in the incident form.
This PR fixes that.

### DIFF
--- a/src/signals/incident/containers/IncidentContainer/saga.js
+++ b/src/signals/incident/containers/IncidentContainer/saga.js
@@ -20,6 +20,7 @@ import {
   CREATE_INCIDENT,
   GET_CLASSIFICATION,
   GET_CLASSIFICATION_SUCCESS,
+  UPDATE_INCIDENT,
 } from './constants'
 import {
   createIncidentSuccess,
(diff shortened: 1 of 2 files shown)


kiruys:
When I check out the code, there is an issue when the 'Volgende' button is clicked fast. An error occurs. See screenshot.
Schermafbeelding 2022-03-29 om 09 56 11

src/signals/incident/components/IncidentForm/index.js
```diff
@@ -52,6 +52,8 @@ class IncidentForm extends Component {
       if (this.form.valid) {
         this.setIncident(this.state.formAction)
         this.state.next()
+      } else {
+        this.handleSubmit(null, this.state.next, this.state.formAction)
```
kiruys, line 56: What is the use of this else statement? (and the if (e) in the handleSubmit ?

boris-arkenaar: This makes sure the form is validated and the results are shown on the page. Otherwise, there was a situation when logged in where you'd click the button, you stay on the page, but don't see any feedback. In this case, there's no event, hence the `if (e)` in the `handleSubmit`.

kiruys: I'm not convinced this is necessary, I've tried several things but never get into the else statement.
And it makes the code quite unreadable with the nested ifs.

boris-arkenaar: You are right. The else was indeed not necessary!

kiruys (approved):

PR #2100, opened 2022-03-24
[SIG-4372] Text shown when no contact details are known changed 

SireeKoolenWijkstra:
In summary of call, when there are no known contactdetails, text has changed to "U hebt geen contactgevens ingevuld. We kunnen u niet laten weten wat wij hebben gedaan met uw melding." 

Change in component LinkContainer in IncidentPreview.tsx


kiruys (approved):
Gefeliciteerd!  🎉

PR #2105, opened 2022-03-31
[SIG-4480] placeholder text address field

SireeKoolenWijkstra:
## Signalen

To component StyledPDOKAutosuggest is placeholder with value ' Zoek adres of postcode' added. Both in mobile as desktop version of app is placeholder visible.


kiruys (approved):
Good catch that you spotted the 2 inputs!
Can you adjust the title according to the standard?

PR #2116, opened 2022-04-04
Make the PDOK type configurable

boris-arkenaar:
closes Signalen/product-steering#107

The `type` for the call to PDOK was hardcoded to `adres`. Some municipalities need to be able to have roads as suggestions as well. (There exist roads without any specific address on them.) This PR makes that value configurable.

### DIFF
--- a/src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
+++ b/src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
@@ -13,6 +13,7 @@ const mockResponse = JSON.stringify(JSONResponse)
 
 const onSelect = jest.fn()
 const municipalityQs = 'fq=gemeentenaam:'
+const fieldTypeQs = 'fq=type:'
 const fieldListQs = 'fl='
 
 const renderAndSearch = async (value = 'Dam', props = {}) => {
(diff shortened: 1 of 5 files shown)


src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
```diff
@@ -67,6 +68,34 @@ describe('components/PDOKAutoSuggest', () => {
...
+      await renderAndSearch('Dam', {
+        fieldType: '',
+      })
+      expect(fetch).toHaveBeenCalledWith(
+        expect.not.stringContaining(fieldTypeQs),
+        expect.objectContaining({ method: 'GET' })
+      )
+    })
+  })
+
```
kiruys, line 98: We should also test the response. So in the first test case, addresses are returned, and in the second case addresses and roads.

kiruys, line 98: We should also test the response. So in the first test case, addresses are returned, and in the second case addresses and roads.

src/components/PDOKAutoSuggest/PDOKAutoSuggest.tsx
```diff
@@ -40,16 +41,26 @@ export interface PDOKAutoSuggestProps
 const PDOKAutoSuggest: FC<PDOKAutoSuggestProps> = ({
   fieldList = [],
   municipality = configuration.map.municipality,
+  fieldType = configuration.map.fieldType,
   ...rest
 }) => {
-  const fq = municipality
+  const fieldTypeParam = fieldType
+    ? [['fq', `${fieldTypeFilterName}:(${fieldType})`]]
+    : []
```
kiruys, line 49: Perhaps an idea to have 'adres' as default, and use fieldType when it is something else than 'adres'

kiruys, line 49: Perhaps an idea to have 'adres' as default, and use fieldType when it is something else than 'adres'

boris-arkenaar: The default value of `adres` is set here https://github.com/Amsterdam/signals-frontend/blob/9ba5c6654942fdbf86d1e90041f02be423488ad6/app.base.json#L59

src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
boris-arkenaar: The response is the same in both cases.

boris-arkenaar:
This will need changes to the backend as well.

PR #2179, opened 2022-06-07
[SIG 4594] textareaInput provided when answer Overige in reports concerning clocks

SireeKoolenWijkstra:
When an user answers with Overige in a report concerning clocks a textareaInput is provided so that the user can elaborate if she wants. The answer is shown in the backoffice as Toelichting.

Test has been updated.


kiruys (approved):
👍

PR #2197, opened 2022-06-21
[SIG 4576] Styling of object information and notifications has been changed

SireeKoolenWijkstra:
When multiple objects are shown in the sidebar next to the map, the distances between these objects are enlarged in order to see more easily there are multiple objects. This includes information is that is giving within the object.

Margins are now 12px, 4px and 3 times 12px. Padding has been added to ensure a visual distance to the divider of 2 times 12px.


kiruys (approved):
Code wise lijkt het me prima! Alleen nog even die font-weight eruithalen

PR #2228, opened 2022-07-21
Add GZIP compression in nginx for speed optimization

bartjkdp:
This adds GZIP compression support to NGINX to improve Lighthouse score.

### DIFF
--- a/default.conf
+++ b/default.conf
@@ -11,6 +11,14 @@ server {
     # This will enforce HTTP browsing into HTTPS and avoid ssl stripping attack
     add_header Strict-Transport-Security "max-age=31536000; includeSubdomains;";
 
+    # Add GZIP compression for speed optimization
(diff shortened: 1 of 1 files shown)


default.conf
```diff
@@ -11,6 +11,14 @@ server {
     # This will enforce HTTP browsing into HTTPS and avoid ssl stripping attack
     add_header Strict-Transport-Security "max-age=31536000; includeSubdomains;";
 
+    # Add GZIP compression for speed optimization
+    gzip on;
```
OscarBakker, line 15: what about brotli?

bartjkdp: Could work, but this requires custom compilation of NGINX. Think this is a good first step.

kiruys (approved):

boris-arkenaar (approved):

PR #2229, opened 2022-07-24
Add support for EPSG:4326 & EPSG:3857 projection

bartjkdp:
This adds support for EPSG:4326 & EPSG:3857 map projections. The map configuration defaults to EPSG:28992 to make the change non-breaking.

Currently municipalities can only configure EPSG:28992 base maps. Some municipalities are not happy with the default base maps provided by PDOK:  e.g. house numbers are lacking and coloring of the base map could be more intuïtive.

Therefore they would like to provide their own base map. By using [tileserver-gl](https://github.com/maptiler/tileserver-gl) they can easily provide their own base map, but this requires support for EPSG:4326 projection.

This change can be tested with the following app.json:

```json
{
  "map": {
    "options": {
      "crs": "EPSG:3857"
    },
    "tiles": {
      "args": [
        "https://service.pdok.nl/brt/achtergrondkaart/wmts/v2_0/standaard/EPSG:3857/{z}/{x}/{y}.png"
      ],
      "options": {
        "maxZoom": 19,
        "maxNativeZoom": 19
      }
    }
  }
}
```

### DIFF
--- a/internals/schemas/app.schema.json
+++ b/internals/schemas/app.schema.json
@@ -399,6 +399,15 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "crs": {
+          "description": "The Coordinate Reference System (CRS) of the map",
+          "enum": [
+            "EPSG:28992",
+            "EPSG:3857",
+            "EPSG:4326"
+          ]
+
+        },
         "center": {
           "type": "array",
           "items": {
@@ -425,6 +434,7 @@
         }
       },
       "required": [
+        "crs",
         "center",
         "maxBounds",
         "maxZoom",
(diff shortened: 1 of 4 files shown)


boris-arkenaar (approved):

internals/schemas/app.schema.json
```diff
@@ -394,6 +394,9 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "crs": {
+          "type": "string"
+        },
```
kiruys, line 399: please add a description and the three types of strings that are possible

bartjkdp: Added this 👍

kiruys: thanks!

PR #2230, opened 2022-07-26
Make hardcoded "overig" categories Amsterdam-specific

bartjkdp:
Currently we have some hardcoded "overig" categories in the frontend for Amsterdam. This results in incorrect fallback behavior of the overig category within a main category for the other municipalities. It will always fall back to `overig` when the `MINIMUM_CERTAINTY` of a subcategory is not met, instead of the overig category of the main category (`overig-{main-slug}`).

This PR fixes this behavior and also still supports the Amsterdam-specific naming convention of overige categories, by adding feature flag called `enableAmsterdamSpecificOverigCategories`. When the feature flag is disabled, the consistent naming of overige categories that is used by the other municipalities is used.

### DIFF
--- a/app.amsterdam.json
+++ b/app.amsterdam.json
@@ -8,6 +8,7 @@
     "enableNearIncidents": true,
     "enablePublicSignalMap": false,
     "enableReporter": true,
+    "enableAmsterdamSpecificOverigCategories": true,
     "fetchDistrictsFromBackend": false,
     "fetchQuestionsFromBackend": false,
     "mapFilter24Hours": true,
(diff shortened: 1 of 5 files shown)


boris-arkenaar (approved):

bartjkdp:
@kiruys mentioned another solution that could work as well. Amsterdam could change the way that overige categories are named in a way that is consistent with the other municipalities. Then the hardcoded overige categories can be removed, which would simplify the code of the frontend.

I think this feature flag could help and temporarily be used when switching to the new way of naming.

boris-arkenaar (approved):

kiruys (approved):

PR #2231, opened 2022-07-30
[Sig-4621] show map with meldingen and generic pins

SireeKoolenWijkstra:
PR made because of holiday. Now somebody else will be able to continue if they want. The thing missing is a test to zoom in on address when that is selected in the detailpanel. A double check on the placement of the zoom icons would not be amiss. 

Meldingenkaart now shows a map, it can zoom and it shows, just like the OverviewMap all the issues of the last 24 hours.

### DIFF
--- a/src/signals/IncidentMap/components/IncidentMap.tsx
+++ b/src/signals/IncidentMap/components/IncidentMap.tsx
@@ -1,61 +1,247 @@
 // SPDX-License-Identifier: MPL-2.0
-// Copyright (C) 2022 Gemeente Amsterdam
+// Copyright (C) 2020 - 2021 Gemeente Amsterdam
+import type { Dispatch, FC, SetStateAction } from 'react'
+import { memo, useState, useCallback, useEffect, useMemo } from 'react'
+import { useSelector } from 'react-redux'
+import format from 'date-fns/format'
+import subDays from 'date-fns/addDays'
 
-import { StyledMap } from 'signals/incident/components/form/MapSelectors/Asset/Selector/styled'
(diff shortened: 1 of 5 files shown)


src/signals/IncidentMap/components/IncidentMap.tsx
```diff
@@ -1,8 +1,247 @@
...
+  box-sizing: border-box; // Override box-sizing: content-box set by Leaflet
+  z-index: 2; // position over the site header
+  display: flex;
+
+  @media screen and ${breakpoint('max-width', 'tabletM')} {
+    flex-direction: column;
+  }
+`
+
+export const StyledMap = styled(Map)`
```
OscarBakker, line 49: Map is een implementatie van van styled components, met incidentContainer state erin. Dit heb je niet nodig.

src/signals/IncidentMap/components/IncidentMap.tsx
```diff
@@ -1,8 +1,247 @@
...
+      layerInstance.addLayer(clusteredMarker)
+    })
+
+    return () => {
+      layerInstance.clearLayers()
+    }
+  })
+
+  return (
+    <AssetSelectProvider value={incidentMapContext}>
```
OscarBakker, line 222: Selecteren is niet nodig

src/signals/IncidentMap/components/IncidentMap.tsx
OscarBakker: op te lossen door closeMap() en mapactive door te geven als props of via context en optioneel te maken

PR #2235, opened 2022-08-17
[SIG-4697] changed alt text of logo in amsterdam.json file

vdegraaf:
Ticket: [SIG-4697](https://datapunt.atlassian.net/browse/SIG-4697)

## Context
The alt-text of the logo need to be 'gemeente' specific and should tell where it will redirect since it is also a link. 

## Changes
- changed the alt text in the app.amsterdam.json.

## Questions
- How do we make sure the other municipalities do the same?
- The SiteHeader renders a different header depending on the `isIncidentMap`. However, it also uses the Logo component differently. Resulting, for example, in double hrefs. Why is this?

### DIFF
--- a/app.amsterdam.json
+++ b/app.amsterdam.json
@@ -30,7 +30,7 @@
     "headerTitle": "",
     "helpText": "Wij zijn bereikbaar van maandag tot en met vrijdag van 8.00 tot 18.00 uur.",
     "helpTextHeader": "Lukt het niet om een melding te doen? Bel het telefoonnummer [14 020](tel:14020)",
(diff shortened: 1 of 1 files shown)


kiruys (approved):

kiruys:
Other municipalities should make the change themselves. We can change the logoDescription for Weesp though in this repo: https://github.com/Amsterdam/signalen

kiruys:
The siteheader for incident map uses a different logo to adjust to the new housestyle

kiruys (approved):

PR #2240, opened 2022-08-22
[SIG-4700] change breakpoint searchbar overlay

vdegraaf:
## Context
When a user tried to search for an address on 200% zoom on a desktop, the mobile overlay (fuill screen search) was triggered because it is depending on the amount of pixels, but this is unaccessible with mobile. 

## Changes
- Changed the breakpoint to MobileL so the mobile overlay will only be triggered with mobiles and not tablets.
- Set autofocus to search input when overlay appears

## Signalen


kiruys (approved):
👍

PR #2282, opened 2022-09-13
Make Docker image to be able to run as non-root

bartjkdp:
To improve security and to comply with security policies of environments with more strict security requirements (e.g. the environment of Utrecht) we require the frontend container to be able to run as non-root. This is also [considered best practice](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security).

This PR changes the container user from root to appuser.

⚠️  This change is breaking and requires to make infrastructural changes when deploying. The container port changes from 80 to 8080, as a non-root container is not able to bind to port 80.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,7 @@ RUN yarn add @exodus/schemasafe lodash
 
 COPY --from=base /app/build/. /usr/share/nginx/html/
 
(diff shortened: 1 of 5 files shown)


bartjkdp:
Closed in favor of #2287

PR #2288, opened 2022-09-19
[SIG-4792]: meldingenkaart refactor

vdegraaf:
Ticket: [SIG-4792](https://datapunt.atlassian.net/browse/SIG-4792)

## Context
The User Experience of the meldingenkaart was not as expected. For example, the max-width of the screen is set to 1400px, which is not desirable. Instead of choosing for conditional styling etc, I implemented the same approach as we use for the location map. Which is using `ReactDOM.createPortal` to render the map in the dom.

## Changes
- Change IncidentMapContainer to use createPortal().
- Move the conditional header is App to its own folder in /IncidentMap/components
- Add/adjust tests

## Signalen


src/signals/IncidentMap/components/Header/header.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const Header = () => {
+  const MenuItems = () => (
+    <Item>
+      <Button forwardedAs="a" href="/incident/beschrijf" target="_blank">
+        Doe een melding
+      </Button>
+    </Item>
+  )
+
+  const Navigation = () => (
```
OscarBakker, line 26: only question, why is this declared here instead of put into the main jsx?

src/signals/IncidentMap/components/Header/styled.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const HeaderWrapper = styled.div`
+  position: relative;
+  display: flex;
+  justify-content: space-between;
+  height: ${themeSpacing(HEADER_HEIGHT_DESKTOP)};
+  width: 100%;
+  padding: 0 ${themeSpacing(4)};
+  background-color: ${themeColor('tint', 'level1')};
+  box-shadow: 0 ${themeSpacing(1)} ${themeSpacing(1)} 0 rgba(0, 0, 0, 0.1);
+  z-index: 2;
```
OscarBakker, line 17: why is this z index 2? maybe a good idea for this repo to make a config file with named z indexes. That way you always have a reference to this somewhat magic number.

src/signals/IncidentMap/components/Header/header.tsx
vdegraaf: Refactored the component slightly to use media query like we use in the SiteHeader as well. The Hidden component was difficult to test.

src/signals/IncidentMap/components/Header/styled.tsx
vdegraaf: sharp. This was because the header and footer of the regular application were still rendered. I removed them when on /meldingenkaart and set the index of the map to be `0` and the header relative to that.

src/signals/IncidentMap/components/Header/Header.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+  window.innerHeight = y
+  window.dispatchEvent(new Event('resize'))
+}
+
+describe('Header', () => {
+  beforeEach(() => {
+    jest.spyOn(reactResponsive, 'useMediaQuery').mockReturnValue(false)
+  })
+
+  it('should render heading correctly', () => {
```
OscarBakker, line 21: you meant header?

OscarBakker (approved):
looks good to me! just one type I guess in the comments

PR #2292, opened 2022-09-21
[refactor]: ran auto re-order imports

vdegraaf:
Ticket: none

To get rid of all import warnings (due to the new ESLint configuration) I ran the auto re-order for imports in the following folders:
/containers
/hooks
/models

There are more PR's to come.


OscarBakker (approved):
lgtm

PR #2293, opened 2022-09-21
added istabul skip

vdegraaf:
Ticket: none

Test coverage of IncidentLayer was too low. It is also very hard to test this component and therefore added `/* istanbul ignore next */`. Similar to what was done in a previous implementation.


OscarBakker (approved):
If this ignore comment is used elsewhere then its a pass!

PR #2316, opened 2022-10-10
[SIG-4664]: add icons to filters and incidents

vdegraaf:
Ticket: [SIG-4664](https://datapunt.atlassian.net/browse/SIG-4664)

## Context
This PR adds custom icons to the Filters and incidents on the meldingenkaart. The icons are set in DjangoAdmin. 

## Changes
- Matches icons with filters
- Matches icons with Incidents
- adjust test

## Open Points
- On marker click the icon needs to change
- adjsut test get-filtered-incidents.test.ts


src/shared/services/configuration/map-markers.ts
```diff
@@ -45,3 +45,11 @@ export const currentIncidentIcon = L.icon({
   iconUrl: '/assets/images/area-map/icon-cross.svg',
   iconAnchor: [48, 48],
 })
+
+export const dynamicIcon = (iconUrl?: string) =>
+  L.icon({
+    iconUrl: iconUrl ? iconUrl : '/assets/images/icon-incident-marker.svg',
```
OscarBakker, line 51: could be set as standard param in line 49

src/signals/IncidentMap/components/FilterPanel/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const getSubCategories = (
+  subCategories: Category['sub_categories'] = []
+): SubCategory[] => {
+  const result = subCategories
+    .filter(({ is_public_accessible }) => is_public_accessible)
+    .map((subCategory) => {
+      const { name, _display, slug, _links } = subCategory
+      return {
+        _display: _display,
```
OscarBakker, line 38: consider

return {
    _display,
   filterActive: true,
   icon: _links['sia:icon']?.href,
   name,
   slug,
 }

or 

 return {
  ...subCategory
   filterActive: true,
   icon: _links['sia:icon']?.href,
 }

src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
```diff
@@ -12,9 +12,81 @@ export const getFilteredIncidents = (
...
+interface Icon {
+  slug: string
+  icon: string
+}
+
+const getListOfIcons = (filters: Filter[]) => {
+  const mainCategories: Icon[] = []
+  const subCategories: Icon[] = []
+
+  filters.map((filter) => {
```
OscarBakker, line 68: map should return an array, not be used like a forEach

src/signals/IncidentMap/components/IncidentMap/IncidentMap.tsx
```diff
@@ -54,30 +49,23 @@ export const IncidentMap = () => {
     },
     [setMapMessage, setShowMessage]
   )
+  /* istanbul ignore next */
+  const handleIncidentSelect = useCallback((incident) => {
+    setSelectedIncident(incident)
+    incident && setDrawerState(DrawerState.Open)
```
OscarBakker, line 55: huh, first setSelectedIncident and then check incident for not undefined?

src/signals/IncidentMap/components/FilterPanel/utils.ts
vdegraaf: Not sure I understand. The first suggestion is the same as the code, and the second spreads all properties of subCategory, which we do not want.

src/signals/IncidentMap/components/IncidentMap/IncidentMap.tsx
vdegraaf: Legacy of another try...

src/signals/IncidentMap/components/FilterPanel/utils.ts
OscarBakker: okay then the first suggestion, which is a shorter version

src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
OscarBakker: this is also a nice candidate for a .reduce btw

src/signals/IncidentMap/components/FilterPanel/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  categories: Category[]
+): Filter[] => {
+  const result = categories
+    .filter(({ is_public_accessible }) => is_public_accessible)
+    .map((category) => {
+      const { _display, _links, name, slug } = category
+      let sub_categories
+
+      if (sub_categories) {
+        sub_categories = getSubCategories(sub_categories)
```
OscarBakker, line 15: and this can be camelCase. Its snake case because of our python backend.

src/signals/IncidentMap/components/FilterPanel/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+
+      return {
+        _display: _display,
+        filterActive: true,
+        icon: _links['sia:icon']?.href,
+        name: name,
+        slug: slug,
+        subCategories: sub_categories,
```
OscarBakker, line 24: here you can use the short notation

PR #2333, opened 2022-10-18
[SIG-4810] Subcategorie met objecten tonen in het filterpanel

SireeKoolenWijkstra:
Context:
Filtering on subCategories in Filterpanel needs to be possible. It is not allowed to have selected a subCategory and not also its mainCategory. Icons of subCategory, if they are available, needs to be shown. Otherwise that of the mainCategory and if that's not available the standard marker needs to be shown. Sizes of the markers and icons will be addressed in a seperate story.
Only of the mainCategories Afval and Wegen... the subCategories are shown and only those filters will be shown that have incidents in the map at the current zoomlevel. If the center of the map moves, the filters unchecked need to stay unchecked. 

Changes:
- added a FilterCategories where it is decided whether there are subCategories and if the mainCategories are Afval and Wegen... of if only the mainCategory needs to be shown.
- added a FilterCategory which contains the checkbox, icon and name of category
- added functionality to exclude subcategories when maincategory is not checked
- added compute-nr-of-incidents in order to calculate if there are incidents that are also in the filter

Ticket: [SIG-4810](https://datapunt.atlassian.net/browse/SIG-4810)

## Signalen


src/signals/IncidentMap/components/FilterPanel/FilterPanel.tsx
```diff
@@ -56,19 +56,13 @@ export const FilterPanel = ({ filters, setFilters, setMapMessage }: Props) => {
     <>
       <Heading as="h4">Filter op onderwerp</Heading>
```
vdegraaf, line 57: The filter panel component is three levels deep now while the top level doesn't do much, and FilterCategory a lot. Maybe rearrange them as: 

```
<FilterPanel>
filters.map(

if with subCategories --> <FilterWithDropDown />
else --> <Filter />

)
</FilterPanel >
```

src/signals/IncidentMap/components/IncidentMap/IncidentMap.tsx
```diff
@@ -66,6 +66,19 @@ export const IncidentMap = () => {
...
+      categoryName.forEach((category) => {
+        if (category.filterActive != checked) {
+          updated = updateFilterCategory(category.name, updated)
+        }
+      })
+      setFilters(updated)
+    },
+    [filters]
+  )
+
```
vdegraaf, line 81: No need to move this function to IncidentMap. In the line of separation of concerns, this belongs to the FilterPanel component.

src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
```diff
@@ -12,9 +15,21 @@ export const getFilteredIncidents = (
...
-    activeFilterSlugs.includes(incident.properties.category.parent.slug)
-  )
+  const excludeSubSlugs = filters
+    .filter((filter) => filter.subCategories)
+    .map((filter) => {
+      return filter.subCategories
+        ?.filter((sub) => !sub.filterActive)
+        .map((sub) => sub.slug)
+    })
+    .flat(1)
```
vdegraaf, line 25: I don't think this should be here. In the getFilterCategories function the filters that are available are set. There we should determine which filter should be available for the user as well. In this util (fet-filtered-incidents) we should filter the incidents based on the filters we get. 

A simple solution could be:

Screenshot 2022-10-18 at 13 25 22

src/signals/IncidentMap/components/utils/update-filter-category.ts
```diff
@@ -3,14 +3,20 @@ import type { Filter } from '../../types'
...
         filterActive: !filter.filterActive,
       }
     }
+    if (filter.subCategories) {
+      return {
+        ...filter,
+        subCategories: updateFilterCategory(categoryName, filter.subCategories),
+      }
+    }
     return filter
```
vdegraaf, line 20: When all subCategories are unchecked, the parent should also be unchecked.

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return (
+    <>
+      {showAfvalAndWegenCategories(name) && subCategories ? (
+        <SectionWrapper key={`section-${name}`}>
+          <WrapperFilterCategoryWithIcon>
+            <FilterCategory
+              onToggleCategory={(checked: boolean) => {
+                onToggleCategory([filter].concat(subCategories), checked)
```
vdegraaf, line 36: You have four different implementations of "onToggleCategory" now. I think it is better to put all that logic into the onToggle function it self so it can handle all three different scenarios.

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onToggleCategory: (option: Filter[], select: boolean) => void
+  filter: Filter
+}
+
+export const FilterCategories = ({ onToggleCategory, filter }: Props) => {
+  const [showSubsection, setShowSubsection] = useState<boolean>(true)
+  const { name, filterActive, _display, icon, subCategories } = filter
+
+  const showAfvalAndWegenCategories = (name: string) => {
+    return name === 'Afval' || name === 'Wegen, verkeer, straatmeubilair'
```
vdegraaf, line 26: This is not necessary if you set the filters in the util.

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return (
+    <>
+      {showAfvalAndWegenCategories(name) && subCategories ? (
+        <SectionWrapper key={`section-${name}`}>
```
vdegraaf, line 32: This is an empty div.

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+          </WrapperFilterCategoryWithIcon>
+
+          <SubSection visible={showSubsection} lines={subCategories.length}>
+            {subCategories.map((filterSubCategories) => {
+              const {
+                name: nameChild,
+                filterActive: filterActiveChild,
+                _display: _displayChild,
+                icon: iconChild,
+              } = filterSubCategories
```
vdegraaf, line 59: Why do you rename the properties here?

src/signals/IncidentMap/components/FilterPanel/FilterCategory.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export const FilterCategory = ({
+  selected,
+  text,
+  onToggleCategory,
+  icon,
+}: Props) => {
+  return (
+    <LegendItem htmlFor={text}>
```
vdegraaf, line 21: This looks like bad copy pasting. Cause this is not the Legend.

src/signals/IncidentMap/components/FilterPanel/FilterCategory.tsx
```diff
@@ -0,0 +1,37 @@
...
+        data-testid={text}
+        id={text}
+        checked={selected}
+        onChange={() => {
+          onToggleCategory(!selected)
+        }}
+      />
+      <StyledImg
+        alt="icon"
+        src={icon || 'assets/images/icon-incident-marker.svg'}
```
vdegraaf, line 32: The default marker should be set on the incident itself, not here.

src/signals/IncidentMap/components/FilterPanel/styled.ts
```diff
@@ -50,3 +47,66 @@ export const StyledButton = styled(Button)`
...
+  color: ${(props) =>
+    props.noResults
+      ? themeColor('tint', 'level4')
+      : themeColor('tint', 'level7')};
+`
+
+export const LegendItem = styled.label`
+  display: flex;
+  align-items: center;
+`
```
vdegraaf, line 112: Not very clear naming. Looks like just copy pasting.

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  onToggleCategory={() => {
+                    onToggleCategory(
+                      filterActiveChild
+                        ? [filterSubCategories]
+                        : [filter, filterSubCategories],
+                      !filterActiveChild
+                    )
+                  }}
+                  selected={filterActiveChild}
+                  text={_displayChild || nameChild}
```
vdegraaf, line 71: why is the fallback nameChild?

src/signals/IncidentMap/components/FilterPanel/FilterPanel.tsx
SireeKoolenWijkstra: Done, but now the code is less generic. It will only work for subCategories and mainCategories, if the subCategories ever will have subCategories of their own we're out of luck.

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <SectionWrapper key={`section-${name}`}>
+          <WrapperFilterCategoryWithIcon>
+            <FilterCategory
+              onToggleCategory={(checked: boolean) => {
+                onToggleCategory([filter].concat(subCategories), checked)
+              }}
+              selected={filterActive}
+              text={name}
+              icon={icon}
+            ></FilterCategory>
```
OscarBakker, line 41: this has an empty body, so would use <FilterCategory/>

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  InvisibleButton,
+  SubSection,
+} from './styled'
+
+export interface Props {
+  onToggleCategory: (option: Filter[], select: boolean) => void
+  filter: Filter
+}
+
+export const FilterCategories = ({ onToggleCategory, filter }: Props) => {
```
OscarBakker, line 21: should this be a default export?

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { FilterCategory } from './FilterCategory'
+import {
+  SectionWrapper,
+  WrapperFilterCategoryWithIcon,
+  InvisibleButton,
+  SubSection,
+} from './styled'
+
+export interface Props {
+  onToggleCategory: (option: Filter[], select: boolean) => void
```
OscarBakker, line 17: this could just be the same type as the components that passes onToggleCategory

src/signals/IncidentMap/components/IncidentMap/IncidentMap.tsx
```diff
@@ -66,6 +66,19 @@ export const IncidentMap = () => {
     resetMarkerIcons()
   }, [resetMarkerIcons])
 
+  const toggleFilter = useCallback(
+    (categoryName: Filter[], checked: boolean) => {
```
OscarBakker, line 70: by just having the name you can toggle the checked..so checked should be passed down the tree of components. Then do a find on it and return the result.

src/signals/IncidentMap/components/IncidentMap/IncidentMap.tsx
OscarBakker: true

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
SireeKoolenWijkstra: So that it is obvious that it's the properties of a subCategory and not a mainCategory. I like obvious, makes it easier to read.

src/signals/IncidentMap/components/FilterPanel/FilterCategory.tsx
SireeKoolenWijkstra: changed the name

src/signals/IncidentMap/components/FilterPanel/FilterCategories.tsx
SireeKoolenWijkstra: Done

src/signals/IncidentMap/components/FilterPanel/styled.ts
```diff
@@ -50,3 +47,66 @@ export const StyledButton = styled(Button)`
...
+
+export const CategoryItemText = styled.span<LegendTextProps>`
+  display: flex;
+  align-items: center;
+  color: ${(props) =>
+    props.noResults
+      ? themeColor('tint', 'level4')
+      : themeColor('tint', 'level7')};
+`
+
```
vdegraaf, line 108: This is interesting but not used:
- the noResult prop is there to let the styling know if the filter has any incidents. If not, you can pass this prop and render the text in grey. Maybe we can use this too? (see screenshot)
- "LegendTextProps" is a name from somewhere else. 


Screenshot 2022-11-01 at 11 07 22

src/signals/IncidentMap/components/utils/compute-nr-of-incidents-per-filter.ts
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2022 Gemeente Amsterdam
+import type { Feature } from 'geojson'
+
+import type { Filter, Properties } from '../../types'
+
+export const computeNrOfIncidentsPerFilter = (
```
vdegraaf: Name is not very clear for me. Maybe something like: `getFiltersWithNrOfIncidents` ?

src/signals/IncidentMap/components/IncidentMap/IncidentMap.tsx
```diff
@@ -116,10 +116,12 @@ export const IncidentMap = () => {
...
   useEffect(() => {
-    if (data?.features) {
-      const filteredIncidents = getFilteredIncidents(filters, data.features)
-      setFilteredIncidents(filteredIncidents)
-    }
+    const incidents = data?.features || []
+
+    const filteredIncidents = getFilteredIncidents(filters, incidents)
+    setFilteredIncidents(filteredIncidents)
+    const filtersWithCounts = computeNrOfIncidentsPerFilter(filters, incidents)
```
vdegraaf: Not a big fan of the name, maybe filtersWithNrOfIncidents ?

vdegraaf (commented):
Looking good! Did become a big PR though, so left some comments ;)

Also:
- Where is the separate ticket for the icons?
- Maybe add default icons when the connection to Django Admin is lost (currently the case, and the user sees these original fall back images which are ugly)
- Because we don't show any filters if there are no incidents in that category, the filterPanel changes all the time when going over the map. Which does not give a very nice experience. Maybe we should change it back or consult Linda for a solution. (Or maybe leave the main caterogies always visible and subCategories only subject to change)

src/signals/IncidentMap/components/utils/compute-nr-of-incidents-per-filter.ts
OscarBakker: getFiltersWithincidentCounts? we gebruiken nu count op een paar plekken

src/signals/IncidentMap/components/__test__/mock-incidents.ts
```diff
@@ -1,6 +1,90 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2022 Gemeente Amsterdam
+import type { Feature } from 'geojson'
+
 import type { Incident } from '../../types'
+import type { Properties } from '../../types'
+
+export const mockIncidentsShort: Feature<any, Properties>[] = [
```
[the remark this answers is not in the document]
OscarBakker: @SireeKoolenWijkstra  yeah please tell:) these mocks not that different

src/signals/IncidentMap/components/FilterPanel/FilterCategory.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  selected: true,
+  text: 'mockName',
+  onToggleCategory: mockOnToggleCategory,
+  icon: 'https://ae70d54aca324d0480ca01934240c78f.objectstore.eu/signals/icons/categories/afval/afval.svg?temp_url_sig=6679c552c423eb18ffe55643e5692fb4c348bde4e2bde851f33a7aef8d0474fe&temp_url_expires=1665401494',
+}
+
+describe('FilterCategory', () => {
+  it('toggles the checkbox on click', async () => {
+    render(<FilterCategory {...defaultProps} />)
+    userEvent.click(screen.getByRole('checkbox'))
```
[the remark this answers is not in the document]
OscarBakker: added an expect screen.getByRole('checkbox', {name: /mockName/}) to be checked

OscarBakker (approved):

PR #2350, opened 2022-10-25
[SIG-4784]: zoom to incident on click

vdegraaf:
Ticket: [SIG-4784](https://datapunt.atlassian.net/browse/SIG-4784)

## Context
This PR upgrades the user experience mainly while on mobile.
- When the user selects a marker that is underneath the drawerOverlay we want to move the map slightly up so it is in view. 
- We want to close the drawerOverlay when GPS location is used on mobile.
- When using address input, there should be an overlay
- In general, there should be no weird behaviour while on mobile

## Changes
- Add fly to on IncidentSelect
- pointToLayer: filter so that selected icon stays in place
- add small points to latitude to move the map up when on mobile
- When on clear it removes the pin from the map
- Added a AddressSearchMobile, which is an overlay that pops up when the user uses the address input on mobile
- Changes some css

## Signalen

### DIFF
--- a/src/signals/IncidentMap/components/AddressLocation/AddressSearchMobile.tsx
+++ b/src/signals/IncidentMap/components/AddressLocation/AddressSearchMobile.tsx
@@ -38,11 +38,11 @@ export const AddressSearchMobile = ({
 
   return (
     <Wrapper>
-      <AddressSearch id="AddressSearchMobile">
+      <AddressSearch id="addressSearchMobile">
         <header>
           <Button
             aria-label="Terug"
-            aria-controls="AddressSearchMobile"
+            aria-controls="addressSearchMobile"
             icon={<ChevronLeft />}
             iconSize={16}
             onClick={() => setShowAddressSearchMobile(false)}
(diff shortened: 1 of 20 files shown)


src/signals/IncidentMap/components/AddressLocation/AddressSearchMobile.tsx
```diff
@@ -0,0 +1,70 @@
...
+    (option: PdokResponse) => {
+      setCoordinates(option.data.location)
+      setShowAddressSearchMobile(false)
+    },
+    [setCoordinates, setShowAddressSearchMobile]
+  )
+
+  return (
+    <Wrapper>
+      <AddressSearch id="AddressSearchMobile">
```
OscarBakker, line 41: can this be lowercased?

src/signals/IncidentMap/components/AddressLocation/AddressSearchMobile.tsx
```diff
@@ -0,0 +1,70 @@
...
+            variant="blank"
+          />
+          <StyledPDOKAutoSuggest
+            onClear={() => setCoordinates(undefined)}
+            onData={setOptionsList}
+            onSelect={onAddressSelect}
+            showInlineList={false}
+            value={address}
+            placeholder="Zoom naar adres"
+            autoFocus={true}
```
OscarBakker, line 60: you can just add autoFocus, a shortcut for autoFocus={true}

OscarBakker (approved):

PR #2355, opened 2022-11-02
Change the apiBaseUrl to the new one

vdegraaf:
Ticket: none

We need to talk to a different api endpoint since the migration to Azure.


SireeKoolenWijkstra (approved):
no comment :-)

PR #2364, opened 2022-11-09
[bugfix-4867] Resize filter icons and set default icons

vdegraaf:
Ticket: [SIG-4867](https://datapunt.atlassian.net/browse/SIG-4867)

- The default icon in the filterpanel of the incidentMap was too big. Resizing it to follow the other icons. 
- Meanwhile, set default icon where filters are created.

### DIFF
--- a/src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
+++ b/src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
@@ -51,7 +51,7 @@ export const getFilteredIncidents = (
 
 interface Icon {
   slug: string
-  icon: string
+  icon?: string
 }
 
 const getListOfIcons = (filters: Filter[]) => {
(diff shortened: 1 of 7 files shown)


src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
```diff
@@ -51,7 +51,7 @@ export const getFilteredIncidents = (
 
 interface Icon {
   slug: string
-  icon: string
+  icon?: string
```
OscarBakker, line 54: ow die zat er nog niet op?!

OscarBakker (approved):

src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
vdegraaf: Er zat Partial op de hele type, die alles in the type optioneel maakt...

PR #2365, opened 2022-11-09
[bugfix]: fix display name on filter panel meldingenkaart

vdegraaf:
Ticket: none

The wrong name was displayed in the filter panel.

Screenshot 2022-11-09 at 12 27 48


OscarBakker (approved):
nice

PR #2372, opened 2022-11-14
[bugfix] Correct login page content

vdegraaf:
Ticket: none

The content of the login page was not according to design but accidentally swapped with the expiredLink page.


SireeKoolenWijkstra (approved):
lgtm

PR #2384, opened 2022-11-17
By default set reporterMailHandledNegativeContactEnabled to false

bartjkdp:
To make upgrading the Signalen frontend non-breaking, `reporterMailHandledNegativeContactEnabled` should by default set to false. This results in the original behavior of Signalen.

Municipalities can opt-in to this new functionality by enabling the feature flag.

### DIFF
--- a/app.base.json
+++ b/app.base.json
@@ -16,7 +16,7 @@
     "fetchDistrictsFromBackend": false,
     "fetchQuestionsFromBackend": false,
     "mapFilter24Hours": false,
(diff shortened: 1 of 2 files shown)


OscarBakker:
- [ ] This test needs to be updated 'FAIL src/signals/incident/containers/KtoContainer/index.test.js
[985]'

bartjkdp:
Thanks @OscarBakker. Just updated the test. Think this should be ok, right?

OscarBakker (approved):
lgtm

PR #2393, opened 2022-11-23
[feature 4891] set font to Amsterdam Sans

SireeKoolenWijkstra:
**Context**
Per 15th of December the old font Avenir Next is no longer paid for. To prevent fines font in Sia needs het font to change to Amsterdam Sans.

**Tested by comparing local to acceptation environment** 
- happy flow voor civillians
- happy flow, filters and Signaleringen in back office
- kto-proces, not happy with result
- mijn-meldingen
- meldingenkaart

Ticket: [SIG-4891](https://datapunt.atlassian.net/browse/SIG-4891)

## Signalen

- [ ] Double-check your branch is based on `main` and targets `main`
- [ ] Pull request has tests (we are going for 100% coverage!)
- [ ] Code is well-commented, linted and follows project conventions
- [ ] Committed source code is headed by the correct SPDX license expression

### DIFF
--- a/src/global.css
+++ b/src/global.css
@@ -50,7 +50,7 @@ html {
 
 body {
   font-family: Amsterdam Sans, sans-serif;
-  font-weight: 400 ;
+  font-weight: 400;
   font-size: 1rem;
   line-height: 1.375;
   overflow-y: scroll;
--- a/src/global.css
+++ b/src/global.css
@@ -49,7 +49,7 @@ html {
 }
 
 body {
-  font: 400 1rem/1.375 Avenir Next;
+  font: 400 1rem/1.375 "Amsterdam Sans";
   overflow-y: scroll;
   background-color: #e5e5e5;
 }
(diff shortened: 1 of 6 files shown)


bartjkdp:
Maybe also provide a fallback for other municipalities? Not sure if everyone can use "Amsterdam Sans".

OscarBakker:
> 

agree

SireeKoolenWijkstra:
Inheritance is introduced by removing font setting to Amsterdam Sans in components. Only in Emailpreview was this not possible and is it kept as is.  

Fallback is introduced (set to Sans-serif). According to Jelle almost all municipalities will have their own font, so it had no importance according to the VNG.

src/global.css
```diff
@@ -49,7 +49,10 @@ html {
 }
 
 body {
-  font: 400 1rem/1.375 Avenir Next;
+  font-family: Amsterdam Sans, sans-serif;
```
OscarBakker, line 52: check spaces

OscarBakker (approved):

PR #2394, opened 2022-11-24
[FE] Add feedback incidents.

OscarBakker:
## Changes

Make my incidents list and history more efficient by moving fetch logic to detail component. Also remove detail stuff from context provider.


src/signals/my-incidents/pages/Detail.tsx
```diff
@@ -12,25 +12,23 @@ import { History } from '../components/History/History'
...
   const location = useLocationReferrer() as Location
-  const { incidentsDetail, setIncidentsDetail } = useMyIncidents()
   const incidentDisplay = useRef<string>()
   const locationPathArray = location.pathname.split('/')
   const token = locationPathArray[locationPathArray.length - 2]
   const uuid = locationPathArray[locationPathArray.length - 1]
 
   useEffect(() => {
-    data && setIncidentsDetail(data)
     incidentDisplay.current = data?._display
```
vdegraaf, line 30: Why is a ref needed here?

src/signals/my-incidents/pages/Detail.tsx
```diff
@@ -56,19 +67,16 @@ export const Detail = () => {
         >
           <title>{`Meldingsnummer: ${incidentDisplay.current}`}</title>
         </Helmet>
-        {showMap && incidentsDetail?.location ? (
-          <Map
-            close={() => setShowMap(false)}
-            location={incidentsDetail.location}
-          />
+        {showMap && data?.location ? (
```
vdegraaf, line 70: Location is not optional. So changing the types makes this obsolete. Also, one level down (in IncidentDetailComponent there is an early return that checks if data exists. Might be better to add the early return in this component since you do not want to return anything if you don't have the info.

src/signals/my-incidents/types.ts
```diff
@@ -6,9 +6,7 @@ export interface MyIncidentsValue {
   email?: string
   setEmail: (email: string) => void
   incidentsList?: MyIncident[]
-  incidentsDetail?: MyIncident
   setIncidentsList: (incidentsList: MyIncident[]) => void
```
vdegraaf, line 9: The types here are a bit misleading since `Location` and `sia:attachments` are not optional. Which also leads to extra code. I would suggest creating a separate type for MyIncidentDetail which extends the base one. Example: 

Screenshot 2022-11-28 at 12 48 56

vdegraaf (commented):
Looking good! 
Just a few small comments.

src/signals/my-incidents/pages/Detail.tsx
OscarBakker: yep!

src/signals/my-incidents/types.ts
OscarBakker: done

vdegraaf (approved):

PR #2400, opened 2022-11-29
Improve review apps

bartjkdp:
- Update default configuration of review apps
- Deploy review apps when PR is created and then labeled
- Update to Helm chart version 4.5.0


OscarBakker (approved):
👍

PR #2401, opened 2022-11-30
update package-lock.json

vdegraaf:
package-lock.json needed to be updates since we got an error when pushing a release to the acceptance environment.

error: https://github.com/Amsterdam/signals-frontend/actions/runs/3583938417/jobs/6029976808
solution: https://sebhastian.com/npm-err-cb-never-called/#:~:text=Check%20your%20npm%20version%20and,error%20by%20clearing%20the%20cache.&text=Once%20the%20cache%20is%20cleared,using%20the%20latest%20npm%20version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4364,9 +4364,9 @@
       }
     },
     "node_modules/@xmldom/xmldom": {
-      "version": "0.7.6",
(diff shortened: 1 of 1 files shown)


vdegraaf:
Not needed anymore.

PR #2441, opened 2023-01-03
[fix-4934]: fix incidentMap subcategory filter

vdegraaf:
Ticket: [SIG-4934](https://datapunt.atlassian.net/browse/SIG-4934)

Filtering on subcategory did not work as expected. When selecting/deselecting a subcategory, all incidents of the main category where shown. This was because `activeIncidents` where defined as having an active filter slug on main PR subcategory. While, when a mainCategoryFilter has the option to filter in subcategories as well, an incident can have a deslected subcategory while the main category stays selected. This incident should be treated as deselected, not as selected. 

## Changes
- changes the function that checks whether an incident is active
- adjusted the test. Which had a wrong assertion, therefore passed while faulty.

### DIFF
--- a/src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
+++ b/src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright (C) 2022 Gemeente Amsterdam
-
 import type { Filter, Incident } from '../../types'
+import { showSubCategoryFilter } from '../FilterPanel/utils'
 /*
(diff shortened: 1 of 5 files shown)


src/signals/IncidentMap/components/utils/get-filtered-incidents.ts
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright (C) 2022 Gemeente Amsterdam
```
OscarBakker, line 2: 2023

OscarBakker (approved):
its 2023!

vdegraaf:
> its 2023!

haha check! Updated them all.

PR #2446, opened 2023-01-04
[fix]: Hide 'Extern' button when incident is closed

MauritsR:
### Context

When an incident has been closed (_Afgehandeld_ or _Gennuleerd_), it should not be possible to forward it to an external party. In this case, the "Extern" button should simply not be rendered.

### Note

Besides the issue described above, this PR provides a minor styling fix in the forward to external form: images in portrait orientation were stretched vertically.

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
+++ b/src/signals/incident-management/containers/IncidentDetail/components/DetailHeader/index.js
@@ -9,7 +9,10 @@ import styled from 'styled-components'
 import BackLink from 'components/BackLink'
 import Button from 'components/Button'
 import configuration from 'shared/services/configuration/configuration'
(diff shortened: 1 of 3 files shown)


vdegraaf (approved):
LGTM
note that the globalthreshold of the tests fail. This is because of an issue in main. We can rebase once the issue is fixed there.

UPDATE: @MauritsR , the threshold of main is sufficient again, so please rebase than we can proceed.

vdegraaf:
Merged this PR bypassing the protections since it the blocking issue is the global coverage threshold. Which was not lowered by this PR, but the glitch was already in main when this branch was created. Couldn't rebase since it was a fork of Maurits, this seemed to be the easiest way so we could continue with an acceptance release.

PR #2451, opened 2023-01-05
[Refactor-4656]: nextButton-to-ts

SireeKoolenWijkstra:
**Context**
changed NextButton/index.js to NextButton.tsx. Made explicit parameter of onCLick and added index.ts

Ticket: [SIG-4656](https://datapunt.atlassian.net/browse/SIG-4656)
Ticket: [SIG-4819](https://datapunt.atlassian.net/browse/SIG-4819)

## Signalen

[SIG-4656]: https://datapunt.atlassian.net/browse/SIG-4656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SIG-4819]: https://datapunt.atlassian.net/browse/SIG-4819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/NextButton/NextButton.tsx
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
-// Copyright (C) 2020 - 2022 Gemeente Amsterdam
+// Copyright (C) 2020 - 2023 Gemeente Amsterdam
+import type { BaseSyntheticEvent, ReactNode } from 'react'
```
vdegraaf, line 3: What is the reason you chose `BaseSyntheticEvent` and not just `React.MouseEvent` or something?

src/components/NextButton/NextButton.tsx
```diff
@@ -9,7 +11,13 @@ const StyledButton = styled(Button)`
...
+interface NextButtonProps {
+  className?: string
+  children: ReactNode
+  onClick: (event: BaseSyntheticEvent) => void
+}
+
+const NextButton = ({ className, children, onClick }: NextButtonProps) => (
   <StyledButton
     className={className}
     data-testid="next-button"
```
vdegraaf, line 23: Using typescript makes proptypes obsolete. So you can remove the proptypes. But, there is also a defaultProp, you should implement that in your change as well. 
Also, I should just name it Props, not NextButtonProps. It's only used in this component's context, adding the name is a bit double.

src/signals/incident/components/IncidentNavigation/IncidentNavigation.tsx
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
-// Copyright (C) 2018 - 2022 Gemeente Amsterdam
+// Copyright (C) 2018 - 2023 Gemeente Amsterdam
```
vdegraaf, line 2: Did you change anything here?

src/components/NextButton/NextButton.tsx
```diff
@@ -9,7 +11,13 @@ const StyledButton = styled(Button)`
   margin-right: 15px !important;
 `
 
-const NextButton = ({ className, children, onClick }) => (
+interface NextButtonProps {
+  className?: string
```
vdegraaf, line 15: Don't think className was optional in the previous setup right?

OscarBakker: there is this implementation in IncidentNavigation, seems it should be optional 

```<NextButton
            onClick={(e) => {
              handleSubmit(e, next, wizardStep.formAction)
            }}
            data-testid="next-button"

```

src/components/NextButton/NextButton.tsx
SireeKoolenWijkstra: I've removed the .prototypes, but according to the internet there is a difference between them. https://stackoverflow.com/questions/41746028/proptypes-in-a-typescript-react-application.

Done the other two.

src/signals/incident/components/IncidentNavigation/IncidentNavigation.tsx
SireeKoolenWijkstra: I had, but then overwritten it again except, it seems, for the date change.

src/components/NextButton/NextButton.tsx
SireeKoolenWijkstra: Left it optional.

src/components/NextButton/NextButton.tsx
vdegraaf: That is the type of onClick in this IncidentNavigationProps, but the NextButton component is a generic component that can be used anywhere, so the onClick should be generic as well. Not sure what BaseSyntheticEvent adds to the type though. 
@OscarBakker can you explain why you used this type (since it it not used anywhere else) and why a simple MouseEvent does not suffice?

src/components/NextButton/NextButton.tsx
vdegraaf: They are a different thing indeed, but in this ADR 0007-typescript.md we agreed to remove them in favor of typescript.

vdegraaf (approved):

PR #2460, opened 2023-01-16
[feat-4951]: add subfilters to configuration

vdegraaf:
Ticket: [SIG-4951](https://datapunt.atlassian.net/browse/SIG-4951)

## Context
The values of the maincategories of which the subcatefory filters were enabled was hard coded. Since other municipalities want to have different filters we want to make them configurable.

## Changes
- Added hasSubcategoriesEnabled to the map options in the configuration file
- Adjust tests

## Release notes
- we should add this configuration in the base file in azure config of acceptance and production
- We should set the correct config for Amsterdam in azure config

[SIG-4951]: https://datapunt.atlassian.net/browse/SIG-4951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/signals/IncidentMap/components/FilterPanel/utils.test.ts
+++ b/src/signals/IncidentMap/components/FilterPanel/utils.test.ts
@@ -41,6 +41,7 @@ describe('getFilterCategoriesWithIcon', () => {
 
 describe('showSubCategoryFilter', () => {
   it('should return false on default', () => {
+    configuration.map.optionsIncidentMap.hasSubfiltersEnabled = []
     const result = showSubCategoryFilter('afval')
 
     expect(result).toEqual(false)
--- a/src/signals/IncidentMap/components/FilterPanel/utils.test.ts
+++ b/src/signals/IncidentMap/components/FilterPanel/utils.test.ts
@@ -1,7 +1,11 @@
+import configuration from 'shared/services/configuration/configuration'
 import type Categories from 'types/api/categories'
 
 import { fetchCategoriesResponse } from '../__test__'
 import { getFilterCategoriesWithIcons } from './utils'
+import { showSubCategoryFilter } from './utils'
+
+jest.mock('shared/services/configuration/configuration')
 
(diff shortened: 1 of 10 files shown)


SireeKoolenWijkstra (changes requested):
When overig is selected, then that category will have a chevrondown etc, but you can't uncheck the main category nor the subcategory of overig. Probably because it doesn't have subcategories of its own. It works like a charm with the other categories

src/signals/IncidentMap/components/FilterPanel/utils.test.ts
```diff
@@ -34,3 +38,19 @@ describe('getFilterCategoriesWithIcon', () => {
     )
   })
 })
+
+describe('showSubCategoryFilter', () => {
+  it('should return false on default', () => {
```
OscarBakker, line 43: might wanne add a configuration mock reset here, when adding a new test, you don't want configuration.map.optionsIncidentMap.hasSubfiltersEnabled be set is some other it block.

vdegraaf:
> When overig is selected, then that category will have a chevrondown etc, but you can't uncheck the main category nor the subcategory of overig. Probably because it doesn't have subcategories of its own. It works like a charm with the other categories

It doesn't work because the name of the main and subcategory are the same which causes the update function to break. I changed it to use slugs because they should be identical. However, the slugs of "overig" are also the same, making it impossible to set subFilters to the "Overig" category. Which might not be an issue since it only has one public accessible sub-filter. I added this to the description in the config file.

src/signals/IncidentMap/components/FilterPanel/utils.test.ts
vdegraaf: good one!

SireeKoolenWijkstra (approved):

PR #2462, opened 2023-01-18
[chore]: reset app.base.json optionsIncidentMap config

vdegraaf:
Ticket: none

I added some main categories to the hasSubfiltersEnabled in app.base.json. But this file is the base config for all municipalities. The default should be that no mainFilters have subfilters, so adjusted the file.


OscarBakker (approved):
okay

PR #2471, opened 2023-01-25
[fix-4981]: removed Dashboardlink from login page

SireeKoolenWijkstra:
In the backoffice, on the page where one can enter credentials, the menu item dashboard was visible. This should not be visible when a user is not logged in. Now it is only visisble if user is authenticated. 
Test need not be added, because there is already a test for rendering when not authenticated. 

Ticket: [SIG-4981](https://gemeente-amsterdam.atlassian.net/browse/SIG-4981)

## Signalen

### DIFF
--- a/app.amsterdam.json
+++ b/app.amsterdam.json
@@ -17,7 +17,8 @@
     "reporterMailHandledNegativeContactEnabled": true,
     "showThorButton": true,
     "showVulaanControls": true,
-    "useProjectenSignalType": false
+    "useProjectenSignalType": false,
+    "showDashboard": true
   },
   "head": {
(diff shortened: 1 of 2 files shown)


OscarBakker (commented):
might wanne change this commit message to something like:
 
Add isAuthenticated check for showing Dashboard

In the backoffice, where one can enter credentials, the menu item dashboard was visible. This should not be visible when a user is not logged in. Now it is only visisble if user is authenticated. Test need not be added, because there is already a test for renderingen when not authenticated.

SireeKoolenWijkstra:
@OscarBakker Like, did it.

OscarBakker (approved):

PR #2475, opened 2023-01-26
Do not set max-width on logo, as this breaks aspect ratio's for other municipalities

bartjkdp:
Different municipalities have different aspect ratio's of the logo. The fixed max-width breaks aspect ratio's of different logos. Would removing the max-width work for you?

Would be happy to implement other solution directions as well to support different aspect ratio's.

### DIFF
--- a/src/signals/IncidentMap/components/Header/styled.tsx
+++ b/src/signals/IncidentMap/components/Header/styled.tsx
@@ -30,6 +30,5 @@ export const Title = styled.div`
 
   img {
     height: 100%;
-    max-width: 90px;
   }
 `


vdegraaf (changes requested):
This change will break the Amsterdam Logo. We need to think of a different strategy. Should I create a bug ticket on our side to fix this or will you pick it up @bartjkdp ?

Screenshot 2023-01-27 at 14 06 53

bartjkdp:
@vdegraaf you are right. I had this problems with other logo's as well. Would be great if you can create a bug ticket for this. I will close this PR.

PR #2476, opened 2023-01-27
Make text field required when transitioning from heropenen to afgehandeld as this is enforced in the backend as well

bartjkdp:
The backend already enforces this, so this fixes this in the frontend as well. This gives the user a more friendly message.

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/utils.ts
+++ b/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/utils.ts
@@ -49,13 +49,20 @@ export const textIsRequired = ({
       StatusCode.Heropend,
       StatusCode.ReactieGevraagd,
     ].includes(toStatus)
-  } else {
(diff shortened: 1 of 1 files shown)


vdegraaf (approved):
@bartjkdp didn't functionally tested it, but doesn't seem like a major change. Did you test this locally or will you do that on acceptance?

bartjkdp:
@vdegraaf did test this locally, and it works. Just wanted to make sure you are happy with this solution direction as well. Can add automated test as this is currently not covered by the test suite.

vdegraaf:
> @vdegraaf did test this locally, and it works. Just wanted to make sure you are happy with this solution direction as well. Can add an automated test as this is currently not covered by the test suite.

@bartjkdp would be very nice to have it included in the test!

bartjkdp:
This PR is superseded by https://github.com/Amsterdam/signals-frontend/pull/2488

PR #2488, opened 2023-02-06
[fix-5005]:re-add send email notification

vdegraaf:
Ticket: [SIG-5005](https://gemeente-amsterdam.atlassian.net/browse/SIG-5005)

When an incident had a "verzoek tot heropenen" request and in the backoffice someone set the status to "Afgehandeld", it a says "no email was send". However, an email is send. This PR removes the logic that causes this behaviour. See screenshot for result.

Screenshot 2023-02-06 at 14 29 42

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/StatusForm.test.tsx
+++ b/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/StatusForm.test.tsx
@@ -24,10 +24,7 @@ import { withAppContext } from 'test/utils'
 import type { Incident } from 'types/api/incident'
 import incidentJSON from 'utils/__tests__/fixtures/incident.json'
 
-import StatusForm from '..'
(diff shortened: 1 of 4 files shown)


OscarBakker (approved):

bartjkdp:
This enforces the correct behavior when the feature flag `REPORTER_MAIL_HANDLED_NEGATIVE_CONTACT_ENABLED` (backend) and `reporterMailHandledNegativeContactEnabled` (frontend) is true, but this change breaks Signalen when this feature flag is false. 

When the feature flag is false, the reporter now is not able to transition from "verzoek tot afhandelen" tot "afgehandeld" any more. In reality, no e-mail is sent when the feature flag is false.

image

Do we need to keep this feature flag in Signalen?

bartjkdp:
This PR enforces the correct behavior when the feature flag `reporterMailHandledNegativeContactEnabled` is enabled (also on the backend), but there is still a small bug when the feature flag is disabled. I will create a separate PR for that.

PR #2504, opened 2023-02-21
[feat-4900]: fetch Bar Chart data for Dashboard

vdegraaf:
Ticket: [SIG-4900](https://gemeente-amsterdam.atlassian.net/browse/SIG-4900)

## Context
Part two of the story. It fetches the data for the bar chart. It does multiple calls. For each status a new call with the status and queryString (the parameters of the filter bar) included. 

## Changes
- Fetch data from backend
- Make query dependent on changing queryString
- Fix styling barChart
- Add tests


src/signals/incident-management/containers/Dashboard/components/BarChart/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  return constants.statusList.map((status) => {
+    const params = queryString
+      ? `${queryString}&status=${status.query}`
+      : `status=${status.query}`
+    return `${endpoint}stats/total?${params}`
+  })
+}
+
+export const formatData = (rawData: RawData[]): BarChartValue[] =>
+  rawData.map((value, i) => {
```
OscarBakker, line 20: return can we wirtten like this
```
   rawData.map((value, i) => ({
```

src/signals/incident-management/containers/Dashboard/components/BarChart/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface RawData {
+  total: number
+  results: []
```
OscarBakker, line 3: Arent they any array?

src/signals/incident-management/containers/Dashboard/components/BarChart/BarChart.tsx
```diff
@@ -1,21 +1,85 @@
...
+  } = useFetchAll<RawData>()
+
+  useEffect(() => {
+    getBarChart(queryList)
+  }, [getBarChart, queryList])
+
+  useEffect(() => {
+    if (rawData) {
+      const formattedData = formatData(rawData)
+      const total = getTotalNrOfIncidents(rawData)
```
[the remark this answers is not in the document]
vdegraaf: Not overwriting it, but see what you mean. Changed the name.

src/signals/incident-management/containers/Dashboard/components/BarChart/types.ts
vdegraaf: hmm, true. I made it `undefined[]`, which is technically not true since there won't be any undefined item in the array, but is as close as it gets. Couldn't find a typescript type that declares an array that is always empty.

OscarBakker (approved):

PR #2511, opened 2023-02-27
[chore]: Remove strike Alert

vdegraaf:
Since the strike Alert was temporary I delete the whole functionality in this PR. 

### Release note
- Breaking change! Should remove the feature flag `showStrikeAlert` in the config files.


OscarBakker (approved):
😿

PR #2513, opened 2023-02-28
[fix-5056]: kto negative feedback

vdegraaf:
Ticket: [SIG-5056](https://gemeente-amsterdam.atlassian.net/browse/SIG-5056)

### Context
The negative kto flow had a bug. When in the old flow (`reporterMailHandledNegativeContactEnabled: false`) and the user agreed to receive emails and the status of an incident went from "Verzoek tot heropenen" to "Afgehandeld" the backend did not send an email while the frontend thought it should. 
The backend was correct, since in the old flow in this scenario no email is send. 
This PR fixes the frontend so it shows a message "Email is not send", does not trigger the backend and making a note required. 

Attempt from the FE of sending an email while it should not. 
Screenshot 2023-02-27 at 14 50 35

Status after fix. Scenario of the old flow: 
Screenshot 2023-02-28 at 11 07 52

Status after fix. Scenario of the new flow: 

Screenshot 2023-02-28 at 11 07 25

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/StatusForm.test.tsx
+++ b/src/signals/incident-management/containers/IncidentDetail/components/StatusForm/StatusForm.test.tsx
@@ -11,6 +11,7 @@ import userEvent from '@testing-library/user-event'
 import fetch from 'jest-fetch-mock'
 
 import * as actions from 'containers/App/actions'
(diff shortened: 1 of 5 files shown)


bartjkdp (approved):
Thanks a lot for fixing this. Works as advertised 👍

PR #2518, opened 2023-03-06
[hotfix-5070]: Add fallback signalering and curb requests made for dashboard

OscarBakker:
Ticket: [SIG-5070](https://gemeente-amsterdam.atlassian.net/browse/SIG-5070)

## Signalen


src/signals/incident-management/containers/Dashboard/Dashboard.tsx
```diff
@@ -57,11 +58,13 @@ const Dashboard = () => {
           <AreaChart />
         </StyledRow>
       </Route>
-      <Route path={`${DASHBOARD_URL}/signalering`}>
```
vdegraaf, line 60: Why is the feature flag necessary here? This component is not rendered when the flag is set to false right? This way the page is not accessible when the feature flag is true.

src/signals/incident-management/containers/SignalingContainer/SignalingContainer.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const SignalingContainer = () => {
+  return (
+    <Row>
+      <Column span={12}>
+        <StyledH1>Signalering</StyledH1>
+      </Column>
+      <Signaling />
+    </Row>
+  )
+}
```
vdegraaf, line 21: This seems a little odd. Why is the component moved into the Dashboard container/component but then a single SignallingContainer in the containers folder that imports the component from within Dashboard.

Maybe an idea to move the signalling component back to `src/signals/incident-management/components/` and both Dashboard and Signalling container can refer to that.

src/signals/incident-management/containers/Dashboard/hooks/useDepartments.ts
```diff
@@ -35,7 +36,10 @@ export const useDepartments = (): {
           (code, value) => code + value,
           ''
         )
-        if (!cachedDepartments[departmentCodes]) {
+        if (
```
vdegraaf, line 39: A test should be updated to support this change right?

src/signals/incident-management/containers/Dashboard/hooks/useDepartments.ts
```diff
@@ -12,6 +12,7 @@ import type { Department } from 'types/api/incident'
 import type { DepartmentDetails } from 'types/api/incident'
 import type { Category } from 'types/api/incident'
 
+import configuration from '../../../../../shared/services/configuration/configuration'
```
vdegraaf, line 15: Can be written as `shared/services/configuration/configuration`. Same for the utils. 

Actually double import with line 9

src/signals/incident-management/containers/SignalingContainer/SignalingContainer.tsx
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2023 Gemeente Amsterdam
+import { Column, Heading, Row, themeSpacing } from '@amsterdam/asc-ui'
+import styled from 'styled-components'
+
+import { Signaling } from '../Dashboard/components'
+
+const StyledH1 = styled(Heading)`
```
vdegraaf, line 8: This should be in a styled file.

src/signals/incident-management/containers/SignalingContainer/SignalingContainer.tsx
OscarBakker: well actually, this is the implementation needed for signalling without dashboard. It's like an augmentation. If we're eventually moving signaling to dashboard, why not place it there.

src/signals/incident-management/containers/Dashboard/hooks/useDepartments.ts
OscarBakker: is uses the default config, I could make a test that its actually not working due to another config, you prefer?

src/signals/incident-management/containers/Dashboard/Dashboard.tsx
OscarBakker: - [ ] will delete it

src/signals/incident-management/containers/Dashboard/hooks/useDepartments.ts
OscarBakker: added a test

src/signals/incident-management/containers/SignalingContainer/SignalingContainer.tsx
vdegraaf: I get why there is a Signalling container. But not why the actual component is inside the Dashboard container while it is shared among two different containers. Also it is not sure if we will eventually move signalling into the dashboard. It is a choice for every municipality if they want to have the dashboard feature or not.

PR #2519, opened 2023-03-07
[fix-5065]: Remove container details when type: nearby

vdegraaf:
Ticket: [SIG-5065](https://gemeente-amsterdam.atlassian.net/browse/SIG-5065)

### Context
Incidents that are created based on a previous incident in the container category get the type "nearby" and some details of the old melding. In the backoffice we want to show the container details, but this is not possible. 

### Fix
When the type is set to "nearby" we don't show any container details in the backoffice.

**Bug**

**Fix**


SireeKoolenWijkstra (approved):
It's only an optical solution, but if Jelle wants that I'm fine with it.

OscarBakker (approved):

PR #2539, opened 2023-04-03
[chore] Disable dashboard

vdegraaf:
Disabling dashboard since we're not going forward with it.


SireeKoolenWijkstra (approved):
Good thinking, thank you

PR #2552, opened 2023-04-24
[fix-5134]: doorgedraaide scroller Safari

vdegraaf:
Ticket: [SIG-5134](https://gemeente-amsterdam.atlassian.net/browse/SIG-5134)

- Add overflow-y hidden to filter modal


SireeKoolenWijkstra (approved):
I can't test it on Safari, but it looks good to me. Any idea on why the content overflows on Safari?

vdegraaf:
> I can't test it on Safari, but it looks good to me. Any idea on why the content overflows on Safari?

No idea. But unfortunately browsers act slightly different...

PR #2556, opened 2023-04-25
[chore]: detail test update

vdegraaf:
Added testing for detail component using msw server


src/signals/settings/categories/components/Detail.test.tsx
```diff
@@ -1,49 +1,38 @@
...
 import { CategoryDetail } from './Detail'
 import type { Props } from './Detail'
+import * as API from '../../../../../internals/testing/api'
+import {
+  fetchMock,
+  mockRequestHandler,
+} from '../../../../../internals/testing/msw-server'
 
-jest.mock('hooks/useFetch')
+fetchMock.disableMocks()
```
OscarBakker, line 21: not in beforeAll block?

vdegraaf: This is how they implement it everywhere

vdegraaf:
> lgtm
> 
> For my understanding. Could you have kept the mocking of the hook? I mean it is shorter and all, but the useFetch hook worked before, so why did we have to change it?

Uhm in the end I could've used the useFetch mock indeed. But wanted to try this since it is agreed upon we should use the msw-server

PR #2571, opened 2023-05-10
[fix-5172]: Fix double queries on IncidentDetail

vdegraaf:
Ticket: [SIG-5172](https://gemeente-amsterdam.atlassian.net/browse/SIG-5172)


SireeKoolenWijkstra (approved):
Looking good, doing great

PR #2576, opened 2023-05-16
[feat-5060] hard stop categories / asnwers

vdegraaf:
Ticket: [SIG-5060](https://gemeente-amsterdam.atlassian.net/browse/SIG-5060)

### Context
In certain categories there are answers that are not the responsibility of the municipality. Therefore, we want to stop users to continue with the filling in the form.]

### Feature
- Add an isBlocking property to the meta object of a certain answer (this is used to show the correct message at the GlobalError)
- Add isBlockingAnswer in the validation array (This is used to make the form invalid)

## TODO

- [x] Add tests
- [x] Validate with Linda if we want to show this GlobalError message or show nothing
- [x] If it is possible to invalidate the form with just the isBlocking property on meta that would make for cleaner code


OscarBakker:
I think we should follow the normal validation flow,  thus remove the isBlocking prop. @SireeKoolenWijkstra What you think?

SireeKoolenWijkstra (approved):
lgtm, but might be not a-okay with the waps notification and the global message

PR #2590, opened 2023-06-14
[fix-5245]: Changed protected route to correct roleGroup

vdegraaf:
Ticket: [SIG-5245](https://gemeente-amsterdam.atlassian.net/browse/SIG-5245)

I'd changed the roleGroup incorrectly when adding the settings for the main category, therefore users could not see the pages anymore.


SireeKoolenWijkstra (approved):
lgtm

PR #2593, opened 2023-06-19
[fix]: Fix menu item text

vdegraaf:
Ticket: none

This way the (v1) is only shown when both pages are enabled and vice versa.


SireeKoolenWijkstra (approved):
lgtm

PR #2604, opened 2023-07-04
[feat-5213]: subcategory search in form

vdegraaf:
Ticket: [SIG-5213](https://gemeente-amsterdam.atlassian.net/browse/SIG-5213)

The goal was to implement the InputSearch also in the incidentForm.

### Changes
- Removed unnecessary SelectInputSearch
- Refactored SelectSearch to TS and removed obsolete props
- Adjust SelectSearch so it could also be implemented in CategorySelect
- Refactored CategorySelect to TS
- Adjust and refactored tests


src/signals/incident/components/form/CategorySelect/CategorySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+    useState<SubCategoryOption>()
+
+  const { value } = handler()
+
+  const getSubcategoryBySlug = useCallback(
+    (slug) => subcategoryOptions?.find((s) => s.slug === slug),
+    [subcategoryOptions]
+  )
+
+  const getSubcategoryValue = useCallback(
```
OscarBakker, line 42: also 'by' value?

vdegraaf: sharp.

OscarBakker (approved):

SireeKoolenWijkstra (approved):
lgtm. Just one thing. When making an sub report there's also a menu for subcategories. I would expect to find the same search functionality there is there is now in overview and detailpanel. (I don't know if there are even more places in the application where you can select a subcategory, I don't think so)

PR #2607, opened 2023-07-05
[chore]: Save standard texts filter

vdegraaf:
Ticket: none

Nice to have for the standard texts. This PR adds that the filter options are saved on the overview page when you return from the detail page.

### changes
- Add the filters and queryString to the IncidentManagement context
- Remove the standard text context
- Refactor IncidentManagement context to a hook for testing purposes


src/signals/incident-management/index.js
```diff
@@ -53,24 +53,38 @@ const AreaContainer = lazy(() => import('./containers/AreaContainer'))
 const SignalingContainer = lazy(() => import('./containers/SignalingContainer'))
 
 const IncidentManagement = () => {
-  // const location = useLocationReferrer()
   const districts = useSelector(makeSelectDistricts)
-  const searchQuery = useSelector(makeSelectSearchQuery)
+  const searchQueryIncidents = useSelector(makeSelectSearchQuery)
   const dispatch = useDispatch()
+  const [page, setPage] = useState(1)
```
OscarBakker, line 59: a bit strange when a page is not starting from 0

OscarBakker: i mean normally in code i would always start it from 0 and then + 1 it in the ui. (jsx)

OscarBakker (approved):
Small comment.

PR #2609, opened 2023-07-05
feat/5116-styling-feedback-standard-text-detail

SireeKoolenWijkstra:
Resolved feedback from Linda

**Changes:**

    added flex-direction to column
    add negative margin-top to styledLabel to compensate for margin 9.2px from first innerspan
    removed padding from header

Ticket: [SIG-5116](https://gemeente-amsterdam.atlassian.net/browse/SIG-5116)

## Signalen

### DIFF
--- a/src/signals/incident-management/containers/StandardTextsAdmin/components/DetailPage/styled.ts
+++ b/src/signals/incident-management/containers/StandardTextsAdmin/components/DetailPage/styled.ts
@@ -14,7 +14,6 @@ export const StyledFormFooter = styled(FormFooter)`
 
 export const StyledLabel = styled(Label)`
   font-weight: 400;
-  margin-top: -${themeSpacing(2.25)}; //first child has margin-top of 9px but distance to above element needs to be 24px
 `
 
 export const StyledTextArea = styled(TextArea)<{ showError: boolean }>`
@@ -53,4 +52,5 @@ export const RightSection = styled.div`
 `
 export const StyledColumn = styled(Column)`
   flex-direction: column;
+  margin-top: ${themeSpacing(1)};
 `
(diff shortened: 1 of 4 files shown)


src/signals/incident-management/containers/StandardTextsAdmin/components/DetailPage/styled.ts
```diff
@@ -14,6 +14,7 @@ export const StyledFormFooter = styled(FormFooter)`
 
 export const StyledLabel = styled(Label)`
   font-weight: 400;
+  margin-top: -${themeSpacing(2.25)}; //first child has margin-top of 9px but distance to above element needs to be 24px
```
vdegraaf, line 17: Three orange flags I see here: I would avoid min margins, 2.25 themespacing and comments for styling. 
Tackle it at the root. So in this case the `margin-bottom: ${themeSpacing(6)};` on the RightSection doesn't doe the job. Instead, give every element it's own margin that suits the job, if Linda insists. Because the same behaviour already exists on the Standard Texts.

OscarBakker:
this errors border is not aligning well

Screenshot 2023-07-06 at 14 03 43

src/signals/incident-management/containers/StandardTextsAdmin/components/DetailPage/styled.ts
OscarBakker: yeah agree, maybe change the border of the component above

SireeKoolenWijkstra: Hiya,
talked to Linda, she didn't like the idea of giving every element its own margin, so we agreed I would remove the margin correction on this specific element. I see the flags, but if you need to correct 9 px I don't see how you can avoid 2.25 themespacing (except for using px but in my mind that's even less desirable)

src/signals/incident-management/containers/StandardTextsAdmin/components/PageHeader/styled.ts
```diff
@@ -6,7 +6,6 @@ import styled, { css } from 'styled-components'
 export const StyledSection = styled.section<{ hasBackLink: boolean }>`
   contain: content;
   padding-top: ${themeSpacing(6)};
```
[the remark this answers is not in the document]
SireeKoolenWijkstra: Ok, cool I'll keep that in mind.

PR #2616, opened 2023-07-12
Fix display error for empty IncidentMap

bartjkdp:
This fixes a display error for an empty IncidentMap.

**Before**

image

**After**

image


vdegraaf:
Hi @bartjkdp,
Cannot reproduce the bug, but doesn't seem to harm our side. One test is failing however, can you look into that?

bartjkdp:
@vdegraaf thanks for commenting. Think this was a flaky test (and did not have a re-run button as external contributor). Rebased it on upstream/main, now the test succeeds.

vdegraaf (approved):
@bartjkdp was able to reproduce indeed. Thanks for the fix! I'll merge it.

PR #2627, opened 2023-07-19
[bug] Show name, not public name in SelectedSubcategories

OscarBakker:
Ticket: [SIG-5117](https://gemeente-amsterdam.atlassian.net/browse/SIG-5117)

## Signalen


SireeKoolenWijkstra (approved):
lgtm but why we have a public name if it is not the one we use in public is beyond me.

vdegraaf (changes requested):
I don't think this is a bug. We should show the `public_name` if it exists and else show the name.

vdegraaf (approved):
Alright. Might be a bug. Apperently the public_name is only for the civilians accessible websites..

PR #2638, opened 2023-07-28
Add FRONTEND_TAG to Github Container Registry builds as build argument

bartjkdp:
Currently frontend version tags are not included in builds that are pushed to Github Container Registry.


vdegraaf (approved):
LGTM!

PR #2639, opened 2023-07-31
[feat-5266]: kto over iets anders

vdegraaf:
Ticket: [SIG-5266](https://gemeente-amsterdam.atlassian.net/browse/SIG-5266)

To enable open question for the KTO process I had to change the checkbox and radiobutton lists. 

### Changes
- render textarea when a checkbox or radiobutton option has the open_answer property
- Add form methods that updates the form and validates the input
- Remove the hardcoded "over iets anders"-logic
- Implement utils that map over the answers and pair the correct checked box with the open answer
- Add tests

### DIFF
--- a/src/components/RadioButtonList/RadioButtonList.tsx
+++ b/src/components/RadioButtonList/RadioButtonList.tsx
@@ -114,7 +114,7 @@ const RadioButtonList = <T extends RadioButtonOption>({
     formValidation: RadioButtonListProps['formValidation']
   ) => {
     let component = (
-      <>
+      <Fragment key={option.key || option.name}>
         <StyledLabel
           key={option.key || option.name}
           htmlFor={option.key || option.name}
@@ -168,7 +168,7 @@ const RadioButtonList = <T extends RadioButtonOption>({
               />
             </FormField>
           )}
-      </>
+      </Fragment>
     )
 
     if (radioOptions.some((option) => option.topic)) {
--- a/src/components/RadioButtonList/RadioButtonList.tsx
+++ b/src/components/RadioButtonList/RadioButtonList.tsx
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
-// Copyright (C) 2019 - 2021 Gemeente Amsterdam
-import type { FunctionComponent } from 'react'
+// Copyright (C) 2019 - 2023 Gemeente Amsterdam
 import { Fragment } from 'react'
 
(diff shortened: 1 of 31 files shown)


src/components/RadioButtonList/RadioButtonList.tsx
```diff
@@ -90,27 +108,67 @@ const RadioButtonList = <T extends RadioButtonOption>({
...
+                formValidation.errors[`open_answer-${option.value}`]?.type ===
+                errorType
+              }
+              getError={(errorType) =>
+                formValidation.errors[`open_answer-${option.value}`]?.type ===
+                errorType
+              }
+            >
+              <TextArea
+                {...formValidation.register(`open_answer-${option.value}`)}
```
OscarBakker, line 157: Is the set value necessary and the trigger, I guess register already adds there through descructuring.

OscarBakker: I see that onchange does a thing under the hood, its checks the max content and shows chars left do never mind

src/components/RadioButtonList/RadioButtonList.tsx
```diff
@@ -139,7 +197,9 @@ const RadioButtonList = <T extends RadioButtonOption>({
         id={id}
         {...rest}
       >
-        {radioOptions.map((option, index) => renderRadioButton(option, index))}
+        {radioOptions.map((option, index) =>
+          renderRadioButton(option, index, formValidation)
```
OscarBakker, line 201: I see this was already a function, but would it not be better just add jsx here? or make a special Button component within RAdioButtonList/

src/signals/incident/containers/KtoContainer/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (index !== -1) {
+      resultArray.splice(index + 1, 0, value)
+    }
+  })
+
+  return resultArray
+}
+
+export const getMergedOpenAnswers = (object: FormData): FormData => {
+  const openAnswerPairs: Record<string, any> = {}
```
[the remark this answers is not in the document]
vdegraaf: This gives some typing errors, so left it as is.

src/components/RadioButtonList/RadioButtonList.tsx
vdegraaf: You're right!

OscarBakker (approved):

PR #2642, opened 2023-08-01
[chore]: Refactor radioButtonList

vdegraaf:
Ticket: none

Refactor radiobuttonList so that the buttonOption component as it's own file.

### DIFF
--- a/src/components/RadioButtonList/RadioButtonOption.tsx
+++ b/src/components/RadioButtonList/RadioButtonOption.tsx
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2019 - 2023 Gemeente Amsterdam
+
+import { Fragment } from 'react'
+
+import FormField from 'components/FormField'
+import RadioButtonComponent from 'components/RadioButton'
+import TextArea from 'components/TextArea'
+
+import type { RadioButtonListProps } from './RadioButtonList'
+import { StyledLabel } from './styled'
+import type { RadioButtonOption as RadioButtonOptionType } from './types'
+import TopicLabel from '../TopicLabel'
+
+interface Props {
+  defaultValue?: string
+  formValidation: RadioButtonListProps['formValidation']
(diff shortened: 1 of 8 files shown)


src/components/RadioButtonList/RadioButtonOption.tsx
```diff
@@ -0,0 +1,103 @@
...
+  defaultValue,
+  formValidation,
+  groupName,
+  index,
+  option,
+  onChange,
+  radioOptions,
+}: Props) => {
+  let component = (
+    <Fragment key={option.key || option.name}>
```
OscarBakker, line 35: why one of both?

src/components/RadioButtonList/RadioButtonOption.tsx
```diff
@@ -0,0 +1,103 @@
...
+              errorType
+            }
+            getError={(errorType) =>
+              formValidation.errors[`open_answer-${option.value}`]?.type ===
+              errorType
+            }
+          >
+            <TextArea
+              {...formValidation.register(`open_answer-${option.value}`)}
+              maxRows={5}
```
OscarBakker, line 76: could come from default value but still configurable.

src/components/RadioButtonList/RadioButtonOption.tsx
OscarBakker: seems key is always there right?

OscarBakker (approved):
some suggestions, but already approved!

src/components/RadioButtonList/RadioButtonOption.tsx
vdegraaf: Good point. Seems like it yes. Didn;t change the old code though. But will fix it.

vdegraaf: Never mind, it can be an empty string. Then you want to take the name property.

Screenshot 2023-08-02 at 09 53 10

PR #2649, opened 2023-08-07
[fix-5362]: Standaard teksten opslaan

vdegraaf:
Ticket: [SIG-5362](https://gemeente-amsterdam.atlassian.net/browse/SIG-5362)

Fixes:
- We redirected too fast to the overview page, therefore the post/patch request never came through
- Removed timeout since it didn't work. The overview page wasn't updated. Will create a backend story to wait until the search endpoint is updated before returning a success
- Changed the validation so that the validation error is removed when a category is selected


SireeKoolenWijkstra (approved):
Lgtm.

PR #2650, opened 2023-08-09
[fix-5292]: disappearing notification

vdegraaf:
Ticket: [SIG-5292](https://gemeente-amsterdam.atlassian.net/browse/SIG-5292)

The notification of an action was barely noticeable since it was removed very quickly. That was caused by the useFetchResponseNotification that was set to TYPE_LOCAL. 

- Changed it to TYPE_GLOBAL so that it does not disappear when navigating to another page.
- Changed the functionality of TYPE_GLOBAL slightly so that is disappears after 8 seconds, just likt a TYPE_LOCAL. Only when it has variant ERROR it needs te be closed with the close button
- Moved the isFetchError typeguard to a shared location

### DIFF
--- a/src/signals/shared/type-guards/index.test.ts
+++ b/src/signals/shared/type-guards/index.test.ts
@@ -1,0 +1,43 @@
+import { isFetchError } from './index'
+
+describe('isFetchError', () => {
+  it('should returns true for a FetchError with detail property', () => {
+    const fetchError = { detail: 'An error occurred' }
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const result = isFetchError(fetchError)
+
+    expect(result).toBe(true)
+  })
+
+  it('should returns true for a FetchError with message property', () => {
+    const fetchError = { status: 404, message: 'Not Found' }
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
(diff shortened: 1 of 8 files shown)


src/signals/shared/type-guards/index.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  it('should returns false for a FetchError without detail property', () => {
+    const fetchError = { status: 404, result: 'Not Found' }
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const result = isFetchError(fetchError)
+
+    expect(result).toBe(false)
+  })
+
+  it('should returns false for a boolean value', () => {
```
OscarBakker, line 31: return

src/signals/shared/type-guards/index.ts
```diff
@@ -0,0 +1,10 @@
+import type { FetchError } from 'hooks/useFetch'
+
+export const isFetchError = (
+  error?: boolean | FetchError
+): error is FetchError => {
```
OscarBakker, line 5: whats happing here?

vdegraaf: It's a type guard:) In some cases the error can only be a Fetch error, this make sure only Fetch errors pass. Else you get a lot of typing issues. 
https://www.typescriptlang.org/docs/handbook/advanced-types.html

PR #2658, opened 2023-08-16
[feat-5195]: Add contact change form

OscarBakker:
add a form with with email and phone validation. Add styling. Retrieve history changes after posting reporter change.

**for testers**
- design I changed for consistency. The margins, paddings, gaps are like status in the right column.
- the phonenumber changed state from the design cannot be made from the data we got. Thereby we dont show this state. in the form its ```(gewijzigd)```

Ticket: [SIG-5195](https://gemeente-amsterdam.atlassian.net/browse/SIG-5195)

## Signalen

### DIFF
--- a/src/signals/incident-management/containers/IncidentDetail/components/Detail/components/Contact/Contact.tsx
+++ b/src/signals/incident-management/containers/IncidentDetail/components/Detail/components/Contact/Contact.tsx
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2023 Gemeente Amsterdam
+import {
+  Fragment,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react'
+
+import { useParams } from 'react-router-dom'
+
+import Edit from './Edit'
+import { StyledEditButton, StyledLink } from './styled'
+import { useFetch } from '../../../../../../../../hooks'
+import configuration from '../../../../../../../../shared/services/configuration/configuration'
+import type { Incident } from '../../../../../../../../types/incident'
+import IncidentDetailContext from '../../../../context'
+import type { Result, SignalReporter } from '../../../../types'
+
+type Props = {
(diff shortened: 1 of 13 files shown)


vdegraaf (commented):
Looking good, but still needs some work. Saving the data is a bit flaky and I cannot test the whole flow since the cancel button is not shown. See the comment. For the design I would contact Linda. I think the paddings are a big large. 

Functional:
- When i have no phone number and add one, It doesn't update immediately. If I type "06", it removes the "0"
**validation change!**
- Should check if it's a valid phone number or not
**true, will add**

- What is shown in the history? I did multiple changes, but only see one message about "Verzoek tot wijzigen contactgegevens geannuleerd."
**does it matter for this story? maybe a good one for the interation with the backend**

- If I change the email, it doesn't show a button with "verificatie annuleren" en a message "(verificatie open)". It also shows the old email instead of the new one (this is what is shown in design) 

- If i try to change the email again, I see in the network tab some request outgoing, but the edit box doesn't close. 
**will fix**

src/signals/incident-management/containers/IncidentDetail/components/Detail/components/Contact/Contact.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const params = useParams()
+
+  const onClose = useCallback(() => {
+    setShowEdit(false)
+  }, [])
+
+  const submit = useCallback(
+    (data) => {
+      data.sharing_allowed = incident.reporter.sharing_allowed
```
vdegraaf, line 41: What's happening here?

OscarBakker: the BE needs that value for the post to succeed

OscarBakker:
> 
fixed all points but:
If I don't have a phone number and try to change the email adres, it throws an error that the a phone number is mandatory

now waiting for be

vdegraaf (approved):
LGTM

PR #2664, opened 2023-08-22
[fix-5366]: accessible objects map

vdegraaf:
Ticket: [SIG-5366](https://gemeente-amsterdam.atlassian.net/browse/SIG-5366)


docs/adr/0016-accessibility-modal-focus.md
```diff
@@ -0,0 +1,15 @@
+# Handling tab focus within modals
```
OscarBakker, line 1: nice

OscarBakker (approved):
looking good

PR #2668, opened 2023-08-24
[feat-5130]: Change css so solution works on multiple browsers

vdegraaf:
Ticket: [SIG-5130](https://gemeente-amsterdam.atlassian.net/browse/SIG-5130)


J3vdS (approved):
LGTM

PR #2676, opened 2023-08-28
[chore]: refactor file input renderer to ts

vdegraaf:
Ticket: [SIG-5383](https://gemeente-amsterdam.atlassian.net/browse/SIG-5383)

Refactor FileInputRenderer and FileInput to TS.


src/signals/incident/components/form/FileInput/FileInput.tsx
```diff
@@ -27,7 +32,7 @@ const FileInput = ({ handler, parent, meta }) => {
   )
 
   const checkNumberOfFiles = useCallback(
-    (file, index) => index < maxNumberOfFiles,
+    (_file, index) => index < maxNumberOfFiles,
```
OscarBakker, line 35: personally I would just use a lodash. " _ "
in some languages private methods are prefixed with an underscore, which might be a little confusing here.

src/signals/incident/components/form/FileInputRenderer/FileInputRenderer.tsx
```diff
@@ -1,10 +1,20 @@
 // SPDX-License-Identifier: MPL-2.0
```
OscarBakker, line 1: suggestion: maybe we should put __tests/files into the parent ,cd ../, directory?!

OscarBakker (approved):
a vote of confidence

PR #2688, opened 2023-09-04
[chore]: Make FileInput optional

vdegraaf:
Ticket: none

Make label optional because when the FileInput is rendered in FormField the FormField has already a label.


SireeKoolenWijkstra (approved):
lgtm

PR #2694, opened 2023-09-07
[fix-5430]: Centralize useDeviceMode hook

vdegraaf:
Ticket: [SIG-5430](https://gemeente-amsterdam.atlassian.net/browse/SIG-5430)

While fixing this bug I centralized the useDeviceMode hook so it's globally available and easier in use.

To fix the bug I added some focusTrapOptions when on mobile.

### DIFF
--- a/src/hooks/useDeviceMode.ts
+++ b/src/hooks/useDeviceMode.ts
@@ -10,7 +10,6 @@ export enum DeviceMode {
 }
 
 export interface ModeProp {
-  // prefixing mode with $ to prevent prop bleeding through to the DOM
   $mode: DeviceMode
 }
 
--- a/src/hooks/useDeviceMode.ts
+++ b/src/hooks/useDeviceMode.ts
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2022 - 2023 Gemeente Amsterdam
+import { useEffect, useState } from 'react'
+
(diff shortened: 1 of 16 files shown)


src/hooks/useDeviceMode.ts
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (C) 2022 - 2023 Gemeente Amsterdam
+import { useEffect, useState } from 'react'
+
+import { sizes } from '@amsterdam/asc-ui/lib/theme/default/breakpoints'
+
+export enum DeviceMode {
```
OscarBakker, line 7: Pin is using this enum as well. should this enum be extracted to a higher level?

src/hooks/useDeviceMode.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface ModeProp {
+  $mode: DeviceMode
+}
+
+function getDeviceMode(size: number) {
+  if (size <= sizes.tabletM) {
+    return DeviceMode.Mobile
+  } else {
+    return DeviceMode.Desktop
```
OscarBakker, line 20: unnecessary return

OscarBakker (commented):
Good job

src/hooks/useDeviceMode.ts
vdegraaf: good one. I put is in `types/device' on toplevel

OscarBakker (approved):

PR #2705, opened 2023-09-18
[feat-5454]: Aligned Attachment types

vdegraaf:
Ticket: [SIG-5454](https://gemeente-amsterdam.atlassian.net/browse/SIG-5454)

### Changes
Three different types of attachments:
-  incidentDetail management
-  then a mapped one
-  A different "public" on on my-incidents

Did align the types with what is actually passed as data. Many times undefined was used while the actual value was null.

### DIFF
--- a/src/components/AttachmentViewer/AttachmentViewer.tsx
+++ b/src/components/AttachmentViewer/AttachmentViewer.tsx
@@ -27,11 +27,13 @@ import {
   Title,
   Wrapper,
 } from './styles'
-// TODO: add caption?
+
 export interface FormattedAttachment {
-  location: string
+  caption: string | null
   createdAt: string | null
   createdBy: string | null
+  location: string
+  stateShown: string
 }
 
 interface Props {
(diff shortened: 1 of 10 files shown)


src/components/AttachmentViewer/AttachmentViewer.tsx
```diff
@@ -29,15 +29,15 @@ import {
   Wrapper,
 } from './styles'
 
-export interface Attachment {
+export interface FormattedAttachment {
   location: string
-  createdAt?: string
-  createdBy?: string
+  createdAt: string | null
```
SireeKoolenWijkstra, line 34: Just for my curiosity. Yesterday you gave feedback that you preferred ?: string to string | undefined since that was shorter and our coding style. Now we change from ? string to string | null. I get that if null is possible it is preferred to undefined, but doesn't the ? indicate the value could be there, it could also not be there?

vdegraaf: Not sure what you mean with ", but doesn't the ? indicate the value could be there, it could also not be there?" But if you mean that a question mark can represent `null` then that is incorrect. And this is why we use null: 

Screenshot 2023-09-19 at 08 40 07

PR #2718, opened 2023-09-27
[bugfix-5487] stop unnecessary calls to the BE from the incidentmap

SireeKoolenWijkstra:
On the incidentmap, when a user used either the GPSLocation or the search bar for an address, the application entered a loop of calls to the BE. This was due to a dependency on a function (isMobile) that referenced a new function with every render.

Also noticed that Leaflet has a new update and updated the package. Saw no weird stuff with the maps.

Ticket: [SIG-5487](https://gemeente-amsterdam.atlassian.net/browse/SIG-5487)

## Signalen

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "immutable": "^4.0.0-rc.12",
         "invariant": "^2.2.4",
         "keycloak-js": "^20.0.3",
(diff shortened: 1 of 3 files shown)


vdegraaf (approved):
Works like a charm!

PR #2728, opened 2023-10-12
[feat-5434]:Add locatie kaart mobiel

OscarBakker:
change mobile styling address / container search.

Ticket: [SIG-5434](https://gemeente-amsterdam.atlassian.net/browse/SIG-5434)

## Signalen


src/signals/incident/components/form/MapSelectors/Asset/Selector/DetailPanel/DetailPanel.tsx
```diff
@@ -21,50 +19,30 @@ import {
...
-    iconUrl: '/assets/images/area-map/icon-pin.svg',
-    iconSize: [40, 40],
-  },
-  typeValue: NEARBY_TYPE,
-}
-
-const DetailPanel: FC<DetailPanelProps> = ({ language = {} }) => {
-  const shouldRenderAddressPanel = useMediaQuery({
+const DetailPanel: FC<DetailPanelProps> = ({ language }) => {
+  const shouldRenderMobileVersion = useMediaQuery({
```
SireeKoolenWijkstra, line 41: Don't we have a custom made hook for that useDeviceMode?

SireeKoolenWijkstra (commented):
funtional tests vindings:
- [ ] The sticky panell (footer) is not so sticky. I think it should be about 16 px from the bottom.
- [ ] When I select an address, but then hit the back button it keeps the selected address. I know this is behaviour that is already in the app, but it's not behaviour I would expect. I expect to not have selected an address. 
- [ ] kudos for the positioning of the back button when entering input in the searchbar. It looks much better than the the design. 
- [ ] the distance between the back icoon and the text Zoek op adres... is 8px, but should be 12 px according to design. 
- [ ] the letter size is too big, should be 1.125 of the Zoek of adres-text
- [ ] there is no icon in the legenda button and the text has changed from Uitleg to Legenda (it wasn't part of the acc critt, though so we can take it up in a different story)

SireeKoolenWijkstra (approved):

PR #2739, opened 2023-10-24
[bugfix]: implement useWatch to see if changes occur in inputfield of form

SireeKoolenWijkstra:
**Implement useWatch**
FormValues are not always updated because Detail is not always rendered anew. After 3 clicks a bug appears. Therefor implemented useWatch, so updates are monitored and only the effected component is rerendered

## Signalen

### DIFF
--- a/src/signals/settings/categories/components/CategoryForm.test.tsx
+++ b/src/signals/settings/categories/components/CategoryForm.test.tsx
@@ -29,7 +29,6 @@ const mockFormValues = {
 }
 
 const defaultProps: Omit<Props, 'formMethods'> = {
-  formValues: mockFormValues,
   history: [
     {
       identifier: 'UPDATED_CATEGORY_217043',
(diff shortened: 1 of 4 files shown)


vdegraaf (commented):
@SireeKoolenWijkstra not sure what the bug is I should be testing. Is there a bug ticket? Or else a way to reproduce? Since I do not know where I should "click three times" and the bug appears.

image
 If you uncheck and check the openbaar tonen checkbox a couple of times the Naam Openbaar (fieldgroup public_name) doesn't reappear. There is no bug ticket unfortunately.

vdegraaf (commented):
The fix works! But you can do it with less code by just adding `formMethods.watch('is_public_accessible')` to the Detail component. Then it will trigger a rerender when that field changes.

Changed it. Thanks.

vdegraaf (commented):
I'd expect just one line of code added. Why did you change all the rest?

vdegraaf (approved):

PR #2750, opened 2023-11-09
[chore-5471]: Change label color with newDesignDystem prop

vdegraaf:
Ticket: [SIG-5471](https://gemeente-amsterdam.atlassian.net/browse/SIG-5471)

Changing color labels when design according to new design system

[SIG-5471]: https://gemeente-amsterdam.atlassian.net/browse/SIG-5471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Label/Label.test.tsx
+++ b/src/components/Label/Label.test.tsx
@@ -52,6 +52,6 @@ describe('signals/incident-management/components/Label', () => {
 
     expect(screen.getByText('Label 1')).not.toHaveStyleRule('color', 'inherit')
     expect(screen.getByText('Label 2')).toHaveStyleRule('color', 'inherit')
-    expect(screen.getByText('Label 2')).toHaveStyleRule('color', 'inherit')
+    expect(screen.getByText('Label 3')).toHaveStyleRule('color', 'inherit')
   })
 })
--- a/src/components/Label/Label.test.tsx
+++ b/src/components/Label/Label.test.tsx
@@ -42,11 +42,16 @@ describe('signals/incident-management/components/Label', () => {
           <Label isGroupHeader>Label 1</Label>
 
           <Label>Label 2</Label>
+
+          <Label isGroupHeader isNewDesignSystem>
+            Label 3
+          </Label>
         </Fragment>
       )
     )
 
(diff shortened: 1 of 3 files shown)


src/components/Label/Label.test.tsx
```diff
@@ -42,11 +42,16 @@ describe('signals/incident-management/components/Label', () => {
...
+          <Label isGroupHeader isNewDesignSystem>
+            Label 3
+          </Label>
         </Fragment>
       )
     )
 
     expect(screen.getByText('Label 1')).not.toHaveStyleRule('color', 'inherit')
     expect(screen.getByText('Label 2')).toHaveStyleRule('color', 'inherit')
+    expect(screen.getByText('Label 2')).toHaveStyleRule('color', 'inherit')
```
SireeKoolenWijkstra, line 55: Shouldn't this be label 3 ?

src/components/Label/Label.tsx
```diff
@@ -1,13 +1,15 @@
...
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
 
 interface LabelProps {
   /** When true, will render the label with a plain font instead of a bold font */
   inline?: boolean
   /** When false, the Label component will render as an inline-block element without the red header colour */
   isGroupHeader?: boolean
+  /** when true, a label is also black when rendered as group header */
+  isNewDesignSystem?: boolean
```
SireeKoolenWijkstra, line 12: Will this be removed at some time in the future? Because the name (is NewDesignSystem) doesn't clue me in about group headers and that's used for that. Right now I'm thinking this has something to do with the NL design styling..

OscarBakker (approved):
The new design system maybe should have a version number or reference. Well looks good!

Good point. Referenced the new design system for now.

src/components/Label/Label.test.tsx
vdegraaf: sharp.

src/components/Label/Label.tsx
vdegraaf: True. I referred to the new design system for now.

PR #2765, opened 2023-11-28
[feat-5485]: street name only search filterform 

vdegraaf:
Context
Feedback from testers was that beaviour of showing address in inputfield inconsistent was. Sometimes it showed just the streetname, sometimes it showed the streetname and the municipality. Also discovered that clicking Nieuw Filter didn't clear the value of the inputfield.
Now only unique streetnames are shown

Changes
dependent on prop streetNameOnly will serviceParams and formatPdokResponse change their behaviour.
Filtering for unique streetnames
removed defaultvalue from Autosuggest. -> Autosuggest seems to be a mix of uncontrolled and controlled feautures (with a useState to set the defaultvalue and an inPutRef). Story has been created (https://gemeente-amsterdam.atlassian.net/browse/SIG-5574) in Jira for refactoring.
tests for meeting threshold added

For testers
PDOK sends back all matches, so also the matches with the municipality. This shows most when searching for Ams. Can't do nothing about that.

Ticket: [SIG-5485](https://gemeente-amsterdam.atlassian.net/browse/SIG-5485)

[SIG-5485]: https://gemeente-amsterdam.atlassian.net/browse/SIG-5485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
+++ b/src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
@@ -43,6 +43,28 @@ const renderAndSearch = async (value = 'Dam', props = {}) => {
   await screen.findByTestId('auto-suggest')
 }
 
+const renderAndSearchWithStreetNameOnlyAsTrue = async (
+  value = 'dor',
+  props = {}
+) => {
+  render(
+    withAppContext(
+      <PDOKAutoSuggest onSelect={onSelect} streetNameOnly={true} {...props} />
+    )
+  )
+  const input = screen.getByRole('textbox') as HTMLInputElement
+
+  input.focus()
+
+  fireEvent.change(input, { target: { value } })
+
+  act(() => {
+    jest.advanceTimersByTime(INPUT_DELAY)
+  })
+
+  await screen.findByTestId('auto-suggest')
+}
+
 describe('components/PDOKAutoSuggest', () => {
   beforeEach(() => {
     fetch.mockResponse(mockResponse)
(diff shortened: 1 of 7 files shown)


src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
```diff
@@ -43,6 +43,28 @@ const renderAndSearch = async (value = 'Dam', props = {}) => {
   await screen.findByTestId('auto-suggest')
 }
 
+const renderAndSearchWithStreetNameOnlyAsTrue = async (
+  value = 'dor',
+  props = {}
+) => {
+  render(
+    withAppContext(
+      <PDOKAutoSuggest onSelect={onSelect} streetNameOnly={true} {...props} />
```
OscarBakker, line 52: you can leave {true}

OscarBakker: why not add a third arg to renderAndSearch with streetNameOnly?

OscarBakker (changes requested):

src/components/PDOKAutoSuggest/PDOKAutoSuggest.test.tsx
SireeKoolenWijkstra: Euh, isn't streetNameOnly by default false and that's why I set it here expressly to true?

vdegraaf (commented):
@SireeKoolenWijkstra wasn't it a criteria that only unique street names were shown in the dropdown?

vdegraaf (commented):
Looks good! But cannot approve because I made the PR. -> Nice! Ik zal hem mergen

SireeKoolenWijkstra (approved):
lgt Vincent, dus heb ik de PR gemerged.

PR #2772, opened 2023-11-30
[chore] When address response is within 100ms, clear timeout and set new address

OscarBakker:
**context**
when you click on the map on an object or just a new location there used to be a flickering. New address coming in from the reverse proxy endpoint and for an under the 100 ms no address. Therefore I build a threshold of 100ms that it will use the last known address before replacing it with the new one(or none). 

This will make sure the new address from the reserve proxy endpoint will be used instantly.

## Signalen


SireeKoolenWijkstra (approved):
lgtm

PR #2776, opened 2023-12-05
[feat-5551]: Add delete functionality

vdegraaf:
Ticket: [SIG-5551](https://gemeente-amsterdam.atlassian.net/browse/SIG-5551)

[SIG-5551]: https://gemeente-amsterdam.atlassian.net/browse/SIG-5551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/signals/incident-management/containers/IncidentSplitContainer/components/IncidentSplitForm/__tests__/IncidentSplitForm.test.js
+++ b/src/signals/incident-management/containers/IncidentSplitContainer/components/IncidentSplitForm/__tests__/IncidentSplitForm.test.js
@@ -68,8 +68,11 @@ describe('IncidentSplitForm', () => {
 
     expect(onSubmit).toHaveBeenCalledWith({
       department: parentIncidentFixture.directingDepartment,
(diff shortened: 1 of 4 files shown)


SireeKoolenWijkstra (changes requested):
At first glance:
- [ ] If I add more than one subissue, I would expect to also be able to delete subissue 1. There is no icon there and it also doesn't appear when I add another subissue.
- [ ] In continuation on that. When I've added subissues, clicked Opslaan and then want to add more subissues, I would expect to see immediately teh deleteIcon.

SireeKoolenWijkstra (approved):
Er zit nog een linterwarning, maar anders kan ie door

vdegraaf:
> Er zit nog een linterwarning, maar anders kan ie door

Thanks. The linting is not related to the changes in this PR. So I won't fix it here since I'm not sure if I beak anything.

PR #2785, opened 2023-12-13
[fix]: object select bugs

vdegraaf:
Ticket: none

Fixing:

- When on desktop and in container of light category, the list of object in the detail panel are endless. Now the submit button is hidden at the end of the list. We want it sticky
- It was possible to select endless amount of ligths and containers regardless of the max asset warning. Fixed this issue.
- Removed useCurrentAddress since it didn't have any impact it seemed


src/signals/incident/components/form/MapSelectors/Asset/Selector_v2_removeafterfinishepic5440/DetailPanel/styled.tsx
```diff
@@ -27,9 +26,29 @@ export const StyledAssetList = styled(AssetList)`
   }
 `
 
-export const StyledButton = styled(Button)`
-  margin-top: ${themeSpacing(4)};
+export const StyledButton = styled(Button)<{
+  $isMobile?: boolean
+  $hasSubmitButton?: boolean
+}>`
+  position: fixed;
```
OscarBakker, line 33: better yet to use position sticky. now its positioned to the screen, with sticky ''offset relative to its nearest scrolling ancestor and [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block)''

OscarBakker (approved):
looks good. one small comment

src/components/IconList/IconList.tsx
```diff
@@ -42,21 +44,31 @@ export const IconListItem = ({
...
     undefined
   )
-  const [disabled, setDisabled] = useState<boolean>(false)
+  const { maxAssetWarning } = useSelector(makeSelectMaxAssetWarning)
+
+  useEffect(() => {
+    setCheckedState(checked)
+  }, [checked, maxAssetWarning])
+
+  const disableOnClick = useCallback(() => {
```
OscarBakker, line 53: this should be a useMemo

PR #2802, opened 2024-01-08
[fix-5574] : autosuggest remove default value

vdegraaf:
Ticket: [SIG-5574](https://gemeente-amsterdam.atlassian.net/browse/SIG-5574)

The name "default value" was actually used to show and hide the inline button. So changed it.

[SIG-5574]: https://gemeente-amsterdam.atlassian.net/browse/SIG-5574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/AutoSuggest/AutoSuggest.tsx
+++ b/src/components/AutoSuggest/AutoSuggest.tsx
@@ -74,7 +74,7 @@ const AutoSuggest = ({
   showListChanged,
   ...rest
 }: AutoSuggestProps) => {
-  const [showClearButton, setShowClearButton] = useState(!!value)
+  const [showInlineButton, setShowInlineButton] = useState(!!value)
 
   const [data, setData] = useState<RevGeo>()
   const [initialRender, setInitialRender] = useState(false)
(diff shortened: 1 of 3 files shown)


src/components/AutoSuggest/AutoSuggest.tsx
```diff
@@ -387,7 +386,7 @@ const AutoSuggest = ({
             variant="blank"
           />
         ) : (
-          <ClearInput
+          <InlineButton
```
alimpens, line 389: Maybe out of scope, but: this renders a button that doesn't really do anything other than setting the focus on the input. Shouldn't this just render an icon? Similar to what they do [here](https://onderzoek.amsterdam.nl/zoek)

vdegraaf: True. But then the input is not focussed anymore. Will leave it for now.

alimpens (approved):

PR #2810, opened 2024-01-29
[fix]: Fix api key bug

alimpens:
Always setting an X-Api-Key header for WFS layer fetches causes issues for Den Bosch.

See mail 'Koppeling met MOON kapot na nieuwe versie' on 29-1-2024.

This PR makes sure that we're only setting a header when X-Api-Key is defined in the config.


vdegraaf (commented):
Works as expected! But should there be a test case?

vdegraaf (approved):
LGTM! One comment: don't think "async" is necessary in both cases.

vdegraaf (commented):
@alimpens Also a linting error

PR #2816, opened 2024-02-06
[fix]: Add formatISO type bug tests

alimpens:
No ticket.

[This PR](https://github.com/Amsterdam/signals-frontend/pull/2813) created a bug when trying to submit a _melding_ where you answered ‘Nu’ to the question ‘Wanneer was het?’. [This PR](https://github.com/Amsterdam/signals-frontend/pull/2815) fixed that bug.

The current PR adds tests to prevent that bug from happening in the future.

To discuss:

- Maybe we want to turn the check around? So:

```
      typeof incident.dateTime === 'number'
        ? formatISO(incident.dateTime as number),
        :  formatISO(Date.now())
```

- Should we change the [Incident type](https://github.com/Amsterdam/signals-frontend/blob/b53f397d5efe5fc127eb0a4813e50a27dd37fb73/src/types/incident.ts#L25) as well? It says it can be `number | null | string`, but it can be `undefined` as well.

### DIFF
--- a/src/signals/incident/services/map-controls-to-params/index.test.ts
+++ b/src/signals/incident/services/map-controls-to-params/index.test.ts
@@ -55,4 +55,22 @@ describe('map-controls-to-params', () => {
       varFromMapPaths: 'bar',
     })
   })
+
+  it('should return the current date when dateTime is `now`', () => {
+    const incidentWithDateTimeNow = { ...mock, dateTime: 'now' }
+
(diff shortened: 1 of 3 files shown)


vdegraaf:
Good suggestion! Let's change it to: 

`typeof incident.dateTime === 'number'
        ? formatISO(incident.dateTime),
        :  formatISO(Date.now())`

vdegraaf (approved):
LGTM

PR #2819, opened 2024-02-08
[fix]: Don't use addNote label when used in form

alimpens:
No ticket

Textarea's in step 2 showed the label twice, because is was added in TextareaInput and again in AddNote.
This PR makes it possible to pass an `inForm` prop to Addnote, which doesn't render the label in AddNote.


vdegraaf (approved):
LGTM

PR #2822, opened 2024-02-15
[fix]: Remove optional time selector

alimpens:
Making the time selector optional caused an error when you answered Wanneer was het > Eerder > Vandaag before 9.00. Without a time selector, the datetime is always 9.00, which means the `inPast` validation isn't met before 9.00.

Discussed this with UX and decided to remove the optional time selector.


vdegraaf (approved):
LGTM

PR #2840, opened 2024-04-09
[feat-5683]: Allow 'Assigned user email' column to be sorted

vdegraaf:
Ticket: [SIG-5683](https://gemeente-amsterdam.atlassian.net/browse/SIG-5683)

Includes some refactoring:
- Changing keys to English
- Sorting information is now a label instead of key, so removed the key from `type SortOption`
- [bug] When `fetchDistrictsFromBackend` is enabled only the header title was changed. However, the query should have changed too to use `area_name` instead of "stadsdeel". 

[SIG-5683]: https://gemeente-amsterdam.atlassian.net/browse/SIG-5683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
+++ b/src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
@@ -162,7 +162,7 @@ const List: FunctionComponent<ListProps> = ({
             <ThSort
               StyledComponent={ThPriority}
               sortOption={SortOptions.PRIORITY_ASC}
-              headerText={SortOptionLabels.URGENTIE}
+              headerText={SortOptionLabels.PRIORITY}
               ordering={ordering}
               changeOrder={changeOrder}
               sortingDisabled={sortingDisabled}
@@ -179,7 +179,7 @@ const List: FunctionComponent<ListProps> = ({
             <ThSort
               StyledComponent={ThDate}
               sortOption={SortOptions.CREATED_AT_DESC}
-              headerText={SortOptionLabels.DATUM}
+              headerText={SortOptionLabels.DATE}
               ordering={ordering}
               changeOrder={changeOrder}
             />
@@ -205,7 +205,7 @@ const List: FunctionComponent<ListProps> = ({
               headerText={
                 configuration.featureFlags.fetchDistrictsFromBackend
                   ? configuration.language.district
-                  : SortOptionLabels.STADSDEEL
+                  : SortOptionLabels.DISTRICT
               }
               ordering={ordering}
               changeOrder={changeOrder}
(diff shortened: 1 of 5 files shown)


src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
```diff
@@ -33,7 +33,6 @@ import {
   StyledList,
   Table,
   TdStyle,
-  Th,
```
alimpens, line 36: If this `Th` isn't used anymore, can we remove it from `styles.ts` as well?

src/signals/incident-management/containers/IncidentOverviewPage/contants.ts
```diff
@@ -50,16 +52,24 @@ export type SortOption = {
 
 export const sortOptionsList: SortOption[] = [
   {
-    key: SortOptionKeys.ADRES,
-    label: SortOptionLabels.ADRES,
+    key: SortOptionKeys.ADDRESS,
+    label: SortOptionLabels.ADDRESS,
+    asc: SortOptions.ASSIGNED_USER_EMAIL_ASC,
```
alimpens, line 57: This seems to be swapped. Shouldn't the key `SortOptionKeys.ADDRESS` have `SortOptions.ADDRESS_ASC`?

src/signals/incident-management/containers/IncidentOverviewPage/contants.ts
```diff
@@ -20,23 +20,25 @@ export enum SortOptions {
...
+  ADDRESS = 'adres',
+  DATE = 'datum',
   ID = 'id',
-  STADSDEEL = 'stadsdeel',
+  DISTRICT = 'stadsdeel',
   STATUS = 'status',
   SUBCATEGORY = 'subcategorie',
-  URGENTIE = 'urgentie',
+  PRIORITY = 'urgentie',
+  ASSIGNED_USER_EMAIL = 'toegewezen_aan',
```
alimpens, line 30: This key is used in the sorting description, so it shows `Sorteer op toegewezen_aan (a-z)` now. Shouldn't we use the label instead of the key for that? And turn it into lowercase using CSS?

If we do, I think we can get rid of the `key` property in `sortOptionsList` entirely?

src/signals/incident-management/containers/IncidentOverviewPage/components/List/index.tsx
vdegraaf: I removed the export. It is used inside the styles file.

src/signals/incident-management/containers/IncidentOverviewPage/contants.ts
vdegraaf: Sharp

alimpens (approved):

PR #2841, opened 2024-04-09
[fix-5687]: Increase reply mail max length

alimpens:
Ticket: [SIG-5687](https://gemeente-amsterdam.atlassian.net/browse/SIG-5687)

This is the FE work needed to increase the reply mail max length. ~~This also needs some BE work though, so this PR will stay on 'Draft' until that work is finished~~ The BE work is done and deployed to acc.

[SIG-5687]: https://gemeente-amsterdam.atlassian.net/browse/SIG-5687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdegraaf (approved):
yes, looking good!

PR #2853, opened 2024-06-20
Show attribution on incident map

bartjkdp:
This PR makes sure we also show the attribution on the incident map. We already show attribution on the map while reporting.

It also uses CSS to hide the Leaflet attribution, instead of a DOM manipulation. This solution avoids timing issues.

image

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -26,9 +26,11 @@ const StyledMap = styled(MapComponent)`
     cursor: all-scroll;
   }
 
-  .leaflet-control-attribution a,
-  span {
-    display: none;
+  .leaflet-control-attribution {
+    a,
+    span:first-of-type {
+      display: none;
+    }
   }
 `
 
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -25,6 +25,11 @@ const StyledMap = styled(MapComponent)`
   &.leaflet-drag-target {
     cursor: all-scroll;
   }
+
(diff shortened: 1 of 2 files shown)


src/components/Map/Map.tsx
```diff
@@ -25,6 +25,11 @@ const StyledMap = styled(MapComponent)`
   &.leaflet-drag-target {
     cursor: all-scroll;
   }
+
+  .leaflet-control-attribution a,
+  span {
+    display: none;
+  }
 `
```
vdegraaf, line 33: This wil also hide the map references since all spans are set to display none. If you just select the first span it wil just hide the `|` and leaflet references.


```suggestion
.leaflet-control-attribution {
    a,
    span:first-of-type {
      display: none;
    }
  }
```

vdegraaf: Even better is not to use CSS but set the prefix of the attributionControl to `false` on the maps that shouldn't show it.

Example: `map.attributionControl.setPrefix(false)`

bartjkdp: For now I went with your suggestion for the CSS solution, as I think this is the most robust solution.

vdegraaf:
Hi @bartjkdp, discussed with Aram and we disaggree that this is the most robust solution. Hiding the prefix is a functionality of leaflet self, so we should use it as designed and not hide span's of which we don't know what is in it. 

I created a [PR here](https://github.com/Amsterdam/signals-frontend/pull/2886) with the desired solution.  Therefore closing this one.

PR #2923, opened 2024-08-19
[chore]: Update Leaflet types

alimpens:


src/shared/services/map-location/map-location.ts
```diff
@@ -8,15 +8,20 @@ import type { Incident } from 'types/api/incident'
...
 
-export const sanitizeCoordinates = (coordinates: LatLngTuple): LatLngTuple =>
-  coordinates.sort((a, b) => (a > b ? 1 : -1)).reverse() as LatLngTuple
+const convertCoordsToLatLng = (coordinates: LatLngTuple) => {
+  const coordsWithoutAltitude = [coordinates[0], coordinates[1]]
+
+  return coordsWithoutAltitude
+    .sort((a, b) => (a > b ? 1 : -1))
+    .reverse() as LatLngTuple
+}
```
vdegraaf, line 17: I think it is a bit strange the type is casted to the same incomming type while the function converts it to a different structure. An idea to create a new LatLng type which is used as return type for this function? This should also be used where you change the type to [number,number] now. Like this: 

Screenshot 2024-08-20 at 14 19 32

vdegraaf (approved):

PR #2988, opened 2024-10-16
Only show link to reports map when it is enabled

bartjkdp:
This fix only shows the links to the reports map when it is enabled.


vdegraaf (approved):
Makes senses!

